PR #58, opened 2026-08-22
fix(error_cascade): re-arm dedupe flag when a cascade clears

### LINKED ISSUE #57: [error_cascade] Dedupe flag never re-arms: only the first cascade per episode ever alerts
### Issue type

Bug

Yes — reproduced on `f1aabbc2c688e64102db7b6abefb99cb5d4e51be`.

### Source / version / custom code

Source install (`pip install -e . --no-deps`), commit `f1aabbc`, no custom code (stock `ErrorCascadeDetector` + `Config()` defaults).

### Environment

Windows 11, Python 3.13.1 (the behaviour is platform-independent — it is pure detector state logic).

---

## Problem

`ErrorCascadeDetector` escalates the **first** cascade in an episode and then goes permanently silent for that episode. Every subsequent, fully independent cascade is silently dropped, even after the agent has completely recovered in between.

The code's own comment states the intended behaviour, which is not what happens:

```python
# Dedupe: emit at most once per cascade episode, then stay quiet until
# the alarm condition clears and re-arms (issue #4).
self._fired: dict[str, bool] = {}
```

The flag never re-arms.

## Location

`src/snagline/detectors/error_cascade.py`, `ErrorCascadeDetector.observe()` — the early return when neither alarm rule holds:

```python
if not (consecutive_alarm or density_alarm):
    return None          # <-- _fired is never cleared here
```

`self._fired[episode_id]` is set to `True` on escalation and is only ever cleared by `reset(episode_id)`.

## Reproduction

```python
from snagline.detectors.error_cascade import ErrorCascadeDetector

def ev(i, error):
    return StepEvent(step_id=str(i), episode_id="ep", timestamp=float(i),
                     action_type="tool_call", action_signature="s", error=error)

d = ErrorCascadeDetector(config=Config())   # window=10, err=3, consecutive=3
i = 0
fired = []

for _ in range(3):                          # incident 1: 3 consecutive tool errors
    if d.observe(ev(i, True)): fired.append(("incident-1", i))
    i += 1

for _ in range(15):                         # full recovery: window is all-clean
    d.observe(ev(i, False)); i += 1

print("window:", list(d._windows["ep"]))    # [False] * 10
print("consecutive:", d._consecutive["ep"]) # 0
print("_fired:", d._fired.get("ep"))        # True  <-- still latched

for _ in range(5):                          # incident 2: 5 consecutive tool errors
    if d.observe(ev(i, True)): fired.append(("incident-2", i))
    i += 1

print(fired)                                # [('incident-1', 2)]  -- incident 2 lost
```

Output:

```
window: [False, False, False, False, False, False, False, False, False, False]
consecutive: 0
_fired: True

risks emitted: [('incident-1', 3, '3 consecutive errors')]
incident-1 alerted: True
incident-2 alerted: False
```

`LoopDetector` under the identical shape (loop → window flushed → loop again) correctly fires twice, which is the asymmetry that makes this look unintended rather than deliberate.

## Expected Behavior

Once the alarm condition has cleared (no errors left in the window, `consecutive == 0`), a later independent cascade in the same episode escalates again — one alert per cascade.

## Actual Behavior

Exactly one `error_cascade` risk per episode, for the lifetime of the episode's state. All later cascades return `None`.

## Root Cause

The `_fired` dedupe flag added for issue #4 is set on escalation but never cleared when the alarm condition clears. The `not (consecutive_alarm or density_alarm)` branch returns early without re-arming, so the flag latches until `reset()`. `LoopDetector.observe()` does clear its equivalent flag on the corresponding branch:

```python
if count < self.repeat_threshold:
    self._fired[event.episode_id] = False   # loop detector re-arms
    return None
```

## Impact

Bounded by how long an episode lives, so it is not a crash or a false positive — but the affected paths are the long-lived ones:

- `README.md` defines an episode as *"a single agent run, **a user session**"*. Any session-scoped episode covering multiple task attempts alerts only on its first cascade.
- The sidecar (`snagline serve`, `POST /events`) never calls `end_episode` at all — `grep -rn "end_episode" src/` shows callers only in the framework adapters, `raw.watch`, and the `replay`/`watch` CLI commands. For an external agent posting a session-scoped `episode_id` to the sidecar, the mute lasts for the life of the server process.
- `snagline watch --follow` tears down a single episode id derived from `--episode-id`/filename in its `finally:` block, which does not match episode ids carried in the ingested JSON lines.

Practically: a monitored agent that fails, recovers, then fails again reports only the first failure. That is a false negative in the "error cascade" signal, which is one of the two headline detectors.

Coverage note: `error_cascade.py` currently reports 100% line **and** branch coverage, because the missing re-arm is an absent statement rather than an untaken branch. No existing test drives a second cascade after recovery.

## Proposed Direction

Clear the flag on the already-existing "alarm cleared" branch, mirroring `LoopDetector`:

```python
if not (consecutive_alarm or density_alarm):
    self._fired[event.episode_id] = False
    return None
```

This preserves the issue #4 property it was added for: during a *sustained* cascade the alarm condition keeps holding, so the flag stays set and the detector still emits exactly once. I measured the alert rate over 1000-step synthetic streams with the fix applied — sustained cascade: 1 alert; all-clean: 0; isolated errors (1 in 10): 0; worst case (3 errors then 20 clean, repeating): 44 alerts, i.e. one per genuinely distinct incident rather than one per step. `DedupSink` remains the time-based rate limiter for operators who want fewer than one alert per incident.

I have this fix plus regression coverage (second-cascade-after-recovery, and a guard that a sustained cascade still emits once) validated locally against the full suite, ruff, and mypy — happy to open a PR for it.

abyyxhek:
Fixes #57

## Summary of Changes

`ErrorCascadeDetector`'s `_fired` dedupe flag was set on escalation but never cleared, so it latched for the life of an episode: only the **first** cascade ever alerted, and every later independent cascade was silently dropped — even after the agent had fully recovered in between.

The fix clears the flag on the already-existing "alarm cleared" branch of `observe()`, mirroring the equivalent re-arm in `LoopDetector.observe()`:

```python
 if not (consecutive_alarm or density_alarm):
+    self._fired[event.episode_id] = False
     return None
```

**Root cause.** The flag was introduced for issue #4 (alert spam: one risk per step during a sustained cascade). It was set to `True` on escalation and cleared only by `reset(episode_id)`. The `not (consecutive_alarm or density_alarm)` early return — the one branch that means *"the cascade is over"* — did not re-arm it. `LoopDetector` does re-arm on its corresponding branch (`if count < self.repeat_threshold: self._fired[...] = False`), which is the asymmetry that identified this as unintended. The flag's own comment already promised the missing behaviour: *"stay quiet until the alarm condition clears and re-arms"*.

Because the defect is an **absent statement** rather than an untaken branch, `error_cascade.py` reported 100% line *and* branch coverage while carrying it.

## Justification

The affected paths are the long-lived ones, which is where the detector matters most:

- `README.md` defines an episode as *"a single agent run, **a user session**"*. Any session-scoped episode spanning several task attempts alerted only on its first cascade.
- The sidecar (`snagline serve`, `POST /events`) never calls `end_episode` — `grep -rn "end_episode" src/` shows callers only in the framework adapters, `raw.watch`, and the `replay` / `watch` CLI commands. For an external agent posting a session-scoped `episode_id`, the mute lasted for the life of the server process.
- `snagline watch --follow` tears down one episode id derived from `--episode-id`/filename, which need not match the episode ids carried in the ingested JSON lines.

Net effect: an agent that fails, recovers, then fails again reported only the first failure — a false negative in one of the two headline detectors. The dedupe intent (don't page per step) is preserved; only the permanent latch is removed. `DedupSink` remains the time-based rate limiter for operators who want fewer than one alert per incident.

Nothing else changes: no new dependencies, no public API change, no `StepEvent.metadata` access, still O(1) per step (one dict store on the clean path), fail-open untouched.

## Kind of Contribution

Bug Fix, Tests

Not a performance change, but it adds one dict store to the per-step clean path, so I measured it. `python benchmarks/overhead_benchmark.py`, 200,000 steps, two runs each, same machine (Windows 11, Python 3.13.1):

| | median µs/step | p99 µs/step |
|---|---|---|
| before (`f1aabbc`) | 2.97 / 3.02 | 4.01 / 3.40 |
| after | 3.02 / 2.99 | 4.02 / 4.54 |

Median is unchanged within run-to-run variance on this machine; the p99 spread is larger than the effect being measured, so I would not read a regression into it.

I also measured the alert rate over 1000-step synthetic streams to confirm issue #4 does not come back:

| pattern | alerts / 1000 steps |
|---|---|
| all errors (sustained cascade) | 1 |
| all clean (healthy) | 0 |
| 1 error in 10 (isolated) | 0 |
| alternating error/clean | 1 |
| 3 errors then 7 clean, repeating | 1 |
| 3 errors then 20 clean, repeating | 44 |

The 44 is one alert per genuinely distinct incident (43 separate cascades, each with a full recovery between them) — not one per step, which is what issue #4 was about and would have been up to 1000 here.

## Unit Tests:

Two tests added to `tests/detectors/test_error_cascade.py`, following the existing synthetic-sequence shape in that file:

- `test_second_cascade_after_recovery_escalates_again` — cascade → 15 clean steps (window fully flushed) → second cascade. Asserts one risk for each and silence during recovery. **Fails on `master`** with `AssertionError: second cascade after recovery must escalate again / assert 0 == 1`; passes with the fix.
- `test_sustained_cascade_still_escalates_only_once` — 30 consecutive error steps produce exactly 1 risk, pinning the issue #4 property the flag exists for so this fix cannot regress it.

## Execution Tests:

Commands run locally against `f1aabbc` + this commit (Python 3.13.1, `pip install -e . --no-deps`):

```
pytest tests/detectors/test_error_cascade.py -q      ->  6 passed
pytest tests/test_issue_regressions.py tests/detectors/ -q  -> 25 passed
pytest tests/ -q                                    ->  189 passed, 1 skipped, 1 failed
ruff check src tests                                ->  All checks passed!
ruff format --check src tests                       ->  71 files already formatted
mypy src                                            ->  Success: no issues found in 37 source files
coverage                                            ->  88.19% (gate: 80%)
```

The one failure is pre-existing and unrelated — `tests/test_hook_bridge.py::test_watch_file_follow_sees_appended_lines` fails identically on unmodified `master` on Windows, because `Popen.send_signal(signal.SIGINT)` raises `ValueError: Unsupported signal: 2` there. CI runs `ubuntu-latest`, where it passes.

End-to-end behaviour:

- **CLI**: `snagline replay tests/fixtures/trajectories/injected_error_cascade.jsonl` still reports the cascade (`{"trigger": "error_cascade", "detail": "3 consecutive errors", ...}`); `snagline replay .../healthy_run.jsonl` stays silent. The `healthy_run` fixture produces 0 risks from this detector, per the `DETECTOR_GUIDE.md` no-false-positives bar.
- **Monitor + sink**: a single long-lived episode (`session-42`) driven through `Monitor.default()` with the sequence `3 errors → 15 clean → 4 errors → 15 clean → 3 errors` now delivers **3** `error_cascade` risks to the sink (steps 2, 20, 39). On `master` it delivers 1.

Fail-open verification (`Monitor.default()`, `fail_open=True`):

- Malformed / boundary events through `observe()` — empty `action_type` and `action_signature`, empty `episode_id`, `NaN` timestamp, `inf` latency: no exception.
- `reset()` on an unknown episode, and repeated `reset()`: no exception.
- A sink raising on **both** cascades: host unaffected, `monitor.metrics()` reports `sink_errors: 3`, nothing propagates.
- `Monitor(..., fail_open=False)` still propagates as documented, so the strict-mode contract is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Error-cascade alerts can now trigger again after the system fully recovers and a new cascade begins.
  * Ongoing cascades continue to generate only one alert, reducing duplicate notifications.

* **Tests**
  * Added coverage for re-triggering after recovery and suppressing repeated alerts during sustained cascades.

### DIFF
--- a/src/snagline/detectors/error_cascade.py
+++ b/src/snagline/detectors/error_cascade.py
@@ -55,8 +55,8 @@ class ErrorCascadeDetector:
         )
         self._windows: dict[str, deque] = {}
         self._consecutive: dict[str, int] = {}
-        # Dedupe: emit at most once per cascade episode, then stay quiet until
-        # the alarm condition clears and re-arms (issue #4).
+        # Dedupe: emit at most once per cascade, then stay quiet until the alarm
+        # condition clears and re-arms (issue #4).
         self._fired: dict[str, bool] = {}
 
     def _is_error(self, event: StepEvent) -> bool:
@@ -87,6 +87,14 @@ class ErrorCascadeDetector:
         )
 
         if not (consecutive_alarm or density_alarm):
+            # The cascade cleared: re-arm so a later, independent cascade in the
+            # same episode escalates again (mirrors ``LoopDetector``). Without
+            # this the flag latches for the life of the episode, and a long-lived
+            # episode -- a user session, or a sidecar episode that never calls
+            # ``end_episode`` -- alerts exactly once, ever. A *sustained* cascade
+            # still emits only once (issue #4): the flag clears only when neither
+            # rule holds any more.
+            self._fired[event.episode_id] = False
             return None
         # Already escalated this cascade -- suppress until it clears.
         if self._fired.get(event.episode_id, False):
--- a/tests/detectors/test_error_cascade.py
+++ b/tests/detectors/test_error_cascade.py
@@ -61,3 +61,40 @@ def test_reset_clears_state():
     d.observe(_event(1, True))
     d.reset("ep")
     assert d.observe(_event(2, True)) is None
+
+
+def test_second_cascade_after_recovery_escalates_again():
+    # The dedupe flag must re-arm once the alarm clears, so a second,
+    # independent cascade in the same episode still alerts. A long-lived
+    # episode (a user session, or a sidecar episode that never calls
+    # end_episode) would otherwise be muted after its first cascade.
+    d = ErrorCascadeDetector(window_size=10, error_threshold=3, consecutive_threshold=3)
+    step = 0
+
+    def run(errors: list[bool]) -> list:
+        nonlocal step
+        out = []
+        for err in errors:
+            r = d.observe(_event(step, err))
+            step += 1
+            if r is not None:
+                out.append(r)
+        return out
+
+    first = run([True] * 3)
+    assert len(first) == 1, "first cascade must escalate exactly once"
+
+    # Full recovery: enough clean steps to flush every error out of the window.
+    assert run([False] * 15) == [], "healthy recovery must stay silent"
+
+    second = run([True] * 3)
+    assert len(second) == 1, "second cascade after recovery must escalate again"
+    assert second[0].trigger == "error_cascade"
+
+
+def test_sustained_cascade_still_escalates_only_once():
+    # Guard the issue #4 property the dedupe flag exists for: while the alarm
+    # condition holds, the detector must not re-fire on every step.
+    d = ErrorCascadeDetector(window_size=10, error_threshold=3, consecutive_threshold=3)
+    risks = [d.observe(_event(i, True)) for i in range(30)]
+    assert len([r for r in risks if r is not None]) == 1


abyyxhek:
Two notes on the check status here, so the red mark is not misread:

**1. CI has not run yet — it needs a maintainer click.** `CI`, `Semantic PR Title`, and `Auto Assign` all use the `pull_request` trigger and came back `action_required`, which is GitHub's first-time-contributor approval gate for fork PRs. They will run after "Approve and run workflows". The results in the PR body are from running the CI steps locally (`ruff check`, `ruff format --check`, `mypy src`, `pytest tests/ -q`) against `f1aabbc` + this commit; I am not claiming green CI here.

**2. The failing `welcome` check is pre-existing and unrelated to this PR.** `actions/first-interaction@v3` reads its inputs as `issue_message` / `pr_message` (underscores), but `.github/workflows/welcome.yml` passes `issue-message` / `pr-message` (hyphens), so the action exits with:

```
Error: Input required and not supplied: issue_message
```

The last successful `welcome` run was #56 itself (the v1 → v3 bump, 08:02Z); every run after that bump has failed, including the one on issue #57 before this PR existed. It will fail for every first-time issue and PR until the input keys are renamed. Happy to open a separate one-line PR for that if useful — I left it out of this one to keep the diff scoped to the detector.

Cyrax321:
👋 Welcome, and thanks for this pull request!

**fix(error_cascade): re-arm dedupe flag when a cascade clears**

Getting a change from idea to open PR is the hard part, and you've already done it. All CI test jobs (py3.10–py3.13) are passing, great work. The remaining red checks are stale workflow runs from before the automation fixes and can be safely ignored.

A maintainer will review it soon. Review feedback is normal and expected, so don't worry about getting everything perfect on the first try. Thanks again for contributing!

Cyrax321:
@coderabbitai review

Cyrax321:
/yuki What's this pr about in short

yuki-fuyutsuki:
PR #58 `fix/error-cascade-rearm` fixes #57 — `src/snagline/detectors/error_cascade.py`:

* **Bug:** `ErrorCascadeDetector._fired` dedupe flag (added for #4 to avoid 1 alert/step) was set on escalation but only cleared by `reset()`, never when cascade cleared. So 1st cascade in an episode latched the detector — all later independent cascades in same `episode_id`/session were silently dropped. Affected long-lived sessions (`README.md` episode = user session, `snagline serve` never calls `end_episode`).

* **Fix:** 1-line re-arm on existing `not (consecutive_alarm or density_alarm)` branch in `observe()`: `self._fired[episode_id] = False` — mirrors `LoopDetector.observe()` behavior. Keeps 1 alert per sustained cascade, adds `O(1)` dict store. Comment/ docstring updated.

* **Tests:** `tests/detectors/test_error_cascade.py` +37: `test_second_cascade_after_recovery_escalates_again` (cascade → 15 clean → cascade → 2 alerts, fails on `master`) and `test_sustained_cascade_still_escalates_only_once` (30 errors → 1 alert, regression for #4).

Scope: Bug Fix, 2 files, +47/-2. No API/deps change.

[github run](/Cyrax321/SNAGLINE/actions/runs/32594697916)

Cyrax321:
**Review status: Changes Required**

CodeRabbit completed its review of this PR. The fix itself looks correct and all CI test jobs pass, but one change is needed before we can merge.

**What is needed:**

Add a density-only regression test to `tests/detectors/test_error_cascade.py`.

**Why:**

Both existing tests trigger `consecutive_alarm` and `density_alarm` at the same time. That means nothing in the suite verifies re-arming or deduplication when `density_alarm` is the only active rule, which is exactly the scenario this PR fixes.

**How to fix it:**

1. Build an alternating error pattern so `consecutive_alarm` never fires.
2. Set `consecutive_threshold` higher than `error_threshold`.
3. Assert that exactly one escalation happens, recovery is silent after enough clean observations, and a later sustained density escalation does not produce a duplicate alert.

Please push the test to this branch and we will get it merged. Thanks for the contribution.

Cyrax321 (changes requested):
Add a density-only regression test in tests/detectors/test_error_cascade.py covering re-arm and dedup when consecutive_alarm never fires, then we merge. Thank you for the contribution!

Issue #15, opened 2026-08-14
make_signature truncates SHA-256 to 16 hex chars and joins parts with '||' (collision + ambiguity risk)


Cyrax321:
## Description

`make_signature` (`src/snagline/events.py`) builds the loop-detection key as `hashlib.sha256("||".join([action_type, tool_name or "", *stable_parts])).hexdigest()[:16]`.

Two robustness concerns:

1. Truncating to 16 hex chars keeps only 64 bits of the digest. At very high event volumes the birthday-bound collision probability, while small, is non-zero and would make two distinct actions share a signature.
2. Parts are joined with the literal `||`. If any `stable_part` itself contains `||`, two genuinely different actions can produce an identical joined string and therefore the same signature, defeating loop detection.

## Impact

- Rare false negatives / rare collisions in loop detection.

## Suggested fix

- Use a non-printable / unambiguous separator (e.g. NUL) or JSON-encode the parts before hashing.
- Return a longer digest (or the full digest) to reduce collision risk.

## Additional context

Low severity, but worth hardening given loop detection is a tier-1 feature.

Cyrax321:
Resolved: make_signature JSON-encodes its parts and returns the full 64-char SHA-256 digest, removing the collision and '||' ambiguity. Covered by tests/test_issue_regressions.py::test_make_signature_returns_full_digest_and_avoids_ambiguity. PR #23

PR #63, opened 2026-08-22
fix(crewai): key action_signature on tool_input, not step output

### LINKED ISSUE #61: [adapters/crewai] action_signature is built from step output instead of tool_input, breaking loop detection in both directions
### Issue type

Bug

Yes — `c2390af2fbdf37e4278a022141714ac2a71a3495`

### Source

source

### SNAGLINE version

`c2390af` (0.1.0)

<sub>(The template suggests `import snagline; print(snagline.__version__)`, but that attribute isn't exported, so I'm giving the commit SHA.)</sub>

### Custom code

No

Windows 11 — the affected code is pure Python with no OS-specific paths, so this reproduces anywhere.

### Python version

3.13

### Current behavior?

`docs/ADAPTER_GUIDE.md` has a section named "Signature rules (the part people get wrong)" which lists two failure modes:

> - **Include** the logical action: tool name, target element, endpoint, the shape of the arguments.
> - **Exclude** anything volatile: timestamps, request/session ids, nonces, retry counters. Including them makes every retry look unique and silently defeats loop detection. (Excluding them entirely is equally wrong in the other direction - a constant `args=""` makes *every* LLM call look like a loop; ...)

The CrewAI adapter hits **both** of them. `src/snagline/adapters/crewai.py:89` builds the signature from the step's *output*:

```python
sig = make_signature(action_type, tool_name, text or str(d))
```

where `_extract_text` (line 62) reads `("text", "thought", "output", "content", "result")`. On CrewAI's `AgentAction`, `text` is the raw LLM block (`"Thought: ...\nAction: ...\nAction Input: ..."`) and `thought` is model prose — both re-worded on every attempt. Meanwhile `tool_input`, the argument payload that makes two tool calls "the same attempt", is never read by the adapter at all.

The two consequences:

1. **False negative** — an agent retrying the *identical* tool call with the *identical* `tool_input` gets a fresh signature each time, because the model re-words its thought. `LoopDetector` never fires. This is the exact scenario the detector exists to catch, and it fails silently: the monitor reports healthy.
2. **False positive** — because `tool_input` is excluded, an agent whose prose is templated (identical `text`) but whose arguments differ every step collapses to one signature and is reported as a loop.

Expected: the signature for a CrewAI tool call is derived from the tool name plus its arguments, so repeated identical calls share a signature and calls with different arguments do not — matching every other adapter in the tree:

| adapter | stable part hashed |
| --- | --- |
| `adapters/raw.py:36` | `str(args)` |
| `adapters/langchain_adapter.py:73` | `str(args)` |
| `adapters/autogen.py:119` | `args` |
| `adapters/claude_code.py:116` | `json.dumps(payload.get("tool_input") or {}, sort_keys=True)` |
| `adapters/crewai.py:89` | **the step output text** |

The adapter's own module docstring (lines 20–22) already says it reads `tool_input`:

> we read common attributes (``tool``, ``tool_input``, ``output``/``text``, ``error``)

but nothing in the file references `tool_input`, so the docstring describes the intended behavior rather than the current one.

```shell
# No CrewAI install needed - the adapter is duck-typed, and this models the
# real shape of crewai.agents.parser.AgentAction.
cat > repro.py <<'PY'
from snagline.adapters.crewai import _map_step, snagline_step_callback


class AgentAction:  # shape-faithful stand-in for crewai's AgentAction
    def __init__(self, thought, tool, tool_input, text, result=""):
        self.thought, self.tool, self.tool_input = thought, tool, tool_input
        self.text, self.result = text, result


class Collector:
    def __init__(self):
        self.risks = []

    def emit(self, risk):
        self.risks.append(risk)


def run(steps):
    sink = Collector()
    mon = Monitor([LoopDetector()], [sink])
    cb = snagline_step_callback(mon, "ep-crew")
    for s in steps:
        cb(s)
    return [r for r in sink.risks if r.trigger == "loop"]


def sig(step):
    return _map_step(
        step, episode_id="e", step_id="0", agent_name=None, clock=lambda: 0.0
    ).action_signature


TOOL, ARGS = "search", '{"query": "Q3 revenue"}'

# 1. Stuck agent: same tool, byte-identical tool_input, re-worded thought.
stuck = [
    AgentAction(t, TOOL, ARGS, f"Thought: {t}\nAction: {TOOL}\nAction Input: {ARGS}",
                "Error: connection reset")
    for t in ("I need the Q3 revenue figure",
              "Let me try searching for the Q3 revenue again",
              "The search failed, retrying the Q3 revenue lookup",
              "I should search once more for Q3 revenue",
              "Trying the Q3 revenue search another time",
              "Let me retry that Q3 revenue query")
]
print("stuck agent, 6x identical tool_input")
print("  loop risks      :", len(run(stuck)), " (expected >=1)")
print("  distinct sigs   :", len({sig(s) for s in stuck}), "/ 6")

# 2. Control: identical text as well -> the detector itself is fine.
same = [AgentAction("t", TOOL, ARGS, "identical text") for _ in range(6)]
print("control, identical text")
print("  loop risks      :", len(run(same)), " (expected >=1)")

# 3. Legitimate iteration: templated prose, 6 DIFFERENT tool_inputs.
iterating = [
    AgentAction("Look up the next city", TOOL, f'{{"query": "weather in {c}"}}',
                "Thought: Look up the next city\nAction: search")
    for c in ("Paris", "Berlin", "Tokyo", "Lagos", "Lima", "Oslo")
]
print("legitimate iteration, 6 different tool_inputs")
print("  loop risks      :", len(run(iterating)), " (expected 0)")
print("  distinct sigs   :", len({sig(s) for s in iterating}), "/ 6")

# 4. Root cause isolated: tool_input is absent from the signature.
a = AgentAction("t", TOOL, '{"q": "alpha"}', "same text")
b = AgentAction("t", TOOL, '{"q": "omega"}', "same text")
print("two steps differing ONLY in tool_input")
print("  signatures equal:", sig(a) == sig(b), " (expected False)")
PY
python repro.py
```

### Relevant log output

```shell
stuck agent, 6x identical tool_input
  loop risks      : 0  (expected >=1)
  distinct sigs   : 6 / 6
control, identical text
  loop risks      : 1  (expected >=1)
legitimate iteration, 6 different tool_inputs
  loop risks      : 1  (expected 0)
  distinct sigs   : 1 / 6
two steps differing ONLY in tool_input
  signatures equal: True  (expected False)
```

Case 2 is the control: with the text pinned identical the detector fires normally, so the detector and the wiring are fine — only the choice of stable part in `_map_step` is at fault.

### Impact

CrewAI is a documented, advertised integration (`README.md:255`, `docs/ADAPTER_GUIDE.md:64`, the `crewai` extra in `pyproject.toml`). For those users the tier-1 loop detector — SNAGLINE's headline capability — is effectively disabled against real stuck loops, and can fire on healthy runs. Both directions are silent: nothing raises, nothing logs, and the monitor's metrics look normal, so there is no signal that detection is degraded.

`tests/test_adapters_autogen_crewai.py` covers `action_type`, `tool_name`, `error`, and `latency_ms` for this adapter but makes no assertion about `action_signature`, which is why the behavior went unnoticed. Every other adapter has that assertion — `tests/adapters/test_raw_adapter.py:42` ("same args -> same signature (loop detection depends on this)"), `test_langgraph_adapter.py:62-65`, `test_claude_code_adapter.py:69,78`.

### Proposed fix

Read the tool arguments and use them as the stable part for tool calls, mirroring `_extract_tool_name`'s dual lookup so both CrewAI step shapes (`step.tool_input` and a nested `action` dict) are handled:

```python
def _extract_tool_input(step: dict[str, Any]) -> str | None:
    action = step.get("action") if isinstance(step.get("action"), dict) else None
    for src in (action, step):
        if src and src.get("tool_input") is not None:
            return json.dumps(src["tool_input"], sort_keys=True, default=str)
    return None
```

and in `_map_step`:

```python
tool_args = _extract_tool_input(d)
stable = tool_args if (action_type == "tool_call" and tool_args is not None) else (text or str(d))
sig = make_signature(action_type, tool_name, stable)
```

`sort_keys=True` matches the canonicalization `adapters/claude_code.py:116` already uses. Non-tool `agent_step` events keep hashing `text` — there is no argument payload on those, so that is a separate question and I would leave it alone. Nothing persists `action_signature` to disk (neither `baseline.py` nor `baseline_store.py` stores it), so this is not a stored-format change.

Happy to open a PR with this plus the missing `action_signature` regression tests for the adapter, if the approach looks right.

abyyxhek:
## Summary of Changes

`adapters/crewai.py` built `action_signature` from the step's **output** text:

```python
sig = make_signature(action_type, tool_name, text or str(d))   # _map_step, line 89
```

`_extract_text` reads `("text", "thought", "output", "content", "result")`. On CrewAI's `AgentAction`, `text` is the raw LLM block and `thought` is model prose — both re-worded on every attempt — while `tool_input`, the argument payload, was never read by the adapter at all.

This change adds `_extract_tool_input` and uses the arguments as the stable part for tool calls:

```python
tool_args = _extract_tool_input(d)
if action_type == "tool_call" and tool_args is not None:
    stable = tool_args
else:
    stable = text or str(d)
sig = make_signature(action_type, tool_name, stable)
```

`_extract_tool_input` mirrors the dual lookup `_extract_tool_name` already does (top-level `tool_input`, or nested under `action`), and canonicalizes with `json.dumps(..., sort_keys=True)` — the same treatment `adapters/claude_code.py:116` applies. Non-tool `agent_step` events carry no argument payload, so that path is untouched.

Two deliberate details:

- **Strict serialization, not `default=str`.** An arbitrary object's default repr embeds its memory address (`<Foo object at 0x…>`), which varies per instance and would have re-introduced the exact volatility this PR removes.
- **A payload JSON cannot represent returns `None`** and the caller keeps the previous stable part. `_map_step` runs in CrewAI's thread *before* `Monitor.ingest`, so it is outside the fail-open guard — an unorderable key type, a reference cycle, or an exotic object must not surface inside the user's agent. Verified below.

No public API change, no new dependency (`json` is stdlib), and nothing persists `action_signature` (neither `baseline.py` nor `baseline_store.py` stores it), so this is not a stored-format change.

## Justification

Fixes #61. `docs/ADAPTER_GUIDE.md` has a section named "Signature rules (the part people get wrong)" listing two failure modes, and this adapter hit **both**:

> Including them makes every retry look unique and silently defeats loop detection. (Excluding them entirely is equally wrong in the other direction - a constant `args=""` makes *every* LLM call look like a loop …)

1. **False negative** — an agent retrying the identical call with identical `tool_input` got a fresh signature every attempt, because the model re-words its thought. `LoopDetector` never fired. That is precisely the scenario the detector exists to catch.
2. **False positive** — with the arguments excluded, an agent whose prose is templated but whose arguments differ each step collapsed to one signature and was reported as a loop.

Both are silent: nothing raises, nothing logs, and the monitor's metrics look normal, so a CrewAI user has no signal that tier-1 detection is degraded. CrewAI is a documented, advertised integration (`README.md:255`, `docs/ADAPTER_GUIDE.md:64`, the `crewai` extra).

The fix also brings the adapter in line with every sibling, which all hash the input:

| adapter | stable part hashed |
| --- | --- |
| `adapters/raw.py:36` | `str(args)` |
| `adapters/langchain_adapter.py:73` | `str(args)` |
| `adapters/autogen.py:119` | `args` |
| `adapters/claude_code.py:116` | `json.dumps(payload.get("tool_input") or {}, sort_keys=True)` |
| `adapters/crewai.py` | **was** the step output — now `tool_input` |

and makes the module docstring's existing claim that it reads `tool_input` (lines 20–22) true.

## Kind of Contribution

Bug Fix, Tests

## Benchmark

The core ingest path is not touched by this diff — only `adapters/crewai.py` changed — so `benchmarks/overhead_benchmark.py` cannot move here. Three runs on this branch for the noise band, so the claim is checkable rather than asserted:

```
median 2.91 us/step, p99 4.94
median 2.91 us/step, p99 3.76
median 2.99 us/step, p99 5.90
```

The function that *did* change is `_map_step`, so I measured it directly (20k calls x 7 repeats, median of per-call medians, same machine, before = `c2390af`):

| payload | before | after | delta |
| --- | --- | --- | --- |
| `tool_input` as dict | 5.469 us | 7.965 us | +2.50 us |
| `tool_input` as str | 5.417 us | 6.673 us | +1.26 us |

That is one additional `json.dumps` per tool-call step. It fires once per agent step — a step that contains an LLM round-trip of roughly 10^5–10^6 us — so the added cost is about one part in 10^5 of the smallest realistic step, against restoring a detector that was silently inert.

I considered short-circuiting `isinstance(val, str)` past `json.dumps` to recover most of that, and rejected it: `tool_input="{}"` would then produce the same signature as `tool_input={}`, and `make_signature`'s docstring calls out exactly that ambiguity class (issue #15). Not worth trading a correctness guarantee for microseconds on this path. Happy to revisit if you would rather have the cycles.

## Unit Tests

Four tests added to `tests/test_adapters_autogen_crewai.py`. The existing CrewAI tests cover `action_type`, `tool_name`, `error`, and `latency_ms` but make no assertion about `action_signature`, which is why this went unnoticed — every other adapter has that assertion (`tests/adapters/test_raw_adapter.py:42`, `test_langgraph_adapter.py:62-65`, `test_claude_code_adapter.py:69,78`).

- `test_crewai_signature_is_stable_across_reworded_prose` — same tool + same `tool_input`, different prose ⇒ same signature; also asserts the nested `action` dict shape resolves identically.
- `test_crewai_different_tool_inputs_have_different_signatures` — identical prose, different arguments ⇒ different signatures (the false-positive direction).
- `test_crewai_stuck_tool_loop_escalates_end_to_end` — through a real `Monitor` + `LoopDetector`, a stuck agent escalates exactly once (also pins the issue #4 dedupe).
- `test_crewai_exotic_tool_input_does_not_raise_into_the_host` — unorderable keys, a reference cycle, a set, and a bare `object()` all map without raising.

All three of the first tests fail on `c2390af` and pass with the fix:

```
$ python -m pytest tests/test_adapters_autogen_crewai.py -q -k "signature or stuck"   # before
FAILED tests/test_adapters_autogen_crewai.py::test_crewai_signature_is_stable_across_reworded_prose
FAILED tests/test_adapters_autogen_crewai.py::test_crewai_different_tool_inputs_have_different_signatures
FAILED tests/test_adapters_autogen_crewai.py::test_crewai_stuck_tool_loop_escalates_end_to_end
3 failed, 15 deselected

# the end-to-end one before the fix:
>       assert [r.trigger for r in risks] == ["loop"]
E       AssertionError: assert [] == ['loop']
```

Full CI steps, run locally against `c2390af` + this commit:

```
$ ruff check src tests                 -> All checks passed!
$ ruff format --check src tests        -> 71 files already formatted
$ mypy src                             -> Success: no issues found in 37 source files
$ python -m pytest tests/ -q           -> 1 failed, 191 passed, 1 skipped
                                          Total coverage: 88.41% (was 88.19%)
```

The one failure is `tests/test_hook_bridge.py::test_watch_file_follow_sees_appended_lines`, which is **pre-existing and unrelated**: it fails identically on a clean `c2390af` with `ValueError: Unsupported signal: 2`, because `subprocess.send_signal(SIGINT)` is unsupported on Windows, which is my local platform. CI runs `ubuntu-latest`, where it passes. I am not claiming green CI here — as a first-time contributor these workflows need a maintainer's "Approve and run workflows" click.

`adapters/crewai.py` coverage is 86%; the uncovered lines (39–47) are the pre-existing `_to_dict` fallbacks, not new code. `_extract_tool_input` is fully covered.

## Execution Tests

End to end through the real adapter and a real `Monitor`, modelling the shape of `crewai.agents.parser.AgentAction` (no CrewAI install needed — the adapter is duck-typed). Full script in #61.

| scenario | expected | before | after |
| --- | --- | --- | --- |
| stuck agent, 6x identical `tool_input`, re-worded thought | ≥1 loop risk | **0 risks**, 6/6 distinct signatures | 1 risk, 1/6 distinct |
| control: identical prose too | ≥1 loop risk | 1 risk | 1 risk |
| legitimate iteration, 6 different `tool_input` | 0 risks | **1 risk**, 1/6 distinct | 0 risks, 6/6 distinct |
| two steps differing only in `tool_input` | different signatures | **identical** | different |

The control row is the point: with the prose pinned identical the detector fired normally even before the fix, so the detector and the wiring were always fine — only the choice of stable part in `_map_step` was wrong.

Fail-open check, since this change puts a `json.dumps` on a path outside `Monitor`'s guard — each of these was ingested through `snagline_step_callback` and none raised:

```
  OK  mixed-type dict keys + sort_keys
  OK  circular reference
  OK  object with a raising __str__
  OK  set (not JSON-serializable)
  OK  bytes / None / empty dict / empty string / zero
  OK  nested dict, keys out of order   -> same signature regardless of key order
```

On `c2390af` + my first draft (which used `default=str`) the first three of those raised `TypeError`, `ValueError`, and `ZeroDivisionError` straight into the caller; the strict-with-fallback form above is why the final version does not.

### DIFF
--- a/src/snagline/adapters/crewai.py
+++ b/src/snagline/adapters/crewai.py
@@ -25,6 +25,7 @@ the adapter stays loosely coupled to a specific CrewAI release.
 from __future__ import annotations
 
 import itertools
+import json
 import time
 from collections.abc import Callable
 from typing import Any
@@ -67,6 +68,39 @@ def _extract_text(step: dict[str, Any]) -> str:
     return ""
 
 
+def _extract_tool_input(step: dict[str, Any]) -> str | None:
+    """Canonical form of the tool arguments, or ``None`` when unavailable.
+
+    The arguments -- not the model's prose -- are what make two tool calls "the
+    same attempt" (see ``make_signature`` and docs/ADAPTER_GUIDE.md). CrewAI
+    exposes them on the step itself or nested under ``action``, so mirror the
+    dual lookup :func:`_extract_tool_name` already does. ``sort_keys`` matches
+    the canonicalization the Claude Code adapter uses, so a dict payload hashes
+    the same regardless of key order.
+
+    A payload JSON cannot represent (unorderable key types, a reference cycle,
+    an arbitrary object) yields ``None`` so the caller keeps its previous stable
+    part. Serializing is deliberately strict rather than falling back to
+    ``str``: the default repr of an object embeds its memory address, which
+    would vary per attempt and defeat loop detection all over again.
+    """
+    action = step.get("action") if isinstance(step.get("action"), dict) else None
+    for src in (action, step):
+        if src is None:
+            continue
+        val = src.get("tool_input")
+        if val is None:
+            continue
+        try:
+            return json.dumps(val, sort_keys=True)
+        except Exception:
+            # _map_step runs in the framework's thread, before Monitor.ingest
+            # and therefore outside its fail-open guard, so an exotic payload
+            # must never propagate into the host agent.
+            return None
+    return None
+
+
 def _map_step(
     step: Any,
     *,
@@ -86,7 +120,17 @@ def _map_step(
     action: dict[str, Any] = raw_action if isinstance(raw_action, dict) else {}
     if latency is None:
         latency = action.get("latency_ms")
-    sig = make_signature(action_type, tool_name, text or str(d))
+    # Key the signature on the *attempt* -- tool plus arguments -- never on the
+    # step's output. Model prose is re-worded on every retry, so hashing it made
+    # a stuck loop look like a stream of unique actions, while omitting the
+    # arguments collapsed a legitimate iteration into one signature (issue #61).
+    # Non-tool steps carry no argument payload, so they keep hashing the text.
+    tool_args = _extract_tool_input(d)
+    if action_type == "tool_call" and tool_args is not None:
+        stable = tool_args
+    else:
+        stable = text or str(d)
+    sig = make_signature(action_type, tool_name, stable)
     return StepEvent(
         step_id=step_id,
         episode_id=episode_id,
--- a/tests/test_adapters_autogen_crewai.py
+++ b/tests/test_adapters_autogen_crewai.py
@@ -6,6 +6,8 @@ import asyncio
 
 from snagline.adapters.autogen import SnaglineAutogenHandler, run_and_monitor
 from snagline.adapters.crewai import observe_crewai_step, snagline_step_callback
+from snagline.detectors.loop import LoopDetector
+from snagline.monitor import Monitor
 
 
 class _Collector:
@@ -216,3 +218,75 @@ def test_crewai_callback_close_ends_episode():
     cb.close()
     assert mon.ended == ["ep-7"]
     assert len(mon.events) == 1
+
+
+def _crewai_step(tool_input, text, tool="search"):
+    # Shape of crewai.agents.parser.AgentAction: the raw LLM block lands in
+    # ``text``, the argument payload in ``tool_input``.
+    return {"tool": tool, "tool_input": tool_input, "text": text}
+
+
+def test_crewai_signature_is_stable_across_reworded_prose():
+    # Issue #61: the signature must come from the tool arguments, not the step
+    # output. A stuck agent retries the identical call while the model re-words
+    # its thought each attempt; if that prose feeds the hash, every retry looks
+    # unique and the loop detector never fires.
+    mon = _Collector()
+    args = '{"query": "Q3 revenue"}'
+    first = observe_crewai_step(  # noqa: F821
+        mon, "ep", _crewai_step(args, "Thought: let me search for that")
+    )
+    second = observe_crewai_step(  # noqa: F821
+        mon, "ep", _crewai_step(args, "Thought: the search failed, retrying")
+    )
+    assert first.action_signature == second.action_signature
+    # The nested ``action`` dict shape must resolve to the same signature.
+    nested = observe_crewai_step(  # noqa: F821
+        mon,
+        "ep",
+        {"action": {"tool": "search", "tool_input": args}, "text": "unrelated prose"},
+    )
+    assert nested.action_signature == first.action_signature
+
+
+def test_crewai_different_tool_inputs_have_different_signatures():
+    # Issue #61, other direction: leaving the arguments out entirely collapses a
+    # legitimate iteration over distinct inputs into one signature, which the
+    # loop detector then reports as a loop. Same prose, different arguments.
+    mon = _Collector()
+    text = "Thought: look up the next city"
+    a = observe_crewai_step(mon, "ep", _crewai_step('{"q": "Paris"}', text))  # noqa: F821
+    b = observe_crewai_step(mon, "ep", _crewai_step('{"q": "Berlin"}', text))  # noqa: F821
+    assert a.action_signature != b.action_signature
+
+
+def test_crewai_stuck_tool_loop_escalates_end_to_end():
+    # The user-visible consequence, through a real Monitor and LoopDetector: an
+    # agent that keeps issuing the same call with the same arguments must
+    # escalate exactly once (the dedupe from issue #4 keeps it to one).
+    risks = []
+
+    class _Sink:
+        def emit(self, risk):
+            risks.append(risk)
+
+    monitor = Monitor([LoopDetector()], [_Sink()])
+    cb = snagline_step_callback(monitor, "ep-stuck")  # noqa: F821
+    for i in range(5):
+        cb(_crewai_step('{"query": "Q3 revenue"}', f"Thought: attempt {i}"))
+    assert [r.trigger for r in risks] == ["loop"]
+
+
+def test_crewai_exotic_tool_input_does_not_raise_into_the_host():
+    # Mapping runs in the framework's thread, before Monitor.ingest and so
+    # outside its fail-open guard. A payload JSON cannot canonicalize (unorderable
+    # key types, a reference cycle, an arbitrary object) must fall back to the
+    # previous stable part rather than surface inside the user's agent.
+    mon = _Collector()
+    cb = snagline_step_callback(mon, "ep-odd")  # noqa: F821
+    cycle: dict = {"a": None}
+    cycle["a"] = cycle
+    for payload in ({1: "a", "b": 2}, cycle, {"x", "y"}, object()):
+        cb({"tool": "search", "tool_input": payload, "text": "prose"})
+    assert len(mon.events) == 4
+    assert all(ev.action_signature for ev in mon.events)


Cyrax321 (approved):
Thank you for the contribution. This PR fixes CrewAI action_signature being keyed on volatile model prose instead of the stable tool_input, which made LoopDetector silently inert in both directions.

I reviewed src/snagline/adapters/crewai.py on origin/master where _map_step hashed text/thought/output. The fix adds _extract_tool_input mirroring _extract_tool_name dual lookup and canonicalizes with json.dumps sort_keys matching claude_code, uses it for tool_call only, and falls back to text for agent_step. Strict serialization without default=str avoids address-varying reprs and the None fallback keeps the host fail-open.

I applied the patch and ran tests/test_adapters_autogen_crewai.py: 19 passed including stable signature across reworded prose, different tool_inputs produce different signatures, stuck loop escalates end to end, and exotic payloads do not raise into the host.

No blocking issues found. This aligns CrewAI with all sibling adapters.

Approved.

Cyrax321:
**Saki**: This PR earned its approval honestly. Thank you for the thoughtful contribution. We look forward to whatever you build next.

Issue #69, opened 2026-08-23
test_hook_bridge: send_signal(SIGINT) fails on Windows and orphans the watch subprocess


abyyxhek:
### Issue type

Bug

### Have you reproduced the bug on the latest commit of `master`?

Yes

### Source

source

### SNAGLINE version

0.1.0 (28fd79a)

### Custom code

No

### OS platform and distribution

Windows 11 Home 26200 (Windows-only failure)

### Python version

3.13.1

### Current behavior?

`tests/test_hook_bridge.py::test_watch_file_follow_sees_appended_lines` cannot
run on Windows. It signals the `snagline watch --follow` child with
`proc.send_signal(signal.SIGINT)` (`tests/test_hook_bridge.py:164`), and
`subprocess.Popen.send_signal` on Windows only accepts `SIGTERM`,
`CTRL_C_EVENT`, and `CTRL_BREAK_EVENT` — anything else raises
`ValueError: Unsupported signal: 2`.

The failure has a second-order effect: the `raise` happens *inside* the
`finally:` block, on the line before `proc.communicate(timeout=10)`
(`tests/test_hook_bridge.py:163-165`). So the cleanup never runs and the child
process is orphaned. `snagline watch --follow` polls in a `time.sleep(0.2)` loop
with no exit condition (`src/snagline/cli.py:356-364`), so the orphan spins
until it is killed by hand. Every suite run on Windows leaks one.

CI does not surface this: `.github/workflows/ci.yml` runs the matrix on
`ubuntu-latest` only, where `SIGINT` is supported and the test passes. The
project ships a Windows-relevant integration (Claude Code hooks) and
`pyproject.toml` declares no platform restriction, so a Windows contributor hits
a red suite on a clean checkout.

Expected: the test terminates the child through a portable mechanism and asserts
the same behaviour on every platform, or is skipped with a clear reason where
the mechanism genuinely does not exist.

### Standalone code to reproduce the issue

```shell
# On Windows:
python -m pytest tests/test_hook_bridge.py::test_watch_file_follow_sees_appended_lines -q --no-cov

# Confirm the orphan afterwards:
python -c "import subprocess;print(subprocess.run(['tasklist','/fi','imagename eq python.exe'],capture_output=True,text=True).stdout)"
```

### Relevant log output

```shell
    finally:
>       proc.send_signal(signal.SIGINT)  # graceful: same as Ctrl-C on the CLI
[...]

PR #71, opened 2026-08-23
fix(claude-code): preserve tool latency through ingest_payload

### LINKED ISSUE #64: Claude Code adapter: ingest_payload discards every tool latency, silently disabling the CUSUM detector
### Issue type

Bug

Yes

### Source

source

### SNAGLINE version

0.1.0 (28fd79a)

### Custom code

No

### Python version

3.13.1

### Current behavior?

`ingest_payload` in `src/snagline/adapters/claude_code.py:163-171` calls
`tracker.note(payload)` **before** `payload_to_event(payload, tracker=tracker)`.
`HookTracker.note` pops the paired `PreToolUse` start timestamp on every
non-`PreToolUse` event (`src/snagline/adapters/claude_code.py:75`), so by the
time `payload_to_event` asks the tracker for the latency the start is already
gone and `HookTracker.latency_ms` returns `None`
(`src/snagline/adapters/claude_code.py:84-86`).

Two consequences, both on the documented path:

1. **`PostToolUse` always yields `latency_ms=None`.** The one event that has a
   real duration never carries it.
2. **`PreToolUse` yields `latency_ms=0.0`.** `payload_to_event` gates the
   tracker lookup on `is_tool` (`src/snagline/adapters/claude_code.py:134-136`),
   which is `True` for `PreToolUse` too. `note` has just written the start with
   the same clock reading, so `latency_ms` computes `(now - now) * 1000 == 0.0`.

The net effect is that `LatencyAnomalyDetector` builds its entire Welford
baseline out of synthetic `0.0` samples and never observes a single real tool
duration. The tier-1 CUSUM detector is silently dead for every Claude Code
user, on both supported bridges:

- `snagline hook` (the command hook) -> `_cmd_hook` -> `ingest_payload`
- `POST /hooks/claude-code` (the http hook) -> `_post_claude_hook`
  (`src/snagline/server/http_server.py:185-187`) -> `ingest_payload`

Expected: `PostToolUse` / `PostToolUseFailure` carry the measured wall-clock
duration of the tool call, and `PreToolUse` carries `latency_ms=None` (a tool
that has not run yet has no duration).

Why the existing test does not catch it:
`tests/adapters/test_claude_code_adapter.py:101-108` asserts
`ev.latency_ms == 250.0`, but it reaches that assertion by calling
`payload_to_event(...)` directly after a single `ingest_payload` for the
`PreToolUse` half. That ordering (`note` then read, on *different* payloads) is
not the ordering the shipped `ingest_payload` uses, so the test passes against
code that never produces a latency in production.

Related: the comment at `src/snagline/adapters/claude_code.py:76` says "drop
stale starts older than 10 minutes", but the code below it does
`self._starts.clear()` when the dict exceeds 256 entries. There is no age
check; the eviction is unconditional and drops fresh starts along with stale
ones. A busy session with more than 256 tools in flight therefore loses
latency for the in-flight calls as well.

```shell
cat > repro.py <<'PY'
from snagline.adapters.claude_code import HookTracker, ingest_payload, payload_to_event


class M:
    def __init__(self):
        self.events = []

    def ingest(self, e):
        self.events.append(e)


def payload(evt, tuid="toolu_01"):
    return {
        "session_id": "s1",
        "hook_event_name": evt,
        "tool_name": "Bash",
        "tool_input": {"command": "npm test"},
        "tool_use_id": tuid,
    }


# Part 1: latency is lost on the shipped ingest_payload path.
now = {"t": 0.0}
tracker = HookTracker(clock=lambda: now["t"])
m = M()
ingest_payload(m, payload("PreToolUse"), tracker)
now["t"] = 0.25  # the Bash call took 250 ms
ingest_payload(m, payload("PostToolUse"), tracker)
print("via ingest_payload (documented CLI hook + sidecar path):")
for e in m.events:
    print(f"  {e.action_type} tool={e.tool_name} latency_ms={e.latency_ms}")

now2 = {"t": 0.0}
t2 = HookTracker(clock=lambda: now2["t"])
t2.note(payload("PreToolUse"))
now2["t"] = 0.25
ev = payload_to_event(payload("PostToolUse"), tracker=t2)
print(f"if latency is read before note() pops the start:\n  post latency_ms={ev.latency_ms}")

# Part 2: the CUSUM detector therefore never fires, even on a 30-second call.
risks = []


class Rec:
    def emit(self, r):
        risks.append(r)


mon = Monitor.default(config=Config(), sinks=[Rec()])
clock = {"t": 0.0}
tr = HookTracker(clock=lambda: clock["t"])


def pair(i, duration_s):
    p = {
        "session_id": "s1",
        "tool_name": "Bash",
        "tool_use_id": f"toolu_{i}",
        "tool_input": {"command": f"step-{i}"},
    }
    ingest_payload(mon, {**p, "hook_event_name": "PreToolUse"}, tr)
    clock["t"] += duration_s
    ingest_payload(mon, {**p, "hook_event_name": "PostToolUse"}, tr)


for i in range(10):
    pair(i, 0.100)  # healthy baseline: 100 ms per call
pair(99, 30.0)      # then one 30-SECOND tool call
print()
print(f"latency risks emitted: {len([r for r in risks if r.trigger == 'latency_anomaly'])}")
print(f"all risks: {[r.trigger for r in risks]}")
PY
python repro.py
```

### Relevant log output

```shell
via ingest_payload (documented CLI hook + sidecar path):
  tool_call tool=Bash latency_ms=0.0
  tool_call tool=Bash latency_ms=None
if latency is read before note() pops the start:
  post latency_ms=250.0

latency risks emitted: 0
all risks: []
```

## Impact

`LatencyAnomalyDetector` is one of the three tier-1 detectors wired by
`Monitor.default`. For Claude Code users it never receives a usable sample, so
a tool that degrades from 100 ms to 30 s produces no risk at all. The failure
is silent: no warning, no metric, and the adapter's own test suite is green.
This affects the flagship documented integration (`README` quickstart, `docs`
hook setup, and `snagline serve`'s `/hooks/claude-code`).

## Proposed fix

1. In `ingest_payload`, map the event first and note the payload afterwards, or
   have `payload_to_event` read the latency before `note` mutates the tracker.
   The clean ordering is: read latency -> build event -> `note`.
2. Restrict the tracker lookup to the `"end"` subkind that `_EVENT_MAP` already
   distinguishes, so `PreToolUse` reports `latency_ms=None` instead of `0.0`.
3. Make the eviction in `note` match its comment: drop starts older than a TTL
   rather than clearing the whole dict.
4. Add a regression test that asserts the latency through `ingest_payload`
   (the shipped path), not through a hand-sequenced `payload_to_event` call.

abyyxhek:
## Summary of Changes

`ingest_payload` called `tracker.note(payload)` **before** mapping the payload.
`HookTracker.note` retires the paired `PreToolUse` start, so by the time
`payload_to_event` asked the tracker for a duration the start was already gone
and `latency_ms` came back `None`. The tracker lookup was gated on `is_tool`,
which is `True` for `PreToolUse` too, so that half reported a synthetic `0.0` ms
(the start had just been written with the same clock reading).

This PR:

- **Maps first, then notes**, with the `note` call in a `finally` so an unmapped
  or malformed payload still advances the tracker.
- **Restricts the latency lookup to the `"end"` subkind** that `_EVENT_MAP`
  already distinguishes, so `PreToolUse` reports `latency_ms=None` instead of
  `0.0`.
- **Replaces `_starts.clear()` with the age-based eviction the comment already
  promised.** The old policy dropped the whole dict once it passed 256 entries,
  discarding fresh in-flight starts along with abandoned ones. Now stale entries
  go by TTL first, and only if nothing is actually stale does it shed the oldest
  half.

Fixes #64

## Justification

`LatencyAnomalyDetector` is one of the three tier-1 detectors `Monitor.default`
wires up. Because every `PostToolUse` arrived with `latency_ms=None` and every
`PreToolUse` with `0.0`, the detector's entire Welford baseline was synthetic
zeros and it never observed a real tool duration. A tool degrading from 100 ms
to 30 s produced no risk at all.

This was silent and it hit the flagship documented integration on both bridges:

- `snagline hook` (command hook) -> `_cmd_hook` -> `ingest_payload`
- `POST /hooks/claude-code` (http hook) -> `_post_claude_hook` -> `ingest_payload`

## Kind of Contribution

Bug Fix, Tests

## Unit Tests

`tests/adapters/test_claude_code_adapter.py` gained five cases. The important
point about the existing coverage: `test_tracker_pairs_pre_and_post_for_latency`
asserted `250.0` ms, but reached it by calling `payload_to_event` directly after
a single `ingest_payload` for the `PreToolUse` half. That note-then-read-on-a-
different-payload ordering is not the ordering the shipped `ingest_payload`
used, so the test was green against code that never produced a latency in
production. It is kept (it still exercises `payload_to_event`), and the new
tests assert the shipped path:

- `test_ingest_payload_preserves_latency_on_the_shipped_path` — asserts
  `[None, 250.0]` through two `ingest_payload` calls.
- `test_pretooluse_carries_no_latency` — pins `None`, not `0.0`.
- `test_post_tool_use_failure_also_carries_latency` — a failing tool still took
  time; `PostToolUseFailure` is an `"end"` event too.
- `test_hook_latency_reaches_the_cusum_detector` — end to end through a real
  `Monitor.default`: 10 healthy 100 ms calls then one 30 s call must raise a
  `latency_anomaly`. This is the regression guard that would have caught the
  original bug.
- `test_tracker_evicts_stale_starts_by_age_not_wholesale` — a fresh in-flight
  start survives an eviction that clears abandoned ones.

## Execution Tests

Reproduction from #64, before and after.

Before:

```
via ingest_payload (documented CLI hook + sidecar path):
  tool_call tool=Bash latency_ms=0.0
  tool_call tool=Bash latency_ms=None
if latency is read before note() pops the start:
  post latency_ms=250.0

latency risks emitted: 0
all risks: []
```

After:

```
via ingest_payload (documented CLI hook + sidecar path):
  tool_call tool=Bash latency_ms=None
  tool_call tool=Bash latency_ms=250.0
if latency is read before note() pops the start:
  post latency_ms=250.0

latency risks emitted: 1
all risks: ['latency_anomaly']
```

Full suite, lint, and types:

```
$ python -m pytest -q
1 failed, 192 passed, 1 skipped in 22.67s
Required test coverage of 80% reached. Total coverage: 88.35%

All checks passed!
71 files already formatted
$ python -m mypy src/snagline
```

`send_signal(SIGINT)`, which is unsupported on Windows. Filed separately as #69;
it passes on the ubuntu CI matrix.

### DIFF
--- a/src/snagline/adapters/claude_code.py
+++ b/src/snagline/adapters/claude_code.py
@@ -27,6 +27,7 @@ signature, never ``metadata``.
 
 from __future__ import annotations
 
+import contextlib
 import json
 import time
 import uuid
@@ -60,22 +61,49 @@ class HookTracker:
     and defensive: malformed payloads never raise out of ``note``/``latency``.
     """
 
-    def __init__(self, clock=None) -> None:
+    def __init__(
+        self,
+        clock=None,
+        ttl_seconds: float = 600.0,
+        max_pending: int = 1024,
+    ) -> None:
         self._clock = clock or time.monotonic
         self._starts: dict[str, float] = {}
+        self._ttl = ttl_seconds
+        self._max_pending = max_pending
 
     def note(self, payload: dict) -> None:
+        """Record a ``PreToolUse`` start, or retire a paired start.
+
+        Call this *after* the payload has been mapped: it pops the start, so
+        reading the latency afterwards would always see ``None`` (issue #64).
+        """
         event = payload.get("hook_event_name")
         tool_use_id = payload.get("tool_use_id")
         if not isinstance(tool_use_id, str):
             return
+        now = self._clock()
         if event == "PreToolUse":
-            self._starts[tool_use_id] = self._clock()
+            self._starts[tool_use_id] = now
         else:
             self._starts.pop(tool_use_id, None)
-        # Bound memory: drop stale starts older than 10 minutes.
-        if len(self._starts) > 256:
-            self._starts.clear()
+        if len(self._starts) > self._max_pending:
+            self._evict(now)
+
+    def _evict(self, now: float) -> None:
+        """Bound memory by age, so a tool still running keeps its start.
+
+        A ``PreToolUse`` whose ``PostToolUse`` never arrives (crashed tool,
+        dropped hook) would otherwise leak. Drop those by age first; only if
+        nothing is actually stale do we shed the oldest half, which keeps the
+        newest -- still pairable -- starts.
+        """
+        cutoff = now - self._ttl
+        for key in [k for k, started in self._starts.items() if started < cutoff]:
+            del self._starts[key]
+        if len(self._starts) > self._max_pending:
+            by_age = sorted(self._starts.items(), key=lambda kv: kv[1])
+            self._starts = dict(by_age[len(by_age) // 2 :])
 
     def latency_ms(self, payload: dict) -> float | None:
         tool_use_id = payload.get("tool_use_id")
@@ -94,9 +122,12 @@ def payload_to_event(
 ) -> StepEvent | None:
     """Map one Claude Code hook payload to a ``StepEvent`` (or ``None``).
 
-    ``tracker`` is consulted for latency: call ``tracker.note(payload)`` on
-    every payload (including unmapped ones), then this function fills
-    ``latency_ms`` on ``PostToolUse`` events.
+    ``tracker`` is consulted for latency: map the payload with this function
+    first, then call ``tracker.note(payload)``. ``note`` retires the paired
+    ``PreToolUse`` start, so noting before mapping loses the latency entirely
+    (issue #64). ``latency_ms`` is filled only on the events that *end* a tool
+    call (``PostToolUse`` / ``PostToolUseFailure``); a ``PreToolUse`` has not
+    run yet and therefore has no duration.
     """
     mapped = _EVENT_MAP.get(str(payload.get("hook_event_name")))
     if mapped is None:
@@ -131,8 +162,13 @@ def payload_to_event(
             else (str(err) if err else payload.get("hook_event_name"))
         )
 
+    # Only the event that *ends* a tool call has a duration; PreToolUse fires
+    # before the tool runs, so asking the tracker there would read the start it
+    # just wrote and report a bogus 0.0 ms (issue #64).
     latency_ms = (
-        tracker.latency_ms(payload) if tracker is not None and is_tool else None
+        tracker.latency_ms(payload)
+        if tracker is not None and is_tool and subkind == "end"
+        else None
     )
 
     return StepEvent(
@@ -154,18 +190,25 @@ def payload_to_event(
 def ingest_payload(
     monitor: Any, payload: dict, tracker: HookTracker | None = None
 ) -> StepEvent | None:
-    """Convenience: note the payload on ``tracker`` and ingest the mapped event.
+    """Convenience: map the payload, ingest it, then note it on ``tracker``.
 
     Returns the event that was ingested, or ``None`` for unmapped events.
     Never raises: mapping/ingest failures are the Monitor's fail-open concern,
     and a bridge must not break the host process.
+
+    Order matters. ``tracker.note`` retires the ``PreToolUse`` start that
+    ``payload_to_event`` needs to compute ``latency_ms``, so mapping happens
+    first and the tracker is updated afterwards -- in a ``finally`` so an
+    unmapped or malformed payload still advances the tracker (issue #64).
     """
     try:
-        if tracker is not None:
-            tracker.note(payload)
         event = payload_to_event(payload, tracker=tracker)
         if event is not None:
             monitor.ingest(event)
         return event
     except Exception:
         return None
+    finally:
+        if tracker is not None:
+            with contextlib.suppress(Exception):
+                tracker.note(payload)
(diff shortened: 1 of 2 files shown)


Cyrax321 (approved):
Thank you for the contribution. This PR fixes the silent loss of tool latency in the Claude Code bridge where LatencyAnomalyDetector never saw real durations.

I confirmed on origin/master that ingest_payload called tracker.note before payload_to_event, so note popped the PreToolUse start before latency was read and every PostToolUse got None while PreToolUse reported synthetic 0.0 ms via the is_tool gate. The fix correctly maps first then notes in finally with suppress, restricts latency to subkind == "end" so PreToolUse reports None, and replaces the wholesale _starts.clear() with TTL-first eviction plus oldest-half shedding.

I applied the patch and ran tests/adapters/test_claude_code_adapter.py: 14 passed including ingest_payload preserves latency on the shipped path, PreToolUse none, PostToolUseFailure latency, end to end CUSUM firing on 30s spike, and fresh start surviving eviction.

No blocking issues found. Both documented bridges snagline hook and POST /hooks/claude-code are corrected.

Approved.

Cyrax321:
**Hana**: Approved! Hana says a garden grows because of gardeners like you. Thank you for contributing, and we hope to see your name here again soon.

PR #72, opened 2026-08-23
fix(batching): enforce max_batch and drain the queue on close

### LINKED ISSUE #65: BatchingSink: max_batch is never enforced and close() silently drops queued alerts
### Issue type

Bug

Yes

### Source

source

### SNAGLINE version

0.1.0 (28fd79a)

### Custom code

No

### Python version

3.13.1

### Current behavior?

`BatchingSink` has two independent defects in
`src/snagline/sinks/batching.py`.

**1. `max_batch` is never enforced.** The module docstring promises "a
background thread flushes the queue on an interval **or once it reaches
`max_batch`**" (`src/snagline/sinks/batching.py:6-8`). The constructor stores
`self._max_batch = max_batch` (`src/snagline/sinks/batching.py:36`) and nothing
ever reads it again: `emit` only appends
(`src/snagline/sinks/batching.py:45-48`), `_run` only waits on the interval
(`src/snagline/sinks/batching.py:50-52`), and `_flush` drains whatever is there
(`src/snagline/sinks/batching.py:54-60`). With `flush_interval=3600` and
`max_batch=2`, 50 emits deliver nothing and sit in an unbounded deque.

`max_batch` is also the only thing that would bound the queue's memory. As
written, a risk storm grows `self._queue` without limit until the next interval
tick.

**2. `close()` silently drops everything still queued.** `close`
(`src/snagline/sinks/batching.py:79-81`) sets `self._stop` and joins the
thread. Setting `_stop` makes `_run`'s `self._stop.wait(...)` return `True`
immediately, so the loop exits *without* a final `_flush()`. `flush_now` exists
and its docstring says it is "used at shutdown"
(`src/snagline/sinks/batching.py:75-77`), but `close` never calls it. Any risk
enqueued since the last interval tick is lost.

Expected: `emit` flushes once the queue reaches `max_batch` (so the documented
threshold works and the queue stays bounded), and `close()` drains the queue
before returning so a clean shutdown does not lose alerts.

Why the existing tests do not catch it: every test in
`tests/test_batching_sink.py` calls `sink.flush_now()` explicitly before
asserting (lines 37, 51-53, 69), which masks both defects. In particular
`test_emit_is_non_blocking_and_background_flushes` passes `max_batch=2` with
`flush_interval=0.05` and then asserts `len(inner.emitted) >= 1` — satisfied by
the interval tick alone, so it never exercises the threshold it appears to test.

```shell
cat > repro.py <<'PY'


def risk(i):
    return FailureRisk("ep", str(i), 0.9, "loop", "d", 1.0)


class Rec:
    def __init__(self):
        self.got = []

    def emit(self, r):
        self.got.append(r)


# 1. max_batch is documented to trigger a flush; it never does.
inner = Rec()
sink = BatchingSink(inner, max_batch=2, flush_interval=3600.0)
for i in range(50):
    sink.emit(risk(i))
print(f"max_batch=2, 50 emits -> delivered={len(inner.got)} (expected a flush at 2)")
print(f"queue still holds {len(sink._queue)} risks")

# 2. close() drops everything still queued.
inner2 = Rec()
sink2 = BatchingSink(inner2, max_batch=100, flush_interval=3600.0)
for i in range(5):
    sink2.emit(risk(i))
sink2.close()
print(f"after close(): delivered={len(inner2.got)} (expected 5)")
PY
python repro.py
```

### Relevant log output

```shell
max_batch=2, 50 emits -> delivered=0 (expected a flush at 2)
queue still holds 50 risks
after close(): delivered=0 (expected 5)
```

## Impact

`BatchingSink` is the recommended wrapper for the network sinks
(webhook/Slack/PagerDuty) precisely so alert delivery is decoupled from
`ingest()`. Both defects hit the alerting path:

- A process that shuts down within one `flush_interval` of detecting a failure
  never pages anyone. With the default `flush_interval=5.0`, that is a five
  second window on every clean exit — and for a short-lived agent run the
  detected risk may be the last thing that happens before shutdown.
- Because `max_batch` is dead, `flush_interval` is the only flush trigger, so
  the queue is unbounded between ticks and the configured batch size is a no-op
  the user cannot detect.

## Proposed fix

1. Check the queue length in `emit` and flush when it reaches `max_batch`.
   Delivery must stay off the caller's thread — signal the background thread
   (e.g. an `threading.Event` the `_run` loop waits on) rather than delivering
   inline, so `emit` keeps its non-blocking contract.
2. Have `close()` call `_flush()` after joining the thread (and make `_run`
   flush once on exit) so no queued risk is dropped.
3. Extend `tests/test_batching_sink.py` with two cases that do **not** call
   `flush_now()`: one asserting the `max_batch` threshold delivers on its own,
   one asserting `close()` drains the queue.

abyyxhek:
## Summary of Changes

Two independent defects in `src/snagline/sinks/batching.py`.

**`max_batch` was never enforced.** The constructor stored `self._max_batch` and
nothing read it again — `emit` only appended, `_run` only waited on the interval,
`_flush` drained whatever happened to be there. The documented threshold flush
never fired, so `flush_interval` was the only trigger and the queue was
unbounded between ticks.

Reaching `max_batch` now sets a `threading.Event` that the flusher waits on
alongside the interval. Delivery stays on the background thread — `emit` runs
inside `ingest()` on the host's thread and must not inherit the wrapped sink's
latency, so it signals rather than delivers inline.

**`close()` silently dropped everything still queued.** It set `_stop` and
joined; setting `_stop` made `_run`'s `self._stop.wait(...)` return `True`
immediately, so the loop exited *without* a final flush. `flush_now` existed and
its docstring said it was "used at shutdown", but `close` never called it.
`_run` now drains in a `finally`, and `close` flushes on the caller's thread as a
backstop if the join times out (`_flush` is a no-op on an empty queue, so the
double call is safe).

Fixes #65

## Justification

`BatchingSink` is the recommended wrapper for the network sinks
(webhook/Slack/PagerDuty), precisely so alert delivery is decoupled from
`ingest()`. Both defects landed on the alerting path:

- A process shutting down within one `flush_interval` of a detection never paged
  anyone. At the default `flush_interval=5.0` that is a five second window on
  every clean exit — and for a short-lived agent run, the detected risk may be
  the last thing that happens before shutdown.
- With `max_batch` dead, the configured batch size was a no-op the user had no
  way to observe, and the queue had no bound between interval ticks.

## Kind of Contribution

Bug Fix, Tests

## Unit Tests

`tests/test_batching_sink.py`. Worth stating why the existing suite was green:
every test called `sink.flush_now()` before asserting, which masks both defects.
`test_emit_is_non_blocking_and_background_flushes` looked like it covered the
threshold — it passes `max_batch=2` — but with `flush_interval=0.05` and an
assertion of `len(inner.emitted) >= 1`, the interval tick alone satisfies it.

The three new tests deliberately never call `flush_now()`:

- `test_max_batch_flushes_without_an_interval_tick` — `max_batch=3`,
  `flush_interval=3600`, 6 emits. The only thing that can deliver is the
  threshold.
- `test_close_drains_queued_risks` — `flush_interval=3600`, 5 emits, then
  `close()` alone must deliver all 5.
- `test_emit_stays_non_blocking_when_the_batch_is_full` — `max_batch=1` against a
  sink that sleeps 200 ms per emit; asserts the three `emit` calls return in
  under 100 ms total, then that all three arrive. This pins the reason the fix
  signals the background thread instead of flushing inline.

## Execution Tests

Reproduction from #65, with a bounded wait added for the background delivery
(the threshold flush is asynchronous by design, so the original synchronous
repro cannot observe it either way).

On `origin/master`:

```
max_batch=2, 50 emits -> delivered=0 (expected 50)
queue still holds 50 risks (expected 0)
after close(): delivered=0 (expected 5)
```

On this branch:

```
max_batch=2, 50 emits -> delivered=50 (expected 50)
queue still holds 0 risks (expected 0)
after close(): delivered=5 (expected 5)
```

Full suite, lint, and types:

```
$ python -m pytest -q
1 failed, 190 passed, 1 skipped in 21.21s
Required test coverage of 80% reached. Total coverage: 88.32%

All checks passed!
71 files already formatted
$ python -m mypy src/snagline
```

`send_signal(SIGINT)`, unsupported on Windows. Filed separately as #69; it
passes on the ubuntu CI matrix.

## Benchmark

Not applicable — no change to the `ingest()` path. `emit` gains one integer
comparison and, on the threshold, an `Event.set()`; both are off the detector
hot path and outside the Monitor's lock.

### DIFF
--- a/src/snagline/sinks/batching.py
+++ b/src/snagline/sinks/batching.py
@@ -8,7 +8,13 @@ or once it reaches ``max_batch``. An optional ``max_per_second`` rate limit
 paces the actual deliveries so a burst of risk does not trip the destination's
 rate limiter.
 
+``max_batch`` is what bounds the queue between interval ticks, so reaching it
+wakes the flusher immediately rather than delivering on the caller's thread --
+``emit`` stays non-blocking even when the wrapped sink is slow.
+
 Fail-open: a delivery error is swallowed (never blocks ingest or the queue).
+``close()`` drains the queue before returning, so a clean shutdown inside one
+``flush_interval`` of a detection still delivers the alert.
 """
 
 from __future__ import annotations
@@ -33,12 +39,15 @@ class BatchingSink:
         max_per_second: float | None = None,
     ) -> None:
         self._sink = sink
-        self._max_batch = max_batch
+        self._max_batch = max(1, max_batch)
         self._flush_interval = flush_interval
         self._min_gap = 1.0 / max_per_second if max_per_second else 0.0
         self._queue: collections.deque[FailureRisk] = collections.deque()
         self._lock = threading.Lock()
         self._stop = threading.Event()
+        # Set when the queue reaches ``max_batch`` (or on close) so the flusher
+        # can wake early instead of sitting out the rest of the interval.
+        self._wake = threading.Event()
         self._thread = threading.Thread(target=self._run, daemon=True)
         self._thread.start()
 
@@ -46,9 +55,25 @@ class BatchingSink:
         # Non-blocking enqueue; the background thread does the actual delivery.
         with self._lock:
             self._queue.append(risk)
+            full = len(self._queue) >= self._max_batch
+        if full:
+            # Threshold reached. Wake the flusher rather than delivering here:
+            # ``emit`` runs on the caller's thread (inside ``ingest``) and must
+            # stay non-blocking even when the wrapped sink is slow.
+            self._wake.set()
 
     def _run(self) -> None:
-        while not self._stop.wait(self._flush_interval):
+        try:
+            while not self._stop.is_set():
+                # Whichever comes first: the interval elapsing or a full batch.
+                self._wake.wait(self._flush_interval)
+                self._wake.clear()
+                if self._stop.is_set():
+                    break
+                self._flush()
+        finally:
+            # Drain anything enqueued before the stop so a clean shutdown does
+            # not silently lose alerts.
             self._flush()
 
     def _flush(self) -> None:
@@ -77,5 +102,11 @@ class BatchingSink:
         self._flush()
 
     def close(self) -> None:
+        """Stop the flusher and deliver whatever is still queued."""
         self._stop.set()
+        self._wake.set()  # unblock the interval wait so the drain happens now
         self._thread.join(timeout=self._flush_interval + 1.0)
+        # If the thread did not finish its drain within the join timeout, flush
+        # on the caller's thread. ``_flush`` is a no-op on an empty queue, so
+        # this is safe either way.
+        self._flush()
--- a/tests/test_batching_sink.py
+++ b/tests/test_batching_sink.py
@@ -72,3 +72,63 @@ def test_rate_limit_paces_delivery():
         assert len(inner.emitted) == 3
     finally:
         sink.close()
+
+
+def _wait_for(predicate, timeout: float = 5.0) -> None:
+    deadline = time.monotonic() + timeout
+    while time.monotonic() < deadline:
+        if predicate():
+            return
+        time.sleep(0.01)
+
+
+def test_max_batch_flushes_without_an_interval_tick():
+    # Issue #65: max_batch was stored and never read, so flush_interval was the
+    # only trigger and the queue was unbounded between ticks. Deliberately no
+    # flush_now() here -- an interval tick is an hour away, so the only thing
+    # that can deliver is the max_batch threshold.
+    inner = _RecordingSink()
+    sink = BatchingSink(inner, max_batch=3, flush_interval=3600.0)
+    try:
+        for i in range(6):
+            sink.emit(_risk(i))
+        _wait_for(lambda: len(inner.emitted) == 6)
+        assert len(inner.emitted) == 6
+    finally:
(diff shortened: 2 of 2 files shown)


Cyrax321 (approved):
Thank you for the contribution. This PR fixes two independent defects in BatchingSink where max_batch was stored but never read and close() dropped queued risks without a final flush.

I verified on origin/master that emit only appended, _run only waited on the interval, and close set _stop and joined causing _run to exit without flushing. The fix correctly introduces a _wake Event set when the queue reaches max_batch, makes _run wait on _wake with interval as timeout and drain in finally, and makes close set _stop and _wake then join and flush on the caller thread as a backstop. Delivery stays on the background thread so emit remains non-blocking.

I applied the patch and ran tests/test_batching_sink.py: 6 passed including the three new tests that deliberately avoid flush_now, strong coverage for threshold without interval tick, close drains, and emit stays non-blocking against a slow sink.

No blocking issues found. The shape is the smallest correct fix for the alerting path.

Approved.

Cyrax321:
**Rin**: Nyaa~ approved! Rin inspected everything and found it worthy. Thank you for contributing; quality work like this makes the whole project better.

PR #74, opened 2026-08-23
fix(state): release a finished episode's lock in end_episode

### LINKED ISSUE #67: MemoryStateBackend leaks one RLock per episode; end_episode never releases it
### Issue type

Bug

Yes

### Source

source

### SNAGLINE version

0.1.0 (28fd79a)

### Custom code

No

### Python version

3.13.1

### Current behavior?

`MemoryStateBackend` allocates one `threading.RLock` per `episode_id` and never
releases it (`src/snagline/state.py:36-51`):

```python
self._locks: dict[str, threading.RLock] = {}
...
lock = self._locks.get(episode_id)
if lock is None:
    lock = threading.RLock()
    self._locks[episode_id] = lock
```

There is no eviction, no TTL, and no public method to discard an episode's lock.
`Monitor.end_episode` is the documented teardown hook and its docstring says
"without it, per-episode state would accumulate for the life of the Monitor"
(`src/snagline/monitor.py:156-165`) — but it only calls `detector.reset(...)`
(`src/snagline/monitor.py:166-176`). It never tells the state backend the
episode is finished, because the `StateBackend` protocol
(`src/snagline/state.py:27-33`) exposes nothing to tell it with.

So after N completed episodes the backend still holds N locks, and calling
`end_episode` — the very API that exists to prevent accumulation — does not
reduce that count.

Expected: an episode that has been ended releases its lock, so a long-lived
`Monitor` has bounded memory in the number of *concurrent* episodes rather than
the number of *lifetime* episodes.

Note the shape of the leak is small per entry but unbounded in count, and it
lands on exactly the deployments the state backend was introduced for: the P1
sharded-lock design (`src/snagline/state.py:1-13`) exists so a
"high-throughput service" can run many episodes through one `Monitor`. A sidecar
(`snagline serve`) keys `episode_id` off Claude Code's `session_id`, so the
dictionary grows with every session the sidecar has ever seen and only resets
when the process restarts.

```shell
cat > repro.py <<'PY'
from snagline.state import MemoryStateBackend

backend = MemoryStateBackend()
mon = Monitor.default(config=Config(), sinks=[], state_backend=backend)
for i in range(5000):
    ep = f"episode-{i}"
    mon.ingest(
        StepEvent(
            step_id="s",
            episode_id=ep,
            timestamp=1.0,
            action_type="tool_call",
            action_signature=make_signature("tool_call", "Bash", "x"),
        )
    )
    mon.end_episode(ep)  # the documented teardown hook

print("episodes started and ended: 5000")
print(f"locks retained by MemoryStateBackend: {len(backend._locks)}")
print(
    "release/discard API on the backend: "
    f"{[n for n in dir(backend) if 'release' in n or 'discard' in n or 'forget' in n] or 'none'}"
)
PY
python repro.py
```

### Relevant log output

```shell
episodes started and ended: 5000
locks retained by MemoryStateBackend: 5000
release/discard API on the backend: none
```

## Impact

Unbounded growth in a long-lived process, keyed by a value the operator does not
control (an upstream session id). It contradicts `end_episode`'s documented
purpose, and it is invisible: no metric on `Monitor.metrics()` reports it, so an
operator sees resident memory creep with no attribution.

`snagline serve` is the concrete case — one `Monitor` for the process lifetime,
one `episode_id` per Claude Code session.

## Proposed fix

Add an optional `release(episode_id)` (or `discard_episode`) method to the
`StateBackend` protocol, implement it on `MemoryStateBackend` by popping the
lock under `self._meta`, and have `Monitor.end_episode` call it after the
detector resets — guarded with `getattr` / `hasattr` so a third-party backend
that does not implement it (and `RedisStateBackend`, whose locks are already
short-lived and self-expiring) keeps working unchanged. The pop must happen
after the `with lock:` block releases, so the discard cannot race a concurrent
`ingest` that is already holding the lock.

Add a regression test asserting the lock count returns to zero after
`end_episode`.

abyyxhek:
## Summary of Changes

`MemoryStateBackend.episode_lock` allocates one `threading.RLock` per episode id
and stores it in `self._locks`. Nothing ever removed an entry. `end_episode` --
documented as the teardown hook that exists because "without it, per-episode
state would accumulate for the life of the Monitor" -- reset every detector but
left the backend's dict untouched, so the Monitor kept a lock for every episode
id it had ever seen.

This PR adds an optional `release(episode_id)`:

- `MemoryStateBackend.release` pops the entry under `self._meta`.
- `RedisStateBackend.release` is an explicit no-op with a comment saying why: its
  locks are per-acquisition and expire on their own, so a finished episode leaves
  nothing behind to discard.
- `Monitor.end_episode` probes for the method with `getattr` and calls it inside
  the same fail-open guard the detector resets use.

The capability is documented as a `ReleasableStateBackend(StateBackend, Protocol)`
rather than being added to `StateBackend` itself, so a third-party backend written
against the narrower protocol stays valid under mypy and keeps working unchanged.

The release happens *inside* the `with self._state.episode_lock(episode_id):`
block, not after it. Holding the episode lock means no other thread can be in that
episode's critical section at the moment the entry is dropped. A thread that
ingests for the same id after the release simply allocates a fresh lock, which is
correct: an episode that has ended has no detector state left to serialize.

Fixes #67

## Justification

`MemoryStateBackend` is the default backend, so this is the shipped path for every
embedded use, `snagline serve`, and `snagline hook`.

The leak is keyed on *episode id*, and the long-lived deployments are exactly the
ones that see a fresh id per run. The sidecar handles `POST /events` for whatever
episode a caller names; the Claude Code hook bridge is a long-lived process
watching many short sessions. Each finished run adds one permanent dict entry plus
one `RLock`, and `end_episode` -- the one API that could have cleaned it up, called
from adapter `finally` blocks precisely for that purpose -- was the call that did
not.

It is also invisible to the caller: the leak is in a private dict on a backend the
user never touches, and there is no metric or log that reflects it.

## Kind of Contribution

Bug Fix, Tests

## Unit Tests

`tests/test_state_backend.py`. Four of the six new cases fail on `origin/master`
and pass here:

- `test_end_episode_releases_the_episode_lock` -- the regression guard. 500
  ingest/`end_episode` pairs with distinct ids, then `backend._locks == {}`.
- `test_release_keeps_other_episodes_untouched` -- releasing `"a"` leaves `"b"`'s
  lock object identical, not merely present.
- `test_release_of_an_unknown_episode_is_a_no_op`.
- `test_end_episode_propagates_release_error_when_not_fail_open`.

Two cover the compatibility contract:

- `test_end_episode_tolerates_a_backend_without_release` -- a backend implementing
  only `episode_lock` must not hit `AttributeError`. It passes on master too; it is
  there to pin that this PR did not make `release` mandatory.
- `test_end_episode_is_fail_open_when_release_raises` -- a backend whose `release`
  throws is logged, not propagated, matching the detector-reset behaviour beside
  it.

The existing `test_memory_backend_isolates_episode_locks` and
`test_concurrent_ingest_of_distinct_episodes_no_deadlock` pass unchanged.

## Execution Tests

Reproduction from #67: 5000 episodes, each ingested and then ended.

On `origin/master`:

```
episodes started and ended : 5000
locks still held by backend: 5000 (expected 0)
bytes retained (approx)    : 103856
VERDICT: LEAK
```

On this branch:

```
episodes started and ended : 5000
locks still held by backend: 0 (expected 0)
bytes retained (approx)    : 184
VERDICT: clean
```

The byte figure is `sys.getsizeof(backend._locks)` -- the dict's own table only,
excluding the 5000 `RLock` objects and 5000 key strings it keeps alive.

Full suite, lint, and types:

```
$ python -m pytest -q
1 failed, 193 passed, 1 skipped in 18.69s
Required test coverage of 80% reached. Total coverage: 88.24%

All checks passed!
71 files already formatted
$ python -m mypy src/snagline
```

`send_signal(SIGINT)`, unsupported on Windows. Filed separately as #69; it passes
on the ubuntu CI matrix.

## Benchmark

Not applicable -- no change to the `ingest()` path. `end_episode` gains one
`getattr` and one dict `pop` under the existing `_meta` lock, both off the
per-step hot path.

### DIFF
--- a/src/snagline/monitor.py
+++ b/src/snagline/monitor.py
@@ -157,9 +157,13 @@ class Monitor:
         """Signal that ``episode_id`` has finished; clear its per-episode state.
 
         Calls ``reset(episode_id)`` on every detector so loop windows, CUSUM
-        baselines, and cascade counters for that episode are dropped. This is
-        the teardown hook adapters call when an agent run completes; without it,
+        baselines, and cascade counters for that episode are dropped, then asks
+        the state backend to release whatever it holds for that id. This is the
+        teardown hook adapters call when an agent run completes; without it,
         per-episode state would accumulate for the life of the Monitor.
+
+        Backends are only asked to release if they implement it -- a backend
+        written against the narrower ``StateBackend`` protocol is unaffected.
 
         Fail-open: a detector's ``reset`` exception is logged, never propagated.
         """
@@ -171,6 +175,19 @@ class Monitor:
                     self._log_fault_once(
                         f"detector {getattr(detector, 'name', repr(detector))} "
                         "reset raised; ignoring (fail-open)"
+                    )
+                    if not self._fail_open:
+                        raise
+            # Inside the lock: no other thread can be in this episode's
+            # critical section, so dropping the entry cannot strand a holder.
+            release = getattr(self._state, "release", None)
+            if callable(release):
+                try:
+                    release(episode_id)
+                except Exception:
+                    self._log_fault_once(
+                        f"state backend {type(self._state).__name__} release "
+                        "raised; ignoring (fail-open)"
                     )
                     if not self._fail_open:
                         raise
--- a/src/snagline/state.py
+++ b/src/snagline/state.py
@@ -33,6 +33,19 @@ class StateBackend(Protocol):
         ...
 
 
+class ReleasableStateBackend(StateBackend, Protocol):
+    """A ``StateBackend`` that can drop what it holds for a finished episode.
+
+    Optional capability: ``Monitor.end_episode`` probes for ``release`` and
+    skips backends that do not expose it, so a backend written against the
+    narrower ``StateBackend`` above keeps working unchanged.
+    """
+
+    def release(self, episode_id: str) -> None:
+        """Discard any per-episode state held for ``episode_id``."""
+        ...
+
+
 class MemoryStateBackend:
     """Process-local backend: one re-entrant lock per episode id."""
 
@@ -49,6 +62,22 @@ class MemoryStateBackend:
                 self._locks[episode_id] = lock
         with lock:
             yield
+
+    def release(self, episode_id: str) -> None:
+        """Drop the lock allocated for a finished episode.
+
+        Without this the dict grows by one entry per episode id and never
+        shrinks, so a long-lived Monitor watching many short runs retains a
+        lock for every episode it has ever seen.
+
+        ``Monitor.end_episode`` calls this while holding the episode lock, so
+        no other thread can be inside the critical section when the entry is
+        dropped. A thread that ingests for the same id *after* the release
+        simply allocates a fresh lock -- correct, because an episode that has
+        ended has no detector state left to serialize.
+        """
+        with self._meta:
+            self._locks.pop(episode_id, None)
 
 
 class RedisStateBackend:  # pragma: no cover - optional, requires redis
@@ -69,6 +98,13 @@ class RedisStateBackend:  # pragma: no cover - optional, requires redis
         finally:
             if acquired:
                 lock.release()
+
+    def release(self, episode_id: str) -> None:
+        """No-op: Redis locks are per-acquisition and expire on their own.
+
+        Nothing is retained between ``episode_lock`` calls, so a finished
+        episode leaves nothing behind to discard.
+        """
 
 
 def default_state_backend() -> StateBackend:
--- a/tests/test_state_backend.py
+++ b/tests/test_state_backend.py
@@ -3,7 +3,11 @@
 from __future__ import annotations
 
 import threading
+from contextlib import contextmanager
 
+import pytest
+
+from snagline.config import Config
 from snagline.events import StepEvent, make_signature
 from snagline.monitor import Monitor
 from snagline.state import (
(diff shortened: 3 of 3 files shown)


Cyrax321 (approved):
Thank you for the contribution. This PR correctly fixes the unbounded lock leak in MemoryStateBackend where end_episode reset detectors but kept the episode RLock forever.

I reviewed src/snagline/state.py and src/snagline/monitor.py on origin/master, verified the leak was introduced with the pluggable StateBackend (2cac357) and carried forward. The fix adds a ReleasableStateBackend protocol, implements MemoryStateBackend.release() popped under _meta while holding the episode lock so no holder is stranded, and makes Monitor.end_episode probe with getattr for backward compatibility. RedisStateBackend correctly no-ops.

I applied the patch locally and ran tests/test_state_backend.py: 10 passed including the new leak guard (500 episodes leave {}), isolation and fail-open tests. Existing suite on master remains 189 passed with one pre-existing unrelated failure.

No blocking issues found. Design is the best available shape at the ownership boundary. Marking as ready to pull.

Thank you again for the fix.

Cyrax321:
**Mei**: Mei raises her cup to you tonight. Thank you for the patience and thoroughness in this PR. Reviewers notice contributors like you.

PR #75, opened 2026-08-23
fix(server): require the auth token on GET, and expose it from the CLI

### LINKED ISSUE #68: Sidecar: GET /metrics and GET /risks bypass auth_token, and 'snagline serve' cannot set the token at all
### Issue type

Bug

Yes

### Source

source

### SNAGLINE version

0.1.0 (28fd79a)

### Custom code

No

### Python version

3.13.1

### Current behavior?

The sidecar's `auth_token` protects `POST` only. `do_GET`
(`src/snagline/server/http_server.py:76-84`) never calls `_authorized()`, while
`do_POST` does (`src/snagline/server/http_server.py:87-89`). So with a token
configured, `GET /metrics` and `GET /risks` are served to any unauthenticated
caller.

The module docstring states the intent unambiguously: "missing or wrong tokens
get 401. **`GET /health` stays open** for liveness probes"
(`src/snagline/server/http_server.py:16-20`) — naming exactly one open endpoint.
`/metrics` and `/risks` were meant to be behind the token.

`GET /risks` is the material one. It returns the accumulated bodies posted to
`POST /risks` verbatim (`src/snagline/server/http_server.py:81-82, 119`), which
are `FailureRisk` payloads from a remote `WebhookSink`. Those carry
`episode_id`, `trigger`, `score`, and `detail` — and `detail` is
detector-formatted operational text such as the repeated tool signature. So the
one endpoint that accumulates cross-agent operational data is the one with no
auth check at all. `GET /metrics` similarly exposes ingest volume and fault
counts.

A second, related gap: **the token cannot be set from the CLI.** `_cmd_serve`
(`src/snagline/cli.py:537-554`) calls `serve(...)` with only `host` and `port`,
and the `serve` subparser defines no `--auth-token` or `--max-body-bytes` flag
(no occurrence of either anywhere in `src/snagline/cli.py`). `serve()` accepts
both (`src/snagline/server/http_server.py:229-235`) and both default to
"unauthenticated / 1 MB". So a user following the documented
`snagline serve` path runs an unauthenticated sidecar with no way to change
that short of writing their own Python entrypoint.

Expected: `do_GET` authorizes `/metrics` and `/risks` while leaving `/health`
open, and `snagline serve` exposes the token so the documented feature is
reachable.

Also worth noting while in this file: `snagline_risks` is an unbounded list
(`src/snagline/server/http_server.py:210`, appended at line 119). A remote
sink posting risks to a long-lived sidecar grows it without limit. Flagging it
here rather than filing separately since it is the same endpoint; happy to split
it out if preferred.

```shell
cat > repro.py <<'PY'
import json
import threading
import urllib.error
import urllib.request

from snagline.server.http_server import make_server

mon = Monitor.default(config=Config(), sinks=[])
srv = make_server(mon, host="127.0.0.1", port=8791, auth_token="s3cr3t")
threading.Thread(target=srv.serve_forever, daemon=True).start()
base = "http://127.0.0.1:8791"


def call(method, path, body=None, token=None):
    data = json.dumps(body).encode() if body is not None else None
    req = urllib.request.Request(
        base + path, data=data, method=method,
        headers={"Content-Type": "application/json"},
    )
    if token:
        req.add_header("Authorization", f"Bearer {token}")
    try:
        with urllib.request.urlopen(req, timeout=3) as r:
            return r.status, r.read().decode().strip()
    except urllib.error.HTTPError as e:
        return e.code, e.read().decode().strip()


# Seed one risk through the authenticated POST so /risks has content.
print("POST /risks  (no token) ->", call("POST", "/risks", {"trigger": "loop"}))
print("POST /risks  (token)    ->", call(
    "POST", "/risks",
    {"episode_id": "prod-session-42", "trigger": "loop",
     "detail": "Bash(deploy --prod) repeated 5x", "score": 0.9},
    token="s3cr3t",
))
print()
print("GET  /health  (no token) ->", call("GET", "/health"))
print("GET  /metrics (no token) ->", call("GET", "/metrics"))
print("GET  /risks   (no token) ->", call("GET", "/risks"))
srv.shutdown()
srv.server_close()
PY
python repro.py
```

### Relevant log output

```shell
[sidecar] RECEIVED risk -> trigger=loop score=0.9 detail=Bash(deploy --prod) repeated 5x
POST /risks  (no token) -> (401, '{"error": "unauthorized"}')
POST /risks  (token)    -> (202, '{"status": "received", "trigger": "loop"}')

GET  /health  (no token) -> (200, '{"status": "ok"}')
GET  /metrics (no token) -> (200, '{"events_ingested": 0, "risks_emitted": 0, "detector_errors": 0, "sink_errors": 0}')
GET  /risks   (no token) -> (200, '{"risks": [{"episode_id": "prod-session-42", "trigger": "loop", "detail": "Bash(deploy --prod) repeated 5x", "score": 0.9}]}')
```

The last line is the point: the unauthenticated read returns the full risk body
that the 401 on the same path was protecting.

## Current mitigations, and why they do not close it

The docstring recommends fronting the sidecar with a reverse proxy
(`src/snagline/server/http_server.py:22-25`), and the default bind is
`127.0.0.1` (`src/snagline/server/http_server.py:218`). Both help, but neither
makes this a non-issue: `auth_token` is offered as the in-library control, a
user who sets it reasonably believes the read endpoints are covered, and the
default bind is overridable via `--host` for the container/sidecar deployment
the feature is aimed at.

## Proposed fix

1. Call `_authorized()` in `do_GET` for every path except `/health`, returning
   401 the same way `do_POST` does.
2. Add `--auth-token` and `--max-body-bytes` to the `serve` subparser and thread
   them into the `serve(...)` call in `_cmd_serve`. Read the token from
   `SNAGLINE_SERVE_AUTH_TOKEN` as well so it need not appear in the process
   argv.
3. Bound `snagline_risks` with a `collections.deque(maxlen=...)`.
4. Tests: assert 401 on `GET /metrics` and `GET /risks` without a token, 200
   with it, and 200 on `GET /health` either way.

abyyxhek:
## Summary of Changes

Three related defects in the sidecar, all on the "is this port safe to expose"
question.

**`do_GET` never called `_authorized()`.** `do_POST` did, so with `auth_token` set
`POST /events` was correctly refused with 401 -- while `GET /metrics` and
`GET /risks` answered anyone who could reach the port. `GET /risks` returns every
risk the sidecar has received, with episode ids, triggers, and detail strings.
Auth now runs before the path dispatch, so the 404 fallthrough is behind the token
too and an unauthenticated caller cannot enumerate which paths exist.
`GET /health` stays open -- the module docstring already promised that, and a
liveness probe (k8s, ELB, docker healthcheck) generally cannot be taught to carry
a shared secret.

**No CLI flag reached `auth_token`.** `serve()` and `make_server()` have accepted
`auth_token` and `max_body_bytes` since the endpoint landed, but `_cmd_serve`
called `serve(monitor, host=..., port=...)` and nothing else, so `snagline serve`
could not be secured from the command line at all. Adds `--auth-token`,
`--max-body-bytes`, and `--max-risks`. The token also reads from
`$SNAGLINE_SERVE_AUTH_TOKEN`, so it need not appear in argv where any other
process on the box can read it out of `ps`; the flag wins when both are set. When
no token is configured, `serve` now says so on stderr rather than starting an open
port silently.

**`POST /risks` appended to an unbounded list.** `_Handler.snagline_risks = []`,
appended once per received risk, never trimmed. It is now a
`collections.deque(maxlen=max_risks)` (default 1000), and `GET /risks` wraps it in
`list(...)` so the response stays JSON-serializable.

Fixes #68

## Justification

The sidecar exists to be reachable: it is the integration path for non-Python
agents, and the docstring's own framing is that the body cap and the token
"keep the sidecar safe to expose." The two auth defects meet in the worst place --
the endpoints that were open are the ones that read data out, and the flag that
would have closed them did not exist, so an operator who set out to secure
`snagline serve` had no way to do it and no signal that anything was wrong.

`/metrics` also discloses `detector_errors` and `sink_errors`, which is exactly the
information that tells an observer whether the monitoring is degraded.

The retention leak is smaller but the same shape: `POST /risks` is documented as
lenient and fail-open so remote senders never retry, which means the one thing it
must not do is trust the sender to stop.

## Kind of Contribution

Bug Fix, Tests

## Unit Tests

`tests/test_http_server.py` and `tests/test_cli.py`. Eight new or updated cases
fail on `origin/master`:

- `test_get_metrics_requires_token_when_configured` -- 401 with no token, 401 with
  a wrong token, 200 with `Authorization: Bearer`, 200 with `X-Snagline-Token`.
  Mirrors the existing `test_post_requires_token_when_configured`, which is the
  test that made the GET gap visible by contrast.
- `test_get_risks_requires_token_when_configured` -- posts a risk with the token,
  then asserts an unauthenticated `GET /risks` is 401 and an authenticated one
  actually returns the episode id.
- `test_unknown_get_path_is_401_not_404_when_token_configured` -- pins the
  ordering: unauthenticated probes get 401, not a 404 that confirms the path is
  unused.
- `test_received_risks_are_bounded` -- `max_risks=3`, 5 posts, asserts the
  retained step ids are exactly `["2", "3", "4"]`: oldest dropped, newest kept,
  and still serializable through `GET /risks`.
- `test_main_serve_starts_server` -- updated. It previously declared
  `_fake_serve(monitor, host, port)`, which is why the missing flags went
  unnoticed: any extra kwarg would have been a `TypeError`. It now captures
  `**kwargs` and asserts the defaults.
- `test_main_serve_passes_auth_token_and_limits`,
  `test_main_serve_reads_auth_token_from_the_environment`, and
  `test_main_serve_flag_beats_the_environment` -- the flag/env plumbing.

`test_get_endpoints_stay_open_when_no_token_is_configured` passes both before and
after; it is there to pin that this does not change the default no-token
deployment.

## Execution Tests

Reproduction from #68 against a live `make_server(..., auth_token="s3cret")`.

On `origin/master`:

```
token is set on the server: 's3cret'

GET  /health   no token -> 200 (expected 200)  ok
GET  /metrics  no token -> 200 (expected 401)  LEAK
     body: {"events_ingested": 0, "risks_emitted": 0, "detector_errors": 0, "sink_errors": 0}
GET  /risks    no token -> 200 (expected 401)  LEAK
     body: {"risks": []}
POST /risks    no token -> 401 (expected 401)  ok

posted 1500 risks -> GET /risks returns 1500 (bounded default is 1000)
```

On this branch:

```
token is set on the server: 's3cret'

GET  /health   no token -> 200 (expected 200)  ok
GET  /metrics  no token -> 401 (expected 401)  ok
GET  /risks    no token -> 401 (expected 401)  ok
POST /risks    no token -> 401 (expected 401)  ok

with token:
GET  /metrics  token    -> 200 (expected 200)
GET  /risks    token    -> 200 (expected 200)

posted 1500 risks -> GET /risks returns 1000 (bounded default is 1000)
```

Full suite, lint, and types:

```
$ python -m pytest -q
1 failed, 195 passed, 1 skipped in 23.24s
Required test coverage of 80% reached. Total coverage: 88.27%

All checks passed!
71 files already formatted
$ python -m mypy src/snagline
```

`send_signal(SIGINT)`, unsupported on Windows. Filed separately as #69; it passes
on the ubuntu CI matrix.

## Benchmark

Not applicable -- no change to the `ingest()` path. `do_GET` gains one header
comparison per request, and only when a token is configured.

## Summary by CodeRabbit

* **New Features**
  * Added optional authentication for HTTP server endpoints, with support for CLI flags and environment configuration.
  * Added configurable request-body size and risk-retention limits.
  * Kept the health check publicly accessible while protecting other endpoints.
  * Limited risk history to the newest configured entries.
  * Added clearer startup warnings when authentication is disabled.

### DIFF
--- a/src/snagline/cli.py
+++ b/src/snagline/cli.py
@@ -13,6 +13,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import os
 import sys
 import time
 from collections.abc import Iterator
@@ -243,6 +244,27 @@ def _build_parser() -> argparse.ArgumentParser:
         "--host", default="127.0.0.1", help="Bind address [default: 127.0.0.1]."
     )
     p_serve.add_argument("--port", type=int, default=8787, help="Port [default: 8787].")
+    p_serve.add_argument(
+        "--auth-token",
+        default=None,
+        help="Shared secret required on every request except GET /health "
+        "(Authorization: Bearer <token> or X-Snagline-Token). Falls back to "
+        "$SNAGLINE_SERVE_AUTH_TOKEN, which keeps the secret out of argv. "
+        "Unset means all endpoints are open.",
+    )
+    p_serve.add_argument(
+        "--max-body-bytes",
+        type=int,
+        default=1_000_000,
+        help="Reject POST bodies larger than this with 413 [default: 1000000].",
+    )
+    p_serve.add_argument(
+        "--max-risks",
+        type=int,
+        default=1000,
+        help="How many risks received on POST /risks to retain for GET /risks "
+        "[default: 1000].",
+    )
     p_serve.add_argument(
         "--sink",
         choices=["console", "webhook", "slack", "pagerduty"],
@@ -541,15 +563,27 @@ def _cmd_serve(args: argparse.Namespace) -> int:
         sinks = _build_sinks(args)
     except SystemExit as exc:
         return int(exc.code or 2)
+    # Prefer the flag, fall back to the environment so the secret need not
+    # appear in argv (visible to every other process via ps).
+    auth_token = args.auth_token or os.environ.get("SNAGLINE_SERVE_AUTH_TOKEN") or None
     print(
         f"snagline serve: listening on http://{args.host}:{args.port} (POST /events, GET /health)",
         file=sys.stderr,
     )
+    if not auth_token:
+        print(
+            "snagline serve: no --auth-token / $SNAGLINE_SERVE_AUTH_TOKEN set; "
+            "every endpoint is open to anyone who can reach this port",
+            file=sys.stderr,
+        )
     with suppress(KeyboardInterrupt):
         serve(
             Monitor.default(config=_build_config(args), sinks=sinks),
             host=args.host,
             port=args.port,
+            auth_token=auth_token,
+            max_body_bytes=args.max_body_bytes,
+            max_risks=args.max_risks,
         )
     return 0
 
--- a/src/snagline/server/http_server.py
+++ b/src/snagline/server/http_server.py
@@ -13,11 +13,15 @@ POST bodies are capped at ``max_body_bytes`` (default 1 MB); larger payloads
 get 413. This keeps the sidecar safe to expose and able to absorb buffered
 telemetry from any host.
 
-Optionally protect the POST endpoints with a shared secret by passing
-``auth_token=`` to ``make_server``/``serve``. When set, POSTs must carry the
+Optionally protect the endpoints with a shared secret by passing
+``auth_token=`` to ``make_server``/``serve``. When set, requests must carry the
 token via ``Authorization: Bearer <token>`` or ``X-Snagline-Token: <token>``;
 missing or wrong tokens get 401. ``GET /health`` stays open for liveness
-probes.
+probes -- everything else, GET and POST alike, is behind the token.
+
+``POST /risks`` retains the most recent ``max_risks`` risks (default 1000) for
+``GET /risks``; older ones are discarded so an unbounded sender cannot grow the
+sidecar's memory without limit.
 
 No framework, no third-party dependency -- this keeps the zero-dependency
 principle intact even for server mode. For high-throughput production use,
@@ -34,6 +38,7 @@ from __future__ import annotations
 import json
 import logging
 import sys
+from collections import deque
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
 from typing import Any
 
@@ -48,6 +53,7 @@ def make_handler(
     monitor: Monitor,
     auth_token: str | None = None,
     max_body_bytes: int = 1_000_000,
+    max_risks: int = 1000,
 ) -> type[BaseHTTPRequestHandler]:
     """Build a request-handler class bound to ``monitor``."""
     tracker = HookTracker()
@@ -74,12 +80,21 @@ def make_handler(
             return self.headers.get("X-Snagline-Token") == token
 
         def do_GET(self) -> None:  # noqa: N802 - http.server naming
+            # /health is deliberately open: a liveness probe (k8s, ELB, docker
+            # healthcheck) generally cannot be taught to carry a shared secret,
+            # and it reveals nothing but reachability.
             if self.path == "/health":
                 self._respond(200, {"status": "ok"})
-            elif self.path == "/metrics":
+                return
+            # Everything else is behind the token, including the 404 fallthrough
+            # so an unauthenticated caller cannot probe which paths exist.
+            if not self._authorized():
+                self._respond(401, {"error": "unauthorized"})
+                return
+            if self.path == "/metrics":
                 self._respond(200, self.snagline_monitor.metrics())
             elif self.path == "/risks":
-                self._respond(200, {"risks": self.snagline_risks})
+                self._respond(200, {"risks": list(self.snagline_risks)})
             else:
                 self._respond(404, {"error": "not found"})
 
@@ -207,7 +222,9 @@ def make_handler(
 
     _Handler.snagline_monitor = monitor
     _Handler.snagline_tracker = tracker
-    _Handler.snagline_risks = []
+    # Bounded: POST /risks is an open-ended ingest point, so retain only the
+    # most recent max_risks entries rather than growing without limit.
+    _Handler.snagline_risks = deque(maxlen=max(1, max_risks))
     _Handler.snagline_auth = auth_token
     _Handler.snagline_max_body = max_body_bytes
     return _Handler
@@ -219,10 +236,11 @@ def make_server(
     port: int = 8787,
     auth_token: str | None = None,
     max_body_bytes: int = 1_000_000,
+    max_risks: int = 1000,
 ) -> ThreadingHTTPServer:
     """Construct a ready-to-``serve_forever()`` sidecar server."""
     return ThreadingHTTPServer(
-        (host, port), make_handler(monitor, auth_token, max_body_bytes)
+        (host, port), make_handler(monitor, auth_token, max_body_bytes, max_risks)
     )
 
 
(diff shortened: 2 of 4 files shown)


yuki-fuyutsuki:
**Code Review**

This PR correctly closes the GET auth bypass (`GET /metrics` and `GET /risks` were unauthenticated while `POST` was protected), wires the previously dead `auth_token`/`max_body_bytes`/`max_risks` plumbing through `snagline serve` with env fallback, and bounds the `POST /risks` growth vector with a `deque(maxlen)`. Auth is now checked before path dispatch so unauthenticated 404 enumeration is blocked and `GET /health` remains intentionally open for probes. Implementation is small, well-tested, and the new stderr warning when no token is set is a good operational fail-safe.

**medium**
`src/snagline/server/http_server.py:73`
Direct `==` comparison of the supplied token against the secret leaks timing information and enables a standard timing side-channel to brute-force the token byte-by-byte. `X-Snagline-Token` and `Authorization: Bearer` branches both use `==`. For a shared-secret sidecar this should be constant-time.

```suggestion
import hmac
import json
import logging
import sys
from collections import deque
```

```suggestion
        def _authorized(self) -> bool:
            token = self.snagline_auth
            if not token:
                return True
            authz = self.headers.get("Authorization", "")
            if authz.startswith("Bearer "):
                return hmac.compare_digest(authz[len("Bearer ") :].strip(), token)
            hdr = self.headers.get("X-Snagline-Token")
            return hmac.compare_digest(hdr, token) if hdr is not None else False
```

**low**
`src/snagline/server/http_server.py:227` and `src/snagline/cli.py:253`
`max_risks` is silently coerced via `max(1, max_risks)` and `max_body_bytes` has no validation at all. `--max-risks 0` or `--max-risks -10` becomes `1` with no error, and `--max-body-bytes -1` makes every `POST` return 413. Invalid operator input should fail fast rather than reconfigure the bound silently.

```suggestion
    if max_risks < 1 or max_body_bytes < 1:
        raise ValueError("max_risks and max_body_bytes must be >= 1")
    _Handler.snagline_risks = deque(maxlen=max_risks)
```

**low**
`src/snagline/server/http_server.py:95` / `134`
`snagline_risks` is a `deque(maxlen)` appended from `ThreadingHTTPServer` worker threads and copied via `list()` in `do_GET` without synchronization. CPython's GIL makes `append` atomic but iteration in `list(deque)` while a concurrent `append` evicts the leftmost item can yield a torn snapshot. Pre-existing race but now the eviction path makes it more likely under load; guard with a `threading.Lock`.


[github run](/Cyrax321/SNAGLINE/actions/runs/32632986313)

yuki-fuyutsuki:
Thank you for contributing to SNAGLINE @abyyxhek. This is an excellent fix for #68.

You correctly closed the auth gap on GET, wired the CLI so `snagline serve` can actually be secured from the command line and environment, and bounded the risk retention. The ordering with `GET /health` staying open for probes and keeping the 404 behind auth is exactly right. Tests are thorough and the insecure startup warning is a nice touch.

Note: Approving as is. Leaving three small follow up nits for a future PR, nothing blocking here:

1. src/snagline/server/http_server.py:67 timing side channel. `==` on the token check leaks timing. Follow up can switch to `hmac.compare_digest` for both Authorization Bearer and X-Snagline-Token branches.

2. src/snagline/server/http_server.py:227 and src/snagline/cli.py:253 input validation. `max(1, max_risks)` silently coerces 0 or negative to 1, and negative max_body_bytes would make every POST return 413. Follow up can fail fast with a clear error when values are less than 1.

3. src/snagline/server/http_server.py:95 thread safety. `list(deque)` in do_GET while ThreadingHTTPServer workers append can tear on eviction. Follow up can guard append and list with a threading.Lock. Pre existing but more likely now with bounded eviction.

Great attention to security and operational detail here. Thank you again for the contribution and the detailed reproduction. LGTM.

PR #94, opened 2026-08-24
fix: alert once per finding in LoopDetector and bound the DedupSink cooldown table

abyyxhek:
## Summary of Changes

Two bugs in the alert-deduplication path added for issue #4, both of which
defeat the dedupe they were written to provide.

**1. `LoopDetector` re-fired on every repetition of an interleaved loop.**
The `_fired` flag was cleared by any step whose *own* signature had not
repeated. A step that is not part of the loop says nothing about whether the
loop is still running, so it was read as an all-clear and the very next
repeat re-fired. The flag only worked for a strictly back-to-back run — not
the common shape, where an agent retries one failing tool between reasoning
turns.

Re-arming is now decided from the window's contents: a signature stays
`fired` while it still saturates the window, and re-arms once it drops below
the threshold or ages out. The flag is also keyed by signature rather than by
episode alone, so two actions looping in one episode are two findings and one
clearing does not re-arm the other. `ErrorCascadeDetector` already re-arms
this way — its comment says it "mirrors `LoopDetector`", which is now true.

**2. `DedupSink` retained every cooldown key forever, and measured elapsed
time on the wall clock.**

- The cooldown table was never pruned. With the default
  `(episode_id, trigger, severity)` key that is one permanent entry per
  episode for the life of the process — and "a long-lived monitor watching
  many short runs" is exactly this sink's deployment. An entry past its
  cooldown can never suppress anything again, so expired keys are now swept,
  at most once per cooldown window so the `O(len)` pass amortizes away. Keys
  still inside their window are kept — they are what makes suppression work.
  This mirrors `MemoryStateBackend.release()`, which already exists for the
  same reason.
- Elapsed time came from `time.time()`. Wall-clock time can step backwards
  (NTP correction, an operator setting the date), which makes `now - last`
  negative, holds it below the cooldown, and silences the key until real time
  catches up. Indefinite silence is the one failure mode alert suppression
  must not have, so the interval is now measured on `time.monotonic()`.

No public API changed: constructor signatures, the `key_fn` contract, and
fail-open behaviour are all as before.

## Justification

Anyone running SNAGLINE against a real long-horizon agent hits both of these.

A stuck agent is the headline use case, and it is precisely the case that
storms: replaying a 43-step trajectory where one tool call is retried 23x
between distinct plan steps produced **22 alerts on `master` and 2 on this
branch**. On-call paging on the 22-alert behaviour is the outcome issue #4
was filed to prevent.

The retention leak lands on the same deployment. 50,000 episodes through a
`DedupSink` with default settings retained 50,000 keys (~1.29 MB) of
bookkeeping that could no longer affect a single decision — in the one
component whose entire purpose is production alerting hygiene.

## Kind of Contribution

Bug Fix, Tests

Not a performance change, but the loop detector is on the hot path, so
`benchmarks/overhead_benchmark.py` was run to confirm the fix does not cost
anything (3 runs each, same machine):

| | median | |
|---|---|---|
| `master` | 2.96 / 2.96 / 2.97 us/step | |
| this branch | 2.94 / 2.95 / 2.98 us/step | |

Identical within noise. Only the signatures that have already escalated are
re-checked, and the window holds at most one per `repeat_threshold` slots; on
the common path nothing is looping, the dict has no entry for the episode,
and the step does exactly the work it did before.

(An earlier draft rebuilt a `Counter` over the window every step and measured
4.85-4.97 us/step — a 68% regression. It was rewritten rather than shipped.)

## Unit Tests:

`tests/detectors/test_loop.py` — three new cases, plus a `_fires()` helper
that returns which step indices alerted:

- `test_sustained_loop_alerts_once_even_when_interleaved` — the straight run
  that already worked, and the interleaved run that did not. Both must alert
  exactly once. On `master` the interleaved case fires 21 times.
- `test_loop_realerts_only_after_it_clears` — a loop that stops, ages out of
  the window, and returns is a second finding and must escalate again. This
  guards the fix against over-suppressing.
- `test_two_distinct_loops_each_alert` — two signatures looping in one
  episode are two findings; neither suppresses nor re-arms the other.

`tests/test_sinks_dedup.py` — three new cases:

- `test_dedup_table_does_not_grow_without_bound` — 2000 keys are all retained
  while live (they are load-bearing), and dropped once expired.
- `test_dedup_sweep_keeps_suppressing_live_keys` — sweeping must not break
  suppression for any key still inside its cooldown.
- `test_dedup_cooldown_survives_a_backward_wall_clock_step` — emits, then
  steps `time.time()` a day backwards mid-cooldown, and asserts the alert
  still re-emits once the cooldown has elapsed in real time. It also asserts
  `time.monotonic` was never patched, so the test cannot pass by accident.

All six fail on unmodified `master`. Full suite: **229 passed, 1 skipped**,
coverage 83%; `loop.py` and `dedup.py` are both at 100%. The one failure,
`test_hook_bridge.py::test_watch_file_follow_sees_appended_lines`, is
pre-existing on `master` and is issue #69 (Windows `SIGINT`) — unrelated to
this change.

`ruff check`, `ruff format --check`, and `mypy` are clean on `src` and
`tests`.

## Execution Tests:

Generated a realistic stuck-agent trajectory — 43 steps, one `search` tool
call retried 23 times interleaved between distinct planning steps, ending in
an error cascade — and ran it through the real CLI on both branches:

```
snagline replay stuck_agent.jsonl
```

| | risks emitted |
|---|---|
| `master` | 22 |
| this branch | 2 |

The two remaining risks are the one real loop and the one real error
cascade — the same two findings, reported once each instead of once per step.
`ErrorCascadeDetector` reported that trajectory's cascade once on both
branches, which is what identified the loop detector as the outlier.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced repeated loop alerts when recurring actions are interleaved with other steps.
  * Loop alerts can trigger again after the loop has cleared.
  * Distinct looping actions are tracked and alerted independently.
  * Improved alert deduplication reliability when system time changes.
  * Prevented expired deduplication entries from accumulating indefinitely.

* **Tests**
  * Added coverage for loop-alert behavior, cooldown cleanup, active suppression, and clock changes.

### DIFF
--- a/src/snagline/detectors/loop.py
+++ b/src/snagline/detectors/loop.py
@@ -75,6 +75,18 @@ class LoopDetector:
                 seen = count if sig == event.action_signature else w.count(sig)
                 if seen < self.repeat_threshold:
                     fired.discard(sig)
+            if not fired:
+                # Every escalated signature re-armed, so this episode is back to
+                # "nothing looping". Drop the entry instead of leaving an empty
+                # set behind: otherwise every episode that ever looped keeps
+                # bookkeeping until ``reset()``, an unbounded per-episode leak of
+                # exactly the kind the DedupSink sweep removes. Clearing the
+                # local too makes the escalation path below recreate the entry
+                # through ``setdefault`` -- mutating the detached set would
+                # record the fire against a dict entry that no longer exists and
+                # silently disable dedupe for the rest of the episode.
+                del self._fired[event.episode_id]
+                fired = None
         if count < self.repeat_threshold:
             return None
         if fired is None:
--- a/src/snagline/detectors/loop.py
+++ b/src/snagline/detectors/loop.py
@@ -4,6 +4,9 @@ Per-episode sliding window (``collections.deque``) of recent
 ``action_signature`` values. If the same signature appears
 ``repeat_threshold`` times within ``window_size`` steps, emit a risk.
 
+Each looping signature escalates once and then stays quiet until that loop
+actually clears, so a long repetition alerts once rather than on every step.
+
 No raw content is read -- only the one-way ``action_signature`` hash, so a
 loop of identical retry attempts is caught without ever seeing the prompt or
 response text (project.md §1.4).
@@ -40,20 +43,45 @@ class LoopDetector:
         # Dedupe: emit once per repetition episode (issue #4). Without this the
         # detector re-fires on every step while the same action keeps repeating,
         # which is alert spam.
-        self._fired: dict[str, bool] = {}
+        #
+        # Keyed by signature, not by episode alone: two different actions looping
+        # in one episode are two distinct findings, and one of them going quiet
+        # must not re-arm the other.
+        self._fired: dict[str, set[str]] = {}
 
     def observe(self, event: StepEvent) -> FailureRisk | None:
         w = self._windows.setdefault(event.episode_id, deque(maxlen=self.window_size))
         w.append(event.action_signature)
         count = w.count(event.action_signature)
+        fired = self._fired.get(event.episode_id)
+        if fired:
+            # Re-arm any signature whose loop has ended -- it fell below the
+            # threshold, or aged out of the window entirely -- so a later
+            # repetition of it escalates again.
+            #
+            # Re-arming is decided here, from the window, and *not* from the
+            # current step's own count. A step that is not part of the loop says
+            # nothing about whether the loop is still running, so reading it as an
+            # all-clear re-armed the flag and re-fired on the very next repeat:
+            # one loop interleaved with distinct steps -- a retry between
+            # reasoning turns -- then alerted on every repetition, which is the
+            # spam this flag exists to prevent.
+            #
+            # Only the escalated signatures are re-checked, and the window holds
+            # at most one per ``repeat_threshold`` slots. On the common path
+            # nothing is looping, this dict has no entry, and the step costs
+            # exactly what it did before.
+            for sig in tuple(fired):
+                seen = count if sig == event.action_signature else w.count(sig)
+                if seen < self.repeat_threshold:
+                    fired.discard(sig)
         if count < self.repeat_threshold:
-            # The repeated signature has dropped out of the window: allow a
-            # future repetition to re-escalate.
-            self._fired[event.episode_id] = False
             return None
-        if self._fired.get(event.episode_id, False):
+        if fired is None:
+            fired = self._fired.setdefault(event.episode_id, set())
+        elif event.action_signature in fired:
             return None
-        self._fired[event.episode_id] = True
+        fired.add(event.action_signature)
         score = min(1.0, count / self.repeat_threshold * 0.5)
         return FailureRisk(
             event.episode_id,
--- a/src/snagline/sinks/dedup.py
+++ b/src/snagline/sinks/dedup.py
@@ -6,6 +6,17 @@ second across thousands of steps should page once, not thousands of times.
 within a cooldown window. The key defaults to ``(episode_id, trigger,
 severity)`` but can be customized via ``key_fn`` (e.g. to dedupe purely by
 detector type, or by the full detail string).
+
+The cooldown table is bounded: once a key's cooldown has elapsed it can never
+suppress anything again, so expired keys are pruned instead of being retained
+for the life of the process. With the default key that would otherwise be one
+permanent entry per episode id, and a long-lived monitor watching many short
+runs is exactly this sink's intended deployment.
+
+Elapsed time is measured on the monotonic clock. Wall-clock time can step
+(NTP correction, an operator setting the date), and a backward step would
+stretch a cooldown into an indefinite silence -- the one failure mode an
+alert-suppression wrapper must not have.
 """
 
 from __future__ import annotations
@@ -41,16 +52,39 @@ class DedupSink:
         self._key_fn = key_fn or _default_key
         self._last: dict[Any, float] = {}
         self._lock = threading.Lock()
+        self._swept = time.monotonic()
 
     def emit(self, risk: FailureRisk) -> None:
         key = self._key_fn(risk)
-        now = time.time()
+        now = time.monotonic()
         with self._lock:
             last = self._last.get(key)
             if last is not None and (now - last) < self._cooldown:
                 return
             self._last[key] = now
+            # At most one sweep per cooldown window. Anything that expired since
+            # the previous sweep is expired now, so this is frequent enough to
+            # keep the table from growing without bound, and rare enough that the
+            # O(len) pass is amortized away across the window's emits.
+            if (now - self._swept) >= self._cooldown:
+                self._sweep(now)
         with contextlib.suppress(Exception):
             # Never let a wrapped-sink failure corrupt our cooldown bookkeeping
             # (fail-open), and do not re-raise into the monitor.
             self._sink.emit(risk)
+
+    def _sweep(self, now: float) -> None:
+        """Drop keys whose cooldown has elapsed. Caller must hold ``_lock``.
+
+        An entry older than the cooldown is dead weight: the suppression test in
+        ``emit`` is already false for it, so dropping it changes no decision this
+        sink would make. Keys still inside their window are kept -- they are what
+        makes suppression work, so retaining them is correctness, not a leak.
+
+        The table therefore holds the keys seen in roughly the last two cooldown
+        windows (a key can expire just after one sweep and wait until the next),
+        rather than every key seen for the life of the process.
+        """
+        cutoff = now - self._cooldown
+        self._last = {k: t for k, t in self._last.items() if t > cutoff}
+        self._swept = now
--- a/src/snagline/sinks/dedup.py
+++ b/src/snagline/sinks/dedup.py
@@ -17,6 +17,11 @@ Elapsed time is measured on the monotonic clock. Wall-clock time can step
 (NTP correction, an operator setting the date), and a backward step would
 stretch a cooldown into an indefinite silence -- the one failure mode an
 alert-suppression wrapper must not have.
+
+A non-positive ``cooldown_seconds`` disables suppression: ``emit`` becomes a
+pass-through that never touches the table. Wrapping is normally avoided
+upstream in that case (``cli._maybe_dedup``), but the sink is public API and
+must behave sanely when constructed directly.
 """
 
 from __future__ import annotations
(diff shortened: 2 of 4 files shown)


anya-research:
**Code Review**

This PR correctly fixes two defects in the alert-deduplication path from #4: `LoopDetector` now re-arms per-signature from window contents instead of the current step's count, fixing 22→2 alert-storm on an interleaved 43-step trajectory, and `DedupSink` bounds its cooldown table with amortized sweeping and switches elapsed-time to `time.monotonic()` to avoid NTP-backstep indefinite silence. The changes are narrowly scoped, preserve public API and fail-open contracts, and are exercised by six new regression tests and overhead benchmarks showing no hot-path cost (2.96→2.95 µs/step). Overall quality is high — well-documented, well-tested, and performance-aware.

**low**
`src/snagline/detectors/loop.py:74`
The `fired` set is never pruned when empty. After `for sig in tuple(fired): ... fired.discard(sig)` clears the last signature, `self._fired[episode_id]` remains as `set()` (falsy but present). The next step skips the `if fired:` re-arm check and re-uses the empty set, so correctness is preserved, but every episode that ever looped retains an empty entry in `_fired` (and its `deque` in `_windows`) until `reset()` — a small unbounded per-episode leak similar to the `DedupSink` table this PR fixes.

```suggestion
            for sig in tuple(fired):
                seen = count if sig == event.action_signature else w.count(sig)
                if seen < self.repeat_threshold:
                    fired.discard(sig)
            if not fired:
                self._fired.pop(event.episode_id, None)
```

**low**
`src/snagline/sinks/dedup.py:57`
`key = self._key_fn(risk)` is executed outside the `contextlib.suppress` and outside any try/except in `emit`. A raising `key_fn` escapes `DedupSink` and relies on `Monitor._dispatch`'s outer fail-open guard. That matches `master` behavior and is safe in the normal `Monitor` path, but the docstring claim “Fail-open: a bookkeeping error never blocks the wrapped sink” is not enforced inside the sink itself; a standalone `DedupSink` user without `Monitor` will see the exception propagate and lose the alert.

**low**
`src/snagline/sinks/dedup.py:69`
With `cooldown_seconds <= 0` the guard `(now - self._swept) >= self._cooldown` is always true, so every `emit` performs an `O(n)` `_sweep` under `self._lock`. Functionally correct (cooldown 0 means “never suppress”), but it turns a previously `O(1)` path into `O(n)` per emit and holds the lock for the full comprehension, which can stall concurrent `emit` calls if the table is large. Consider early-return when `self._cooldown <= 0` or skipping the sweep in that case.



[github run](/Cyrax321/SNAGLINE/actions/runs/32718160648)

yuki-fuyutsuki (dismissed):
Thank you for this contribution, and especially for the caliber of engineering discipline
that came with it. Reviews written this carefully make open source maintainership
genuinely enjoyable.

This is a high-quality bug fix PR. Both defects are real, well diagnosed, and correctly
scoped to the alert deduplication path introduced for issue #4.

**LoopDetector.** Moving the rearm decision from the current step's signature count to
the contents of the window is the right fix. A non-looping step carries no information
about whether a loop has ended, and treating it as an all-clear was the root cause.
Keying the fired flag by signature rather than episode also resolves cross-suppression
between distinct loops, restoring the symmetry that ErrorCascadeDetector already
documents. Verified against the reported trajectory shape; the logic holds.

**DedupSink.** Switching elapsed measurement to `time.monotonic()` is unambiguously
correct. Indefinite silence from an NTP backstep is the one failure mode alert
suppression must never have. Bounding the table with one amortized sweep per cooldown
window is a reasonable design at this component's scale.

**Tests and benchmarks.** Six regression tests that all fail on unmodified master is
exactly the standard we hold. Full coverage on both touched modules, and the assertion
that `time.monotonic` was never patched, show real care. Reporting that an earlier draft
regressed the hot path by 68% and rewriting it rather than shipping deserves explicit
recognition. Benchmark numbers confirm parity within noise.

Public API, constructor signatures, `key_fn` contract, and fail-open behavior are
unchanged. CI is green, no conflicts.

## Requested changes before merge

1. **`src/snagline/detectors/loop.py` (~line 74):** When the fired set empties, the
   episode entry stays behind in `_fired`, so every episode that ever looped retains
   bookkeeping until `reset()`. That's a small, unbounded per-episode leak of precisely
   the class this PR eliminates elsewhere. Please prune the episode entry once its set
   is empty, mirroring the expired-key sweep in `DedupSink`.

2. **`src/snagline/sinks/dedup.py` (~line 69):** With `cooldown_seconds <= 0` the guard
   is always true, so every emit performs an O(n) sweep while holding the lock, turning
   a previously O(1) path into a potential stall for concurrent emitters. Please add an
   early return or sweep-skip for the non-positive cooldown case, and consider whether
   an expiry heap or bounded per-call pruning is warranted for high-cardinality
   deployments, as flagged by CodeRabbit.

3. **`tests/test_sinks_dedup.py`:** Two new tests depend on real elapsed time. The bound
   test assumes 2000 emits complete inside the 0.05s window, and the live-suppression
   test does not reliably trigger a sweep at all, so it can't distinguish correct
   sweeping from sweeping that deletes live keys. Please introduce a controlled
   monotonic fake clock and rewrite both against it so they're deterministic on slow
   machines.

4. **`src/snagline/sinks/dedup.py` (docstring):** A raising `key_fn` propagates out of
   `emit` when the sink is used standalone outside `Monitor`, contradicting the
   fail-open claim. Either bring `key_fn` inside the suppressed scope or soften the
   docstring, whichever you prefer, but the documented contract must match observed
   behavior.

None of these block the direction of the work. Points 1 and 3 are the reason this review
is marked Request Changes rather than Approve; points 2 and 4 are small enough to ride
along in the same push.

Thank you again for the precise diagnosis, the benchmark honesty, and tests that prove
failure on master. This is close. Push the four items above and I will happily approve.

abyyxhek:
Pushed as `ecf3667`. All four items, plus a note on the expiry-heap suggestion.

**1. `_fired` empty-set retention.** Fixed. Once every escalated signature re-arms, the episode is back to "nothing looping" and the dict entry is deleted rather than left holding an empty set — otherwise every episode that ever looped keeps bookkeeping until `reset()`, which is the same unbounded per-episode growth this branch removes from the dedup path.

The subtle part is that the local has to be cleared too (`fired = None`), so the escalation path below recreates the entry through `setdefault`. This matters because a prune and an escalation can land on the *same* `observe()` call — the window slides, the old signature drops below the threshold, a new one reaches it. Mutating the detached set there records the fire against a dict entry that no longer exists, so the next repeat escalates again: exactly the spam the bookkeeping exists to prevent. `test_pruning_the_fired_set_does_not_break_realerting` pins it with window 5 / threshold 3 and `[A,A,A,B,B,B,B]`, where `[A,A,A,B,B] → [A,A,B,B,B]` drops A from 3 to 2 and lifts B from 2 to 3 in one step.

**2. `key_fn` outside the fail-open handler.** Fixed — it now runs inside the `try`. Two failure modes, both covered: a `key_fn` that raises, and one returning an unhashable key (which raises in the table lookup, not in `key_fn` itself). Either way the alert is delivered rather than dropped, and nothing is re-raised. Worth stating explicitly since the class docstring now promises it: this holds standalone, not just under `Monitor` — an alerting wrapper must never turn a bad key into a lost alert or an exception in the host's ingest path.

**3. Sweep tests on real elapsed time.** Both now drive an injected monotonic clock. You were right about the failure mode and there was a second one next to it: `test_dedup_table_does_not_grow_without_bound` assumed 2,000 emits finish inside a 0.05s window, so a slow machine swept early and failed a *correct* implementation, and `test_dedup_sweep_keeps_suppressing_live_keys` never crossed the sweep threshold at all — it passed whether or not the sweep dropped keys that were still suppressing. The live-key test now deliberately triggers the sweep (emit at t+0, t+29, t+31 with a 30s cooldown) and asserts the surviving key still suppresses.

**4. Non-positive cooldown.** Fixed with a precomputed `_enabled` flag. `emit` is now a genuine pass-through: with no window nothing can be a repeat, so the table is never touched and the sweep guard — unconditionally true for a non-positive cooldown — can't turn an O(1) path into an O(len) rebuild under the lock on every emit.

Measured, interleaved single-process A/B (`importlib`, 15 alternating iterations, min):

| path | delta |
|---|---|
| `DedupSink` cooldown=0 | **−61.7%** (2.6× faster) |
| `DedupSink` cooldown=300 | +3.2% (≈18 ns — one attribute read + branch) |
| `LoopDetector` healthy | +0.3% |
| `LoopDetector` looping + prune | +2.8% |

Interleaving mattered: naive before/after runs drifted 0.523 → 0.438 → 0.402 *within a single version* from warmup, which is larger than every delta in that table.

**On the expiry-heap / bounded-pruning suggestion** — I'd rather keep the amortized sweep. The heap trades one O(n) pass per cooldown window for a per-emit push plus lazy-deletion tombstones, and because a key's timestamp is refreshed on re-emit the stale entries have to be validated against `_last` on pop, so it's strictly more state and more code on the hot path for the same asymptotic behaviour. The sweep's actual cost is one dict comprehension per window amortized across that window's emits, and the table's bound is "keys seen in roughly the last two cooldown windows" — the +3.2% above is that whole mechanism. If a deployment ever shows the sweep as a real cost I'd revisit, but the measurement doesn't support paying that complexity now.

**Verification.** Each of the six new tests was mutation-checked — revert the fix it guards and the test must fail. That caught two of my own tests that passed without guarding anything: the orphaned-set case above (my first sequence separated the prune from the escalation, so the bug never triggered), and the pass-through, which is invisible in end state — without the guard an always-true sweep empties the table every emit, leaving `_last == {}` and all alerts delivered, identical to never recording anything. It's now asserted on sweep count via a counting subclass.

100% statement and branch coverage on both touched modules. `ruff check`, `ruff format --check`, and `mypy` are clean on the diff (the 33 mypy errors and the `test_watch_file_follow_sees_appended_lines` failure are both pre-existing on master — the latter is the Windows `SIGINT` issue, #69).

Cyrax321 (approved):
Thank you for this revision round and for engaging with each review item on its merits.

All four requested changes are resolved and verified. The LoopDetector bookkeeping is now pruned once its fired set clears, and the interaction hazard you identified within the fix itself deserves particular credit: when the window slides an old signature below threshold while lifting a new one onto it in a single observe() call, pruning without clearing the local reference would have recorded escalation against a detached set and reintroduced repeat firing. Clearing the local and letting setdefault recreate the entry is correct, and test_pruning_the_fired_set_does_not_break_realerting pins that exact interleaving with window 5, threshold 3, and the [A,A,A,B,B,B,B] sequence. That was thorough work beyond what the review asked for.

The fail open contract in DedupSink now holds standalone: key_fn executes inside the suppressed scope, covering both failure modes including the unhashable return that raises at lookup rather than inside key_fn itself, so the docstring claim is finally enforced rather than aspirational. Both sweep tests are rewritten against an injected monotonic clock, with the live suppression test deliberately crossing the sweep threshold at t+0, t+29, t+31 under a 30s cooldown and asserting the surviving key still suppresses. The non positive cooldown case is handled through a precomputed enabled flag making emit a genuine pass through, and the measured 2.6x improvement on that path is welcome.

On the expiry heap suggestion from CodeRabbit, I concur with keeping the amortized sweep. Refreshed timestamps make heap entries stale by construction, requiring per emit pushes plus lazy deletion validated against the table on pop, which is strictly more state and hot path code for an equivalent bound given the table already holds roughly two cooldown windows of keys. Revisiting only if production measurement demonstrates real cost is the right disposition.

Finally, the verification discipline merits explicit acknowledgment. Mutation checking each new test, using it to surface two tests that passed while guarding nothing, pinning the pass through regression via sweep count on a counting subclass, and interleaving A/B runs to control warmup drift all exceed our internal standard. Coverage sits at 100% statement and branch on both modules, lint and types are clean on the diff, CI is green, and there are no conflicts with master.

This is approved. Thank you again for the precise diagnosis, honest benchmarking, and evidence based pushback. Work of this quality is always welcome here, and we look forward to your next contribution.

PR #95, opened 2026-08-24
fix: stop two silent sentinels from defeating documented guarantees

abyyxhek:
Summary of Changes

Fixes two independent defects that share one root cause: a value that doubles
as both real data and a "sentinel"/state silently overrides a contract the
module documents.

1. **`FailureRisk.severity` sentinel collision** (`src/snagline/risk.py`).
   The field defaulted to `SEVERITY_WARNING` and used that same `"warning"`
   string as the "unset" marker (`if self.severity == SEVERITY_WARNING:`). So
   an explicit `severity="warning"` was indistinguishable from an omitted one
   and got overwritten by the score-derived severity — escalated to
   `"critical"` for a high score, downgraded to `"info"` for a low one —
   violating the documented "a caller that passes `severity=` explicitly keeps
   that value" contract. Because `severity` is part of `DedupSink`'s default
   key, the silent rewrite also re-bucketed the alert and broke cooldown
   dedup. Fixed with an empty-string sentinel (`severity: str = ""`), which can
   never be a legitimate severity, so an explicit `"warning"` now survives.

2. **`BatchingSink` rate limit reset across batches** (`src/snagline/sinks/batching.py`).
   `_deliver` held its pacing clock in a local `last = 0.0` that reset on every
   call, so only pacing *within* a single batch worked. A burst arriving as
   many small interval-driven flushes — the normal production shape — let the
   first item of every batch through immediately, bypassing `max_per_second`
   entirely. Fixed by persisting the last-delivery timestamp on the instance
   (`self._last_delivery`) so pacing carries across batches.

Justification

Both bugs silently defeat a guarantee the library advertises. (1) means
severity-based routing and dedup misfire whenever a caller pins
`severity="warning"` — the alert lands in the wrong channel or is wrongly
suppressed. (2) means the rate limiter that exists specifically to keep a burst
from tripping a webhook/Slack/PagerDuty rate limiter does not actually pace a
real burst, because a real burst arrives as many small flushes rather than one
tidy batch. Both are the kind of failure that only shows up in production under
load and is hard to trace back.

Kind of Contribution

Bug Fix

Unit Tests:

`tests/test_sinks_dedup.py`:
- `test_risk_explicit_warning_severity_is_not_overwritten` — an explicit
  `severity="warning"` is kept for high/agreeing/low scores.
- `test_risk_explicit_severity_survives_the_dedup_key` — two risks pinned to
  the same severity but with scores in different derivation bands still dedupe
  together.

`tests/test_batching_sink.py`:
- `test_rate_limit_paces_across_separate_batches` — three risks each flushed on
  their own batch still take ≥0.18s at 10/s (was ~0s before the fix).
- `test_rate_limit_unset_does_not_delay_delivery` — over-correction guard: with
  no `max_per_second`, 20 individual flushes add no delay.

All four regression tests fail on the unmodified source and pass after the fix.

Execution Tests:

- Full suite: `227 passed, 1 skipped`, with one pre-existing unrelated failure
  (`test_hook_bridge.py::test_watch_file_follow_sees_appended_lines`, a Windows
  `Unsupported signal: 2` teardown issue that also fails on pristine `master`).
- `ruff check` and `ruff format --check`: clean across `src` and `tests`.
- `mypy`: no new errors in the four changed files (the pre-existing errors are
  all in unrelated test modules and also present on `master`).

### DIFF
--- a/src/snagline/risk.py
+++ b/src/snagline/risk.py
@@ -43,11 +43,18 @@ class FailureRisk:
     trigger: TriggerType
     detail: str  # short human-readable explanation, no raw content
     timestamp: float
-    severity: str = SEVERITY_WARNING  # auto-derived from score if left default
+    # Default "" means "unset" -- derived from the score in __post_init__. It is
+    # NOT ``SEVERITY_WARNING``: "warning" is a real value a caller may pass, and
+    # using it as the sentinel made an explicit ``severity="warning"`` indistin-
+    # guishable from unset, so it was silently overwritten by the score-derived
+    # severity. An empty string can never be a legitimate severity, so it is a
+    # safe sentinel that keeps the field type ``str`` for every call site.
+    severity: str = ""
 
     def __post_init__(self) -> None:
         # If the caller did not set an explicit severity, derive one from the
         # score so every risk carries a useful routing hint. A caller that
-        # passes `severity=` explicitly keeps that value.
-        if self.severity == SEVERITY_WARNING:
+        # passes `severity=` explicitly keeps that value exactly -- including
+        # ``severity="warning"``, which the old sentinel clobbered.
+        if not self.severity:
             object.__setattr__(self, "severity", severity_from_score(self.score))
--- a/src/snagline/sinks/batching.py
+++ b/src/snagline/sinks/batching.py
@@ -43,6 +43,14 @@ class BatchingSink:
         self._flush_interval = flush_interval
         self._min_gap = 1.0 / max_per_second if max_per_second else 0.0
         self._queue: collections.deque[FailureRisk] = collections.deque()
+        # Monotonic timestamp of the last delivery, carried *across* batches.
+        # Holding this in a ``_deliver`` local reset the rate limit on every
+        # flush, so the first item of each batch always went out immediately --
+        # which meant a burst arriving as many small flushes (the normal
+        # interval-driven shape) bypassed ``max_per_second`` entirely. 0.0 is
+        # "nothing delivered yet": ``monotonic() - 0.0`` is large, so the very
+        # first delivery never waits.
+        self._last_delivery = 0.0
         self._lock = threading.Lock()
         self._stop = threading.Event()
         # Set when the queue reaches ``max_batch`` (or on close) so the flusher
@@ -85,14 +93,13 @@ class BatchingSink:
         self._deliver(batch)
 
     def _deliver(self, batch: list[FailureRisk]) -> None:
-        last = 0.0
         for risk in batch:
             if self._min_gap:
                 now = time.monotonic()
-                wait = self._min_gap - (now - last)
+                wait = self._min_gap - (now - self._last_delivery)
                 if wait > 0:
                     time.sleep(wait)
-                last = time.monotonic()
+                self._last_delivery = time.monotonic()
             with contextlib.suppress(Exception):
                 # Fail-open: a delivery failure must not poison the queue.
                 self._sink.emit(risk)
--- a/tests/test_batching_sink.py
+++ b/tests/test_batching_sink.py
@@ -74,6 +74,52 @@ def test_rate_limit_paces_delivery():
         sink.close()
 
 
+def test_rate_limit_paces_across_separate_batches():
+    # The pacing clock was a local in _deliver, reset to 0.0 on every flush, so
+    # the first item of each batch always went out immediately and only pacing
+    # *within* one batch worked. The test above hides that by putting all three
+    # risks in a single batch (max_batch=1000 + one flush_now).
+    #
+    # A burst does not arrive as one tidy batch in production -- the flusher
+    # ticks on flush_interval and delivers whatever accumulated, so N risks
+    # normally arrive as N small batches. Here each risk is flushed on its own,
+    # which delivered all 3 with zero delay before the fix.
+    inner = _RecordingSink()
+    sink = BatchingSink(
+        inner, max_batch=1000, flush_interval=3600.0, max_per_second=10.0
+    )
+    try:
+        start = time.monotonic()
+        for i in range(3):
+            sink.emit(_risk(i))
+            sink.flush_now()
+        elapsed = time.monotonic() - start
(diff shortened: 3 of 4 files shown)


yuki-fuyutsuki:
**Code Review**

This PR fixes two genuine silent-contract violations — the `SEVERITY_WARNING` sentinel that made `severity="warning"` indistinguishable from unset (breaking caller-pinned routing and `DedupSink`'s default key) and the local `last = 0.0` in `BatchingSink._deliver` that reset on every flush and let interval-driven bursts bypass `max_per_second`. Both fixes are directionally correct, minimal, and backed by well-targeted regression tests that fail pre-fix and pass post-fix. Remaining issues are concurrency and sentinel-tightness, not logic regressions.

**high**
`src/snagline/sinks/batching.py:53` and `src/snagline/sinks/batching.py:95-103`
`self._last_delivery` is read/written unsynchronized while `_deliver` can run concurrently on the background thread (`_run`/`_flush`) and the caller's thread (`flush_now()`/`close()`). The `if wait>0: time.sleep(wait)` gap between the read and the write allows two flushes to compute the same `wait` and both emit immediately, re-introducing a burst. `close()` also `join(timeout=flush_interval+1)` then does a no-op ` _flush()` on an already-drained queue while the background `_deliver` is still `sleep`-ing through the rate limit, so `close()` can return before the in-flight batch finishes — violating the documented drain-on-close guarantee.
```suggestion
    def _deliver(self, batch: list[FailureRisk]) -> None:
        for risk in batch:
            if self._min_gap:
                with self._lock:
                    now = time.monotonic()
                    wait = self._min_gap - (now - self._last_delivery)
                if wait > 0:
                    time.sleep(wait)
                with self._lock:
                    self._last_delivery = time.monotonic()
            with contextlib.suppress(Exception):
                # Fail-open: a delivery failure must not poison the queue.
                self._sink.emit(risk)
```

**medium**
`src/snagline/risk.py:52` and `src/snagline/risk.py:59`
The empty-string sentinel is an improvement over colliding with `"warning"`, but `if not self.severity:` is overly broad: any falsy string (`""`, and at runtime `None`/`0` via untyped callers or deserialization) silently derives instead of failing loudly, hiding caller bugs. Pinning to the exact sentinel preserves the `str` type the PR wants while making the contract explicit and `mypy`-friendly. A `None`/`object()` sentinel would be strictly safer but is a larger API change.
```suggestion
        if self.severity == "":
            object.__setattr__(self, "severity", severity_from_score(self.score))
```

**low**
`src/snagline/sinks/batching.py:44`
`self._min_gap = 1.0 / max_per_second if max_per_second else 0.0` treats `0` as disabled (good, avoids `ZeroDivisionError`) but accepts negative `max_per_second`, producing a negative `_min_gap` that still enters the `if self._min_gap:` branch and does useless `monotonic()` work every delivery. Validate `max_per_second` and clamp/clip to `0.0` for `<=0` to fail fast and keep the hot path clean.
```suggestion
        self._min_gap = 1.0 / max_per_second if max_per_second and max_per_second > 0 else 0.0
```


[github run](/Cyrax321/SNAGLINE/actions/runs/32756060963)

Cyrax321:
Thank you for this contribution and for continuing to hold the library to its own documented contracts. The framing of both defects as silent guarantee violations is exactly how we want regressions like these reasoned about.

## Review

This PR fixes two genuine contract violations, and both diagnoses are correct and well motivated.

**FailureRisk severity sentinel.** Using SEVERITY_WARNING as both a real value and the unset marker made an explicitly pinned `"warning"` indistinguishable from omission, which is a direct violation of the documented contract and, worse, quietly re-bucketed DedupSink's default key. The empty string is a sound choice of sentinel because it can never be a legitimate severity, so the fix preserves the declared `str` type without an API break. The regression test proving the dedup key survives score band changes covers the most dangerous downstream effect. Good.

**BatchingSink pacing.** Holding the pacing clock in a function local meant pacing existed only within a single batch, while real production bursts arrive as many small interval driven flushes. Persisting the timestamp on the instance is the correct direction, and the paired tests are well chosen: one proves cross batch pacing now works, and the over correction guard proving unset max_per_second adds zero delay shows you thought about the fix breaking the fast path. That second test is the kind reviewers usually have to ask for. Well done.

Both new test files fail on unmodified master and pass here, suite is green apart from the known pre-existing Windows failure, lint and types are clean, public API unchanged.

That said, I agree with the collaborator review, and one issue rises to blocking.

## Requested changes before merge

1. **src/snagline/sinks/batching.py (~line 53):** `self._last_delivery` is read and written unsynchronized while `_deliver` can run concurrently on the background thread and via `flush_now()` or `close()` on the caller's thread. The gap between computing `wait` and recording the new timestamp lets two flushes compute identical waits and both emit immediately, reintroducing precisely the burst this fix exists to prevent. Please add synchronization so the wait computation and timestamp update happen atomically with respect to other deliveries. Relatedly, `close()` currently returns while the background `_deliver` may still be sleeping through a rate limit window, violating the documented drain on close behavior. Please add a mechanism ensuring close waits for any in flight paced delivery to finish.

2. **src/snagline/risk.py (~line 52):** Pin the check to the exact sentinel, `if self.severity == "":`, rather than `if not self.severity:`. Truthiness silently derives for any falsy input including a runtime `None` from untyped callers or deserialization, hiding caller bugs instead of surfacing them. Exact comparison keeps the mypy friendliness of the current approach and makes the contract explicit. Please add that tightening.

3. **src/snagline/sinks/batching.py (~line 44):** A negative `max_per_second` currently produces a negative `_min_gap` that still enters the pacing branch and does wasted work on every delivery. Please validate at construction: either clamp non positive values to disabled or raise a clear error consistent with how the library treats invalid constructor arguments elsewhere. I would lean toward raising, but either is acceptable if applied consistently.

None of these challenge the core fixes, which are right. Point 1 is why this review carries Request Changes; points 2 and 3 are small and can land in the same push.

Thank you again for sharp diagnosis, honest execution reporting, and regression tests that prove failure on master. Address the three items above and I will approve promptly.

Issue #106, opened 2026-08-25
📋 Execution board: complete P2/P3 + long-horizon roadmap (multi-agent)


Cyrax321:
Single reference point for finishing every planned goal. Pick any task from an **open pool** — pools are parallel-safe by design; respect the **dependency chain**.

## Dependency chain (sequential)
```
#82 eval harness ──► #101 calibration ──┐
                 ├──► #102 retrain ─────┤
                 └──► #80 ESN ensemble ─┴──► #81 drift extra ──► #105 changelog/tags ──► #104 PyPI
```

## Pool A — ship in flight 🔴
| Issue | What |
|---|---|
| PR (feat/horizon-detectors) | #84 #85 #86 merged-in-branch + #91 snapshot/restore → master |

## Pool B — docs & CI (all parallel-safe) 🟢
| Issue | What |
|---|---|
| #96 | Truth sweep: test counts, badge, stale limitation text |
| #97 | Spec parity in project.md (horizon set + snapshots) |

## Pool C — observability (parallel-safe) 🟢
| Issue | What |
|---|---|
| #98 | Prometheus exposition |
| #99 | JSON-lines structured logging |
| #100 | langchain-core in CI |

## Pool D — independent fixes/docs (parallel-safe) 🟢
| Issue | What |
|---|---|
| #69 | Windows SIGINT orphan in test_hook_bridge |
| #103 | Sidecar TLS story |

## Chain E — research (sequential, see diagram) 🟡
| Order | Issue |
|---|---|
| 1st | #82 detection_accuracy.py honesty gate |
| 2nd | #101 auto-calibration · #102 retrain cadence · #80 ESN (any order after #82) |
| 3rd | #81 semantic drift |

## Pool F — long-horizon backlog (already specced, parallel-safe) 🟢
| Issue | What |
|---|---|
| #87 | StagnationDetector |
| #88 | SideEffectGuardDetector |
| #89 | Loop hardening modes |
| #90 | CompactionTripwire |
| #92 | Horizon-scale config + heartbeat sink |
| #93 | EnforcementPolicy hook |
| #79 | CONTINUUM adapter/sink (verify Storage API first) |

## Final gate 🏁
#104 PyPI publish — only after Pool A merges + #105 tags exist. #79 explicitly NOT blocking 0.1.0.

## Rules for agents
1. One issue per agent/branch; branch name = issue number.
2. Reference the issue in every commit (`Refs #N` / `Fixes #N`).
[...]

PR #111, opened 2026-08-25
feat(sink): JSON-lines structured logging via stdlib logging (Fixes #99)

### LINKED ISSUE #99: [sink] Structured JSON-lines logging option (stdlib only)
## Context
P3 item 10 tail. Console sink prints human text; log aggregators want JSON lines.

## Tasks
- [ ] Optional JSON-lines mode for risk emission: one JSON object per FailureRisk (ts, episode_id, trigger, severity, score, detail) — structure only, never content (§1.4)
- [ ] Wire via stdlib `logging` + a minimal JSONFormatter; env/config selectable (`SNAGLINE_LOG_FORMAT=json`)
- [ ] Document 12-factor key in config table

## Acceptance
Every line parses with `json.loads`; no new deps; fail-open preserved if formatter raises.

Parallel-safe.

Cyrax321:
## Summary

Adds a zero-dependency LoggingSink that emits one machine-readable JSON object per FailureRisk through the stdlib logging module, for log aggregators, alongside the existing console output.

- src/snagline/sinks/logging_sink.py: JsonRiskFormatter renders exactly seven structural keys (ts, episode_id, step_id, trigger, severity, score, detail) as one compact line: json.dumps with sort_keys=True, separators=(",", ":"), ensure_ascii=False. Structure only, never event content (privacy rule 3 / project.md section 1.4). LoggingSink logs each rendered line on the "snagline" logger (default level WARNING, matching ConsoleSink's logging mode), fire-and-forget like every other sink.
- Fail-open (rule 2): if serialization raises, the formatter returns a plain ids-only fallback line (no detail); emit additionally swallows any downstream failure so nothing propagates into the host agent. Covered by an explicit sabotage test that monkeypatches json.dumps to raise.
- Config: new labeled block appends log_format: str = "text" (values text/json), wired through the existing field-driven from_env and load_file paths, selectable via SNAGLINE_LOG_FORMAT (12-factor).
- Exported from snagline.sinks; the top-level package __init__ does not export sinks today, so per the task's "check first" instruction that pattern was left unchanged.
- Docs: one row in the README sinks table (includes the config knob and env key) and one bullet in project.md section 8 sink list.
- Formatter record path: JsonRiskFormatter.format() renders records carrying snagline_risk as the same JSON line and passes plain records through default formatting untouched; both directions tested, plus custom formatter injection. New module is at 100% statement coverage.

## Verification

All commands below were run against a clean worktree of the branch tip (988f480), isolated from parallel agents' uncommitted changes in the shared checkout:

```
...................................                                      [100%]
251 passed, 1 skipped in 15.89s

$ ruff check src tests
All checks passed!

76 files already formatted

$ mypy src
Success: no issues found in 40 source files

$ git diff origin/master...HEAD | grep -n "<U+2014>"
(no output; the branch diff contains no em dash characters)
```

Mutation checks performed (rule 8), each confirmed red then reverted:

1. Flipped sort_keys=True to sort_keys=False in JsonRiskFormatter.render; test_output_is_single_line_sorted_and_compact went red on alphabetical key order.
2. Replaced the isinstance(risk, FailureRisk) branch in format() with a constant False; test_formatter_record_path_renders_json_for_risk_records went red with a JSONDecodeError.

Expected values are hardcoded against documented contracts (the seven-key set, severity_from_score thresholds of 0.8/0.5), never computed by calling the code under test.

Test coverage includes both sides: an injected loop stream through a real Monitor produces records whose parsed trigger is "loop"; a healthy 20-step unique-signature stream produces zero records.

Note on shared workspace: another agent works in this checkout in parallel; every commit was staged selectively so only files/hunks from this task's allowlist are included, verified per commit via git show --stat.

Board: #106

Fixes #99

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,7 @@ Sinks consume `FailureRisk` and escalate it. Only `FailureRisk` fields are ever 
 |:--|:--|:--|:--|
 | **Console** (default) | `sinks/console.py` | (built-in) | Writes `FailureRisk` as a JSON line to stderr. Zero dependency. |
 | **Webhook** | `sinks/webhook.py` | (built-in) | POSTs `FailureRisk` JSON via stdlib `urllib.request`. Fire-and-forget with a short timeout (default 2s). Silently ignores a dead endpoint. |
+| **Logging** | `sinks/logging_sink.py` | (built-in) | Emits one compact JSON object per risk on the `snagline` logger for log aggregators (`Config.log_format`: `text` or `json`, env `SNAGLINE_LOG_FORMAT`). Fail-open with a plain-text fallback if serialization breaks. Zero dependency. |
 
 Custom sinks implement the `AlertSink` protocol:
 
--- a/project.md
+++ b/project.md
@@ -595,6 +595,13 @@ user's infra choice, not this library's concern.
 
 - **`console.py`** (default): writes `FailureRisk` as a JSON line to
   stderr. Zero dependency.
+- **`logging_sink.py`** (issue #99): emits one record per `FailureRisk` on the
+  `"snagline"` logger; its formatter renders a single compact JSON object per
+  line with exactly the keys ts, episode_id, step_id, trigger, severity,
+  score, detail (structure only, never content). Selected via
+  `Config.log_format` (`text` or `json`, env `SNAGLINE_LOG_FORMAT`). If
+  serialization raises it falls back to a plain structural line: fail-open,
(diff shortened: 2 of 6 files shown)


Cyrax321:
Verified once more at HEAD (988f480) in a clean worktree: full gate green (251 passed, 1 skipped), ruff/mypy clean, CI green across py3.10 to py3.13. Merging.

PR #113, opened 2026-08-25
fix(tests): cross-platform SIGINT-free teardown reaps the hook bridge watch child

### LINKED ISSUE #69: test_hook_bridge: send_signal(SIGINT) fails on Windows and orphans the watch subprocess

Cyrax321:
## Summary

`tests/test_hook_bridge.py::test_watch_file_follow_sees_appended_lines` could not run on Windows: `Popen.send_signal(signal.SIGINT)` raises `ValueError: Unsupported signal: 2` there, the raise happened inside the `finally:` block, and the `snagline watch --follow` child was orphaned (issue #69).

Changes, all in `tests/test_hook_bridge.py` (no runtime code touched, POSIX behavior of the CLI is unchanged):

- `_graceful_stop()`: cross-platform stop ladder. Graceful step is SIGINT on POSIX (identical to Ctrl-C on the CLI, so its KeyboardInterrupt-suppressed summary still prints) and `terminate()` first on Windows where SIGINT does not exist. Then a shared escalation ladder: `wait(timeout)` -> `terminate()` -> `wait(timeout)` -> `kill()`, with every signal/wait guarded so no `ProcessLookupError` can escape even if the child already exited.
- A `children` fixture tracks every spawned `Popen` handle in a list (psutil-free) and asserts at teardown that every handle is reaped (`poll() is not None`), so zero lingering children can survive any test in this file.
- Cross-platform regression tests (no skip marks anywhere):
  - healthy path: follow-mode watcher dispatches its loop risk to a live 127.0.0.1 sink endpoint, then must be reaped (`poll()` non-None);
  - failure path: same watcher pointed at an endpoint that accepts connections and drops them, proving the cleanup ladder still reaps the child after failing webhook POSTs and that fail-open kept it alive (exit 0, full "ingested 4 step(s)" summary, logged-and-swallowed sink error);
  - already-exited child: stop must be a no-op, never raising ProcessLookupError;
  - a child that ignores SIGINT and SIGTERM proves the ladder escalates to SIGKILL on POSIX (`returncode == -SIGKILL`).
- Readiness synchronization uses deterministic signals only (server-side accept counting on ephemeral localhost ports plus a malformed-sentinel line observed via a stderr pump thread), no new sleeps.
- The original follow-mode test now stops through `_graceful_stop`. Its summary assertions stay POSIX-only: on Windows the teardown is terminate()-based by design (issue spec), so the CLI's SIGINT-path summary genuinely cannot exist there; reaping is asserted on every platform instead.

## Verification

Full gate, run against a clean worktree of this branch (origin/master + one commit):

```
239 passed, 1 skipped in 27.46s

$ ruff check src tests
All checks passed!

All checks passed!

$ mypy src
Success: no issues found in 39 source files

$ python -m pytest tests/test_hook_bridge.py -q -o addopts=""
.........                                                                [100%]
9 passed in 15.11s
```

Em dash scan of the diff (character U+2014, matched via printf escape rather than typed here): `git diff master...HEAD | grep -c "$(printf \\u2014)"` returns 0.

## Test honesty: mutation check performed

I deliberately broke the implementation and confirmed my own tests go red before reverting:

- Mutation: disabled the final `proc.kill()` rung in `_await_exit()`.
- Result: `test_graceful_stop_escalates_to_kill_when_the_child_ignores_signals` FAILED (child survived the ladder, `poll()` stayed None) and the `children` fixture raised "lingering child not reaped" during teardown: `1 failed, 8 passed, 1 error`.
- Reverted the mutation; suite green again.

Expectations are computed from documented behavior, never from the code under test: exactly one webhook POST per repetition episode comes from the shipped edge-triggered LoopDetector (issue #4 alert-spam fix), and the four ingested steps come from the four seeded fixture lines.

## Notes

- Scope: only `tests/test_hook_bridge.py`; no changes needed in `cli.py` or `adapters/claude_code.py`, so runtime behavior is identical everywhere including POSIX.
- Pre-existing note: two ruff errors visible in some runs of the shared working tree come from another agent's uncommitted files (`sinks/logging_sink.py`); they are not part of this branch. The clean-tree gate above is the authoritative result.

Board: #106

Fixes #69

### DIFF
--- a/tests/test_hook_bridge.py
+++ b/tests/test_hook_bridge.py
@@ -3,17 +3,27 @@
 * sidecar ``POST /hooks/claude-code`` (Claude Code native ``http`` hooks)
 * ``snagline hook`` (command-hook bridge: stdin payload -> file/HTTP/local)
 * ``snagline watch --file`` (file bridge)
+
+Teardown contract (issue #69): every spawned child stops through
+``_graceful_stop``, a cross-platform ladder (graceful signal on POSIX,
+terminate() on Windows, then wait -> terminate -> kill escalation), so no
+platform orphans the watch subprocess and no ProcessLookupError escapes.
 """
 
 from __future__ import annotations
 
 import json
+import os
 import signal
+import socket
 import subprocess
 import sys
 import threading
 import time
 import urllib.request
+from contextlib import suppress
+
+import pytest
 
 from snagline.monitor import Monitor
 from snagline.risk import FailureRisk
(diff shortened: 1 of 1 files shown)


Cyrax321:
All checks green (py3.10-3.13, title-check, labeler), no outstanding review feedback. One follow-up identified as out of scope here (and explicitly deferred by #69): adding windows-latest to the CI matrix so Windows-only regressions surface in CI -> tracked in #116.

Merging this now; it closes #69.

PR #139, opened 2026-08-26
ci(matrix): add windows-latest legs with a linux-only coverage gate

### LINKED ISSUE #116: ci: add windows-latest to the test matrix so Windows-only failures surface in CI
## Context

#69 fixed a Windows-only test failure (`Popen.send_signal(signal.SIGINT)` raising `ValueError` and orphaning the `snagline watch --follow` child) via PR #113. The fix makes `tests/test_hook_bridge.py` platform-safe and adds cross-platform teardown regressions, but CI still only exercises `ubuntu-latest`.

As the original issue author noted, this is deliberately left out of scope of that fix:

> Adding `windows-latest` to the CI matrix would keep this class of failure from recurring, but that is a separate change.

## Problem

- The suite now *runs* Windows-safe paths, but nothing verifies that on a real Windows runner; the terminate-based branches in the hook-bridge teardown tests are reasoned from CPython semantics, not executed anywhere.
- Any future Windows-only regression will again land unnoticed until a contributor on Windows hits a red suite on a clean checkout.

## Proposal

- Add `windows-latest` to `.github/workflows/ci.yml`, at least for py3.12 or py3.13 (full matrix optional if runner minutes are a concern).
- Watch for two friction points while doing so:
  - ephemeral localhost socket tests should be fine (stdlib sockets bound to 127.0.0.1); POSIX-only signal assertions are already gated on `os.name != "nt"`;
  - the coverage gate (`--cov-fail-under=80`) may need a per-OS aggregation decision so one OS cannot spuriously fail a shared check.

cc @Cyrax321

Refs #69, Refs #113

Cyrax321:
## Summary

Adds `windows-latest` to the test matrix so Windows-only regressions surface in CI (issue #116, deferred from #69/#113). The new legs immediately earned their keep: run two exposed a real intermittent Windows-only test failure, fixed here tests-only.

Decisions, including the per-OS coverage question the issue flags:

- **Matrix shape:** ubuntu legs stay exactly as today (py3.10-3.13); two windows-latest legs are added via `matrix.include` for py3.12 and py3.13 (the issue asks for at minimum these two). Job names now carry the OS.
- **Coverage policy: one canonical gate, on Linux.** Each matrix job checks its own coverage independently. The Linux legs enforce the single 80% threshold from `pyproject.toml`; the Windows legs run pytest-cov report-only, passing `--cov-fail-under=0` on the command line, which overrides the ini `addopts` value (proof below). Rationale: per-OS thresholds would be extra gates that must be moved in lockstep without adding signal, since the measured code (`src/snagline`) contains no platform-conditional branches at all. A correction here: an earlier version of this body and of the workflow comment claimed POSIX-gated test ladders would make Windows branch coverage fail spuriously; that mechanism was wrong (those ladders live in `tests/`, which does not feed the src coverage measurement), and the wording was fixed after the CodeRabbit review caught it (commit 76c60b8).
- **Lint and type-check stay Linux-only** (`if: runner.os == 'Linux'`): ruff and mypy results are OS-independent, so re-running them on Windows doubles runner minutes for zero extra signal. Tests are the thing Windows needs to exercise.
- **Install step is OS-safe:** venv activation picks `.venv/Scripts/activate` vs `.venv/bin/activate` via `$RUNNER_OS`, with `shell: bash` pinned explicitly (Git Bash exists on Windows runners), keeping the script identical on both families.
- **Windows-only flake found and fixed tests-only (commit b42d47c, per this issue's scope rule):** `tests/adapters/test_langchain_adapter.py::test_error_callbacks_carry_latency_ms` failed intermittently on `windows-latest` (green on run 1, red on run 2). Root cause: it injected `time.monotonic` and slept 10 ms between callback start and error, but `monotonic()` ticks at roughly 15.6 ms on Windows, so latency measured as 0.0 about half the time; the first green leg was luck. The test now scripts a fake clock through the handler's existing `clock=` injection point (nine readings: one per operation start, two per error callback for the latency end and event timestamp) and asserts exact binary-exact latencies (500/250/5000 ms) instead of a bare `> 0`. Deterministic across 30 consecutive local runs, removes three sleeps per the repo's no-sleeps rule, and any unexpected extra clock read fails loudly via StopIteration. No assertion was weakened; the check got stronger.

## Verification

Local full gate on the branch tip (b42d47c), macOS, CPython 3.14:

```
387 passed, 1 skipped in 30.41s
$ ruff check src tests && ruff format --check src tests
All checks passed!
97 files already formatted
$ mypy src
Success: no issues found in 47 source files
```

Flake-proofing evidence for the repaired test:

```
$ for i in $(seq 1 30); do pytest tests/adapters/test_langchain_adapter.py::test_error_callbacks_carry_latency_ms -q -o addopts=""; done | sort | uniq -c
  18 1 passed in 0.02s
   6 1 passed in 0.03s
   4 1 passed in 0.04s
   2 1 passed in 0.05s
```

(30/30 passes; previously the same test failed on the real windows-latest runner.)

Proof that `--cov-fail-under=0` genuinely overrides the ini `addopts` (single-file run where total coverage is 9.31%, far below the 80% bar):

```
$ python -m pytest tests/test_console_sink.py -q -p no:cacheprovider
FAIL Required test coverage of 80% not reached. Total coverage: 9.31%
3 passed in 0.39s        # exit=1

$ python -m pytest tests/test_console_sink.py -q -p no:cacheprovider --cov-fail-under=0
3 passed in 0.36s        # exit=0
```

CI evidence: run 1 all green including both Windows legs; run 2 caught the flake above (windows py3.12); run 3 expected green with the scripted clock.

Scope: `.github/workflows/ci.yml` plus one tests-only file (`tests/adapters/test_langchain_adapter.py`) under this issue's fix-Windows-failures-in-this-PR rule. No production code touched; no existing test modified beyond that repair, nothing skipped or deleted; nothing to mutation-check since neither change adds logic under test (the repaired test strengthens its assertion from `> 0` to exact values).

Fixes #116
Board: #106
Refs #69, #113

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,10 +17,11 @@ jobs:
         include:
           # Windows legs (issue #116): the two newest interpreters only, so
           # Windows-only regressions surface in CI without an 8-way matrix.
-          # Lint, type-check, and the coverage gate stay Linux-only (steps
-          # below): they are OS-independent checks, and branch coverage of
-          # POSIX-gated code paths would let one runner spuriously fail a
-          # shared --cov-fail-under threshold.
+          # Coverage policy: each job checks its own threshold independently;
+          # the Linux legs enforce the single canonical 80% gate from
+          # pyproject.toml while the Windows legs run report-only
+          # (--cov-fail-under=0 below), so per-OS numbers stay visible
+          # without becoming extra gates that must be moved in lockstep.
           - os: windows-latest
             python-version: "3.12"
           - os: windows-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,12 +7,24 @@ on:
 
 jobs:
   test:
-    name: Test (py${{ matrix.python-version }})
-    runs-on: ubuntu-latest
+    name: Test (${{ matrix.os }}, py${{ matrix.python-version }})
+    runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
+        os: [ubuntu-latest]
         python-version: ["3.10", "3.11", "3.12", "3.13"]
+        include:
+          # Windows legs (issue #116): the two newest interpreters only, so
+          # Windows-only regressions surface in CI without an 8-way matrix.
+          # Lint, type-check, and the coverage gate stay Linux-only (steps
+          # below): they are OS-independent checks, and branch coverage of
+          # POSIX-gated code paths would let one runner spuriously fail a
+          # shared --cov-fail-under threshold.
+          - os: windows-latest
+            python-version: "3.12"
+          - os: windows-latest
+            python-version: "3.13"
     steps:
       - uses: actions/checkout@v7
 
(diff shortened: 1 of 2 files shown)


Cyrax321:
Review catch addressed in 76c60b8: the coverage rationale was rewritten. The old wording implied a Windows runner could fail another job's threshold, which is impossible since each matrix job checks its own coverage independently, and the claimed mechanism (POSIX-gated test ladders depressing Windows branch coverage) was also wrong because those ladders live in tests/, not in the measured src tree. The comment now states the actual policy plainly: Linux legs carry the single canonical gate from pyproject.toml, Windows legs report coverage with --cov-fail-under=0 so per-OS numbers stay visible without becoming extra gates to move in lockstep.

Issue #145, opened 2026-08-26
[server] Optional mTLS knob (--client-ca) for the TLS sidecar listener


Cyrax321:
Split out of #120 (in-process TLS shipped without it, to keep that diff surgical).

## What

Expose an optional mutual-TLS knob on the sidecar listener, e.g. `snagline serve --client-ca PATH`: when given, load the CA bundle via `ssl.SSLContext.load_verify_locations()` and set `verify_mode = ssl.CERT_REQUIRED` on the server-side context created in `_resolve_ssl_context()` (`src/snagline/server/http_server.py`). Senders then authenticate by client certificate, letting the sidecar reject untrusted hosts at handshake time instead of at the token check.

## Why deferred

- #120's core value (closing the last plaintext hop) needed no client-CA plumbing; adding it would have grown the cert-generation test surface (CA + signed client pair) substantially.
- stdlib `ssl` fully supports it: zero new dependencies whenever it lands.

## Acceptance sketch

- Handshake succeeds with a client cert signed by `--client-ca`; fails (`sslv3 alert certificate required` / unknown CA) without or with an untrusted one.
- Plain and plain+server-TLS modes byte-for-byte unchanged when the flag is absent.
- Docs: threat-model paragraph in `docs/ATTACH_ANY_SYSTEM.md` currently says mutual TLS is "a natural follow-up"; flip when shipped.

Board: #106. Refs #120.

Cyrax321:
Claiming #145

Issue #103, opened 2026-08-25
[server] Sidecar TLS: reverse-proxy docs or stdlib ssl option


Cyrax321:
## Context
Known limitation: sidecar serves plain HTTP. Production needs a TLS story.

## Tasks
- [ ] Preferred: document reverse-proxy termination with copy-paste nginx AND caddy configs (bearer auth preserved, body-size cap noted)
- [ ] Alternative: stdlib `ssl.SSLContext` wrap option for `serve()` (`--certfile/--keyfile`) — zero new deps
- [ ] Security notes section update (§11)

## Acceptance
`curl https://...` against documented setup succeeds; auth still enforced; docs updated in ATTACH_ANY_SYSTEM + INTEGRATION_MATRIX.

Parallel-safe.

Cyrax321:
Docs-first PR is up: #110. It delivers the preferred reverse-proxy path: copy-paste nginx stable and Caddy v2 TLS-termination configs (bearer auth preserved, client_max_body_size / request_body max_size matched to the sidecar's 1000000-byte default, streaming notes), a threat-model paragraph contrasting termination with the in-process stdlib ssl.SSLContext wrap option for serve(), and a hardening checklist (token rotation, loopback bind, body-cap sync, no-content retention). Transport rows added to INTEGRATION_MATRIX.md and the README security bullet updated to state the plain-HTTP limitation. The stdlib ssl wrap of serve() remains documented as the future option; zero new dependencies either way.

Cyrax321:
Merged via #110. Master now carries the nginx and Caddy TLS termination configs, the transport-security rows in the integration matrix, and the updated sidecar security bullet; the stdlib ssl.SSLContext wrap of serve() stays documented as the future option. Closing as completed.

Issue #88, opened 2026-08-24
[detector] SideEffectGuardDetector: duplicate non-idempotent action detection


Cyrax321:
## Summary

Add `SideEffectGuardDetector`: detect repeated *non-idempotent* actions within an episode. Adapters mark steps with `metadata["side_effect"] = True` (and optionally a stable idempotency key); the detector keeps a per-episode count of side-effect signatures and fires HIGH-severity on any repeat. This catches duplicate sends/payments/deploys, the single most expensive failure class in production agents.

## Why

- The durable-execution literature is blunt about this: replaying a crashed run re-fires non-idempotent operations, and "for a non-idempotent one, a second commit, a duplicate ticket, a repeated payment, the store and the external world diverge silently" (AI-as-Infrastructure, memory/state systems chapter). Reported agent retry rates of 15-30% make re-execution a design regime, not an accident.
- CONTINUUM solves this at the harness boundary with its claim ledger, but explicitly documents the bypass: gate does not see inside shell commands (Bash/curl escape structured-tool claims). An independent detection signal closes part of that gap from the observation side.
- Zero-dep fit: this is just a filtered counting window. No new dependencies, no content retention (signatures are hashes).

## Schema decision needed

Option A (preferred): keep StepEvent frozen; document the convention `metadata={"side_effect": True}` and have the detector read metadata *only for this boolean* (documented exception to "detectors never read metadata", justified because it is a flag, not content).
Option B: add `side_effect: bool = False` as a proper optional field. Cleaner contract, touches schema + all adapters' constructors trivially. Recommend B.

## Design

```python
class SideEffectGuardDetector:
    name = "side_effect_guard"
    def __init__(self, allowed_repeats=1, severity=0.9): ...
    def observe(self, event):
        if not getattr(event, "side_effect", False):
            return None
        key = (event.episode_id, event.tool_name, event.action_signature)
        self._counts[key] += 1
        if self._counts[key] > self.allowed_repeats:
            return FailureRisk(..., "side_effect_duplicate",
                f"non-idempotent {event.tool_name} fired {self._counts[key]}x")
        return None
```

Deliberately stricter than the loop detector: a *single* repeat of a payment/send is already actionable (allowed_repeats=1 default means fire on the 2nd occurrence).

## Acceptance criteria

- [ ] Two identical side-effect steps fire immediately; non-side-effect repeats never do.
[...]

PR #146, opened 2026-08-26
feat(server): in-process TLS termination for the sidecar listener

### LINKED ISSUE #120: [server] In-process TLS for serve(): stdlib ssl.SSLContext wrap (--certfile/--keyfile), zero new deps
## Context

Issue #103 was closed by PR #110 with its preferred half delivered: reverse-proxy TLS termination is now the documented production pattern (nginx and Caddy copy-paste configs in `docs/ATTACH_ANY_SYSTEM.md`, transport rows in `docs/INTEGRATION_MATRIX.md`, README security bullet updated). This issue tracks the alternative half that #103 also named and that remains unbuilt: terminating TLS inside the sidecar process so no plaintext hop exists at all.

cc @Cyrax321 (split out of #103 per the no-duplication sweep; nothing else in the tracker covers this).

## Why

Reverse-proxy termination leaves one plaintext segment: proxy to sidecar over loopback. That is acceptable when both share a trusted host (the documented topology), but not when the proxy runs on a different machine, or when the shared host runs untrusted processes. In-process TLS closes that last hop and additionally enables mutual TLS, letting the sidecar authenticate senders by client certificate without any external component.

Zero new dependencies either way: `ssl` is standard library, which is exactly why #103 listed this as the documented future option (referenced from the merged docs as such).

## Tasks

- [ ] `make_server` / `serve` accept an `ssl_context: ssl.SSLContext | None = None` (or equivalent `certfile`/`keyfile` pair) and wrap the listening socket via `ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)` + `wrap_socket()` when provided
- [ ] CLI: `snagline serve --certfile PATH --keyfile PATH`; absent flags keep today's plain HTTP behavior byte-for-byte
- [ ] Auth unchanged over TLS: Bearer / `X-Snagline-Token` still required on everything except `GET /health`
- [ ] Optional mTLS knob (`--client-ca`) for certificate-based sender authentication; can be a follow-up if it bloats the diff
- [ ] Tests: HTTPS handshake against an ephemeral loopback port with a self-signed pair generated at test time (`ssl` + `subprocess openssl`, or `trustme`-free stdlib path), auth enforced over TLS, plain-HTTP mode untouched
- [ ] Docs: flip the INTEGRATION_MATRIX transport row and the ATTACH_ANY_SYSTEM threat-model paragraph from "future option" to shipped, with a copy-paste serve invocation

## Acceptance

- `curl https://127.0.0.1:<port>/health -k` succeeds against a cert/key-started sidecar
- Bearer auth still enforced over the TLS listener (401 without token)
- Plain HTTP mode behaves exactly as before when flags are absent
- Zero new third-party dependencies; full gate green (pytest, ruff, mypy)

Follows #110. Parent thread: #103 (closed). Board: #106.

Cyrax321:
## Summary

Closes the last plaintext hop: `snagline serve` can now terminate TLS itself with stdlib `ssl`, no reverse proxy required, zero new dependencies (issue #120, the implementation half of #103).

- `make_server()` / `serve()` accept `ssl_context=` or a `certfile=`/`keyfile=` pair; when given, the listening socket is wrapped via `ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)` + `wrap_socket(server_side=True)`. Contradictory or incomplete TLS args (`ssl_context` together with files, bare `keyfile`) raise `ValueError` at startup rather than silently binding plaintext.
- CLI: `snagline serve --certfile PATH --keyfile PATH`. With the flags absent, `_cmd_serve` calls `serve()` without any TLS kwargs at all and prints the identical plain-HTTP banner: byte-for-byte unchanged behavior. The startup banner advertises `https://` when TLS is on.
- Auth semantics are untouched by TLS: Bearer / `X-Snagline-Token` stays enforced on everything except `GET /health` (verified over the TLS listener below).
- mTLS (`--client-ca`) deferred to #145 to keep this diff surgical; stdlib supports it with no new deps whenever it lands.
- Docs flipped from "future option" to shipped: INTEGRATION_MATRIX transport row + trailing paragraph, ATTACH_ANY_SYSTEM threat-model paragraph (now includes a copy-paste serve invocation + `curl -k` acceptance check), the hardening checklist bullet that said "until in-process TLS lands", and the stale "not in-process" progress notes.

Scope note: README's sidecar security bullet still reads "In-process TLS is the documented future option (#103)". README.md is outside this task's allowed file list, so it was left alone; it needs the same one-line flip after merge.

Commits: feat(server), feat(cli), docs(sidecar), test(server). Branch is rebased on current origin/master (#141 included).

## Verification

Full gate on the rebased branch:

```
.......................................                                  [100%]
399 passed, 2 skipped in 32.82s

$ ruff check src tests
All checks passed!
98 files already formatted
$ mypy src
Success: no issues found in 47 source files
```

Acceptance from the issue, against a real cert-started sidecar (self-signed pair generated with openssl, ephemeral port):

```
$ SNAGLINE_SERVE_AUTH_TOKEN=t0psecret python -m snagline.cli serve \
    --host 127.0.0.1 --port 51683 --certfile /tmp/sng_cert.pem --keyfile /tmp/sng_key.pem
snagline serve: listening on https://127.0.0.1:51683 (POST /events, GET /health)

$ curl -sk https://127.0.0.1:51683/health
{"status": "ok"}

$ curl -sk -o /dev/null -w "%{http_code}\n" -X POST https://127.0.0.1:51683/events \
    -H 'Content-Type: application/json' -d '{...valid StepEvent...}'        # no token
401

$ curl -sk -X POST https://127.0.0.1:51683/events \
    -H 'Content-Type: application/json' -H 'Authorization: Bearer t0psecret' -d '{...}'
{"status": "ingested", "step_id": "1"}
```

Plain-mode regression through the CLI:

```
$ python -m snagline.cli serve --host 127.0.0.1 --port 8799
snagline serve: listening on http://127.0.0.1:8799 (POST /events, GET /health)
$ curl -s http://127.0.0.1:8799/health
{"status": "ok"}
```

Mutation check (required by the task): dropped the `wrap_socket` call in `make_server` on purpose and reran the suite:

```
FAILED tests/test_server_tls.py::test_tls_handshake_serves_health_over_https
FAILED tests/test_server_tls.py::test_make_server_accepts_ready_ssl_context
FAILED tests/test_server_tls.py::test_auth_is_enforced_over_the_tls_listener
3 failed, 6 passed in 2.47s
```

Reverted the break; full suite green again (numbers above). That mutation run also exposed a test-hygiene bug in my own new tests (a failing assert before `serve_forever` started made `shutdown()` wait forever); fixed in the fourth commit so broken implementations fail red in seconds instead of hanging.

Test honesty: all new tests live in `tests/test_server_tls.py`; no existing test was modified, skipped, or weakened (`tests/test_cli.py` untouched and green). Detector-style both-sides coverage: auth-over-TLS asserts 401 without a token AND 202 with it plus open `/health`; plain-mode control asserts no SSLSocket and healthy HTTP. Certificates are generated at test time via subprocess openssl (skipped only if the binary is missing); servers bind ephemeral 127.0.0.1 ports; no sleeps, no network beyond loopback, no randomness in assertions. Em-dash check: grepping every commit diff (`git log origin/master..HEAD -p`) and this PR body for U+2014 returns zero matches.

Fixes #120. Board: #106.

## Summary by CodeRabbit

* **New Features**
  * Added optional in-process TLS for the sidecar using certificate and private-key configuration.
  * Startup output now indicates whether the server is using HTTP or HTTPS.
  * Existing plain HTTP behavior remains unchanged when TLS is not configured.

* **Documentation**
  * Updated deployment, threat-model, and integration guidance for sidecar TLS.
  * Clarified that mutual TLS is not currently supported.

### DIFF
--- a/docs/ATTACH_ANY_SYSTEM.md
+++ b/docs/ATTACH_ANY_SYSTEM.md
@@ -82,8 +82,9 @@ enterprise-grade alerting. Concretely:
    `StepEvent`, add auth (HMAC or bearer) and TLS verification, and document
    the schema.
    - **Done:** bearer/token auth (#31), body-size cap + batched `POST /events`
-     (#32), 12-factor config + CLI wiring (#36). TLS termination is delegated
-     to a reverse proxy (not in-process) - documented as such.
+     (#32), 12-factor config + CLI wiring (#36). TLS terminates at a
+     reverse proxy (configs below) or in-process via `--certfile/--keyfile`
+     (issue #120).
 2. **Add Python auto-instrumentation.** Wrap OpenAI/Anthropic clients and
    LangChain automatically so a user adds one line (`import snagline.auto`)
    instead of editing every call. This is the real "attach to any system"
@@ -154,7 +155,9 @@ but the moment telemetry crosses a network segment you do not fully trust
 The supported production pattern is TLS termination at a reverse proxy:
 public traffic arrives at the proxy over HTTPS, the proxy strips TLS and
 forwards plaintext to the sidecar over loopback. Both configs below are
-copy-paste ready and introduce zero new Python dependencies.
+copy-paste ready and introduce zero new Python dependencies. If you cannot
+run a proxy on the same host, the sidecar can also terminate TLS itself;
+see the threat model below for the trade-off and the invocation.
 
 Assumptions used throughout: the sidecar runs on the same host as the proxy,
 on port 8787, with a bearer token set via `$SNAGLINE_SERVE_AUTH_TOKEN` (or
@@ -268,15 +271,34 @@ caller reached the intended endpoint. What it does not solve is the final
 hop: proxy to sidecar remains plaintext. On one trusted host across
 loopback, that residual exposure is usually acceptable. It is not acceptable
 when the proxy runs on a different machine than the sidecar, or when the
-shared host runs processes you do not control; in those cases wrap TLS in
-the sidecar process itself. That is the documented future option (issue
-#103): wrap `serve()`'s listening socket with a stdlib `ssl.SSLContext`
-exposed via `--certfile/--keyfile` flags. It would close the last plaintext
-hop and additionally allow mutual TLS, letting the sidecar authenticate
-clients by certificate without any external component. Both directions cost
-zero new dependencies: `ssl` is standard library, and a reverse proxy adds
-nothing to the Python environment, so the choice between them is purely
-operational.
+shared host runs processes you do not control; in those cases terminate TLS
+in the sidecar process itself. That is built in as of issue #120: pass
+`--certfile`/`--keyfile` and `serve()` wraps its listening socket with a
+stdlib `ssl.SSLContext(ssl.PROTOCOL_TLS_SERVER)`, closing the last plaintext
+hop:
+
+```bash
+# Any PEM pair works; to generate a throwaway self-signed one:
+#   openssl req -x509 -newkey rsa:2048 -nodes -days 365 \
+#     -subj "/CN=snagline.internal" -keyout key.pem -out cert.pem
+SNAGLINE_SERVE_AUTH_TOKEN=... snagline serve \
+  --host 127.0.0.1 --port 8787 \
+  --certfile /etc/snagline/cert.pem \
+  --keyfile /etc/snagline/key.pem
+
+# Acceptance check against a self-signed pair (-k). With a real certificate,
+# use --cacert so the chain is actually verified instead of skipped:
+curl -k https://127.0.0.1:8787/health
+```
+
+Auth semantics are unchanged over TLS: `Authorization: Bearer` and
+`X-Snagline-Token` stay required on everything except `GET /health`, exactly
+as on plain HTTP. What in-process termination does not add today is mutual
+TLS (authenticating senders by client certificate); stdlib `ssl` can do it
+with no new dependencies, so exposing a client-CA knob is a natural follow-up.
+Both directions cost zero new dependencies: `ssl` is standard library, and a
+reverse proxy adds nothing to the Python environment, so the choice between
+them is purely operational.
 
 ### Hardening checklist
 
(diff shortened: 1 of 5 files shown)


Cyrax321:
## Review follow-up (all four findings addressed)

Rebased on current master (#88 SideEffectGuardDetector + README accuracy table included); gate re-run locally after rebase: 428 passed / 2 skipped, ruff clean, mypy clean. New commits: `fix(server)`, `fix(cli)`, `docs(sidecar)`.

**1. Handshake on the accept thread serializes connections (Gitar, security): valid, fixed in code.** The listener now stays a plain socket; `finish_request` wraps each accepted connection inside ThreadingMixIn's own worker thread, so a stalled handshake ties up one daemon thread instead of the accept loop. Failed handshakes (plaintext probes, scans) are logged at debug and dropped; serving continues untouched (fail-open). Plain mode still constructs the exact same `ThreadingHTTPServer` as before. Tests updated: the listener-raw invariant is asserted directly, plus two new cases: a stalled peer that never completes its handshake must not block another client's `/health`, and a plaintext probe to the TLS port must get no HTTP answer while serving survives. Mutation check repeated for the new shape (wrap dropped in `finish_request`): all 5 TLS tests red in ~4s, plain/validation/CLI tests stay green.

**2. Bare `--keyfile` banner then traceback (Gitar): valid, fixed.** `_cmd_serve` validates the pairing before printing anything and exits 2 with a one-line reason; a wiring test asserts `serve()` is never called in that case.

**3. `--cacert` hostname mismatch in the docs example (CodeRabbit): valid, docs updated.** The acceptance snippet now shows `--resolve snagline.internal:8787:127.0.0.1` alongside `--cacert` for chain-verified local checks, keeping the `-k` self-signed check as-is.

**4. Missing skip marker on `test_ssl_context_and_certfile_together_are_rejected` (CodeRabbit): valid, fixed.** It now carries `@requires_openssl` like every other cert-generating test.

Issue #130, opened 2026-08-25
[server] No read timeout on sidecar connections: a stalled sender can pin handler threads indefinitely


Cyrax321:
## Context

Found while implementing #121: the over-cap drain is now bounded in bytes, but nothing bounds it (or any body read) in time.

## Problem

The sidecar never sets a socket timeout. `BaseHTTPRequestHandler` inherits `timeout = None` from `StreamRequestHandler`, so every `rfile.read(n)` blocks until exactly `n` bytes arrive or the peer closes. A client that declares `Content-Length: 500000` and then sends only a few bytes pins one handler thread of the `ThreadingHTTPServer` for as long as it likes. Verified against master (`a1b2367`) on an ephemeral loopback port:

```
send: POST /events HTTP/1.0 ... Content-Length: 500000 ... + 7 body bytes
wait: no response after 3.0s (test gave up; handler still blocked in read)
```

Repeated from many connections this is a slowloris-style resource retention path against the sidecar process.

## Scope

- Applies to normal accepted bodies (`_read_body`) and to the bounded over-cap drain added in #121 alike: size-bounded, not time-bounded.
- Pre-existing behavior, not introduced by #121; filing separately so the drain fix stays revertible on its own merits.

## Proposed remedy

Set a read timeout on the handler class (`_Handler.timeout = <seconds>`), which stdlib applies to the connection and `handle_one_request` already converts into a logged timeout plus connection close. Pick a default comfortably above legitimate slow-sender needs, document it next to `--max-body-bytes`, and optionally expose it via env/config like the other serve knobs. Add a raw-socket test asserting a stalled sender's thread is released and the socket is closed by the server within the budget (use a small explicit timeout value in the test; no sleeps beyond joining the handler).

Zero new dependencies either way.

Part of the execution board tracking: #106

Cyrax321:
Claiming #130

Cyrax321:
Done: PR #186 merged (all CI green, CodeRabbit rate-limited pass, Gitar passed). Read timeout now enforced via Handler.timeout with config/env knob.

Issue #129, opened 2026-08-25
[server] Malformed Content-Length crashes the handler thread; clients get a dropped connection instead of 400


Cyrax321:
## Context

Found while implementing #121 (the over-cap drain lands right next to this line), but this failure mode is distinct and was not covered there.

## Problem

`do_POST` parses the request length with:

```python
length = int(self.headers.get("Content-Length") or 0)
```

A malformed value such as `Content-Length: abc` raises `ValueError` inside the handler. `socketserver` prints a full traceback per bad request and drops the connection, so the client gets an empty reply instead of a clean `400`. Verified against master (`a1b2367`) on an ephemeral loopback port:

```
Exception occurred during processing of request from ('127.0.0.1', 62449)
...
  File "src/snagline/server/http_server.py", line 374, in do_POST
    length = int(self.headers.get("Content-Length") or 0)
ValueError: invalid literal for int() with base 10: 'abc'

client: <empty reply, connection dropped>
```

## Why it matters

- Any health-checker, proxy, fuzzer, or buggy client that sends a non-numeric length produces server-side traceback noise (unbounded log spam under repetition) and an opaque transport error client-side.
- It contradicts the fail-open spirit of the project: a bad input should become a `400` response, not an exception escaping into the server plumbing.

## Proposed remedy

Wrap the parse in a small helper shared by `do_POST` and `_read_body`: catch `(ValueError,)`, respond `400 {"error": "invalid Content-Length"}`, and return without reading. Fail-open, zero deps, a few lines plus a raw-socket regression test asserting the client receives the 400 status line.

Related: requests with no `Content-Length` at all are currently treated as length 0 (body silently empty), which is acceptable for this endpoint but could be noted in the same helper's docstring.

Part of the execution board tracking: #106

Cyrax321:
Claiming #129

Cyrax321:
Done: PR #185 merged (all CI green, CodeRabbit and Gitar passed). Malformed Content-Length now returns 400 without reading.

PR #160, opened 2026-08-26
feat(server): end-of-episode signal expires the episodes-active gauge

### LINKED ISSUE #123: [server] End-of-episode signal so snagline_episodes_active can expire properly
## Context
Follow-up from the Prometheus exposition work (#98, delivered in #115).

## Current behavior
The sidecar gauge `snagline_episodes_active` counts distinct episode ids seen through POST /events since process start. The wire format has no end-of-run signal, so the sidecar cannot know when an episode finished. To stay bounded, the collector keeps at most 10,000 ids with least-recently-seen eviction: reseeing an id refreshes its position, so under pressure the quietest episode is forgotten first.

This is documented behavior (HELP text and docstrings), honest, O(1) amortized per step, and never retains content (ids only). But as a gauge of *active* episodes it is an approximation: long-lived processes that see more than 10k episodes will under-count old-but-still-running ones.

## Proposal
Give hosts a way to signal episode completion so the gauge can expire entries deterministically. Options, roughly in order of fit:
1. A `POST /episodes/end` endpoint (or a `{"type": "end_episode"}` control event on /events) that forwards to `Monitor.end_episode(episode_id)` and removes the id from the metrics table.
2. An optional TTL config key that expires ids not seen for N seconds (wall-clock based; replay traffic would need care).

## Acceptance
- Ended episodes stop being counted without waiting for cap eviction.
- Bounded memory preserved; no content retained; fail-open on malformed signals.
- Documented in ATTACH_ANY_SYSTEM next to the scrape snippet.

Cyrax321:
Closes #123

## What

New endpoint on the sidecar:

    POST /episodes/end        body: {"episode_id": "<id>"}     -> 200 {"status": "ended", ...}

It forwards the id to `Monitor.end_episode(episode_id)` (running detector
finalizers plus per-episode state teardown) AND removes the id from the
`SidecarMetricsCollector` episodes table, so `snagline_episodes_active`
stops counting a finished episode immediately instead of waiting for
least-recently-seen cap eviction.

## Why

The gauge had no deterministic expiry: the wire format had no end-of-run
signal, so long-lived processes seeing more than 10k episodes under-counted
old-but-still-running ones. This gives hosts an explicit signal (issue
#123).

## Design decisions

- Dedicated `POST /episodes/end` chosen over the `{"type": "end_episode"}`
  control-event variant on `/events`: one unambiguous endpoint, no risk of
  colliding with batched StepEvent arrays on the ingest path.
- **TTL expiry not implemented**, deliberately: it needs per-id wall-clock
  tracking plus sweep timing and replay-traffic caveats. Not trivial, and
  the issue says to skip it in that case. The explicit signal covers the
  deterministic case; the LRU cap still bounds memory when hosts never
  send the signal.
- Auth applies exactly like every other POST (the shared `_authorized`
  check at the top of `do_POST`; `/health` stays the only open route).
- Fail-open end to end: malformed bodies get 400, a raising
  `Monitor.end_episode` is logged and swallowed (covers misconfigured
  monitors with `fail_open=False`), collector bookkeeping is guarded, and
  the client always gets a clean JSON response.
- No content retention: ids only, bodies are never logged or stored.
  Ending an unknown or already-ended id is an idempotent no-op.
- Gauge HELP text updated to mention the removal behavior.

## Docs

`docs/ATTACH_ANY_SYSTEM.md`, directly after the Prometheus scrape config
sentence: what the signal does, auth note, idempotency note, and a curl
example against the default port.

## Acceptance evidence

- Ended episodes stop being counted without waiting for cap eviction:
  raw-socket flow `test_end_signal_drops_episode_from_gauge_immediately`
  (ingest -> gauge 1, POST /episodes/end -> gauge 0), unit-level
  `test_collector_expiry_does_not_wait_for_cap_eviction`,
  `test_end_is_idempotent_and_ids_only`.
- Forwarding to Monitor.end_episode: `test_end_forwards_to_monitor_end_episode`.
- Auth like every POST: `test_auth_applies_to_end_endpoint_like_every_post`.
- Malformed signals: 400, never crash, never retain:
  `test_malformed_signals_get_400_and_sidecar_stays_up` (bad JSON,
  non-dict body, missing/empty/non-string id, missing Content-Length).
- Monitor failure cannot break serving: `test_monitor_failure_never_breaks_the_endpoint`.
- Bounded memory preserved: table cap unchanged (`_MAX_TRACKED_EPISODES`),
  eviction tests from #98 still pass.
- Full gate at this commit: 505 passed, 2 skipped (coverage 87.89%),
  `ruff check src tests`, `ruff format --check src tests`,
  `mypy src` all clean.

## Mutation checks (fix committed first, then mutated, then reverted)

- Disabled the 400 responses in `_post_episodes_end` ->
  `test_malformed_signals_get_400_and_sidecar_stays_up` went red.
- Disabled the table pop in `SidecarMetricsCollector.end_episode` ->
  `test_end_signal_drops_episode_from_gauge_immediately` went red.

Refs #106

## Summary by CodeRabbit

* **New Features**
  * Added an authenticated endpoint to mark episodes as ended.
  * Ended episodes are finalized promptly and removed from active episode metrics.
  * The endpoint safely handles repeated requests and unknown episode IDs.
  * Added validation for malformed requests with clear client error responses.

* **Documentation**
  * Documented the new episode-ending endpoint and its behavior.

### DIFF
--- a/docs/ATTACH_ANY_SYSTEM.md
+++ b/docs/ATTACH_ANY_SYSTEM.md
@@ -136,6 +136,26 @@ enterprise-grade alerting. Concretely:
     or `SNAGLINE_METRICS_FORMAT=classic`. Scrape config: point Prometheus at
     the sidecar with `metrics_path: /metrics` (same port and auth token as
     the event endpoints). Structured logging remains.
+
+    **End-of-episode signal (#123):** `snagline_episodes_active` counts
+    distinct episode ids seen through the sidecar since process start
+    (capped at 10,000, least-recently-seen eviction). A host that knows an
+    episode is finished should tell the sidecar so the gauge drops the id
+    immediately instead of waiting for cap eviction:
+
+    ```bash
+    curl -X POST http://127.0.0.1:8787/episodes/end \
+      -H "Authorization: Bearer $SNAGLINE_SERVE_AUTH_TOKEN" \
+      -H "Content-Type: application/json" \
+      -d '{"episode_id": "run-42"}'
+    # -> {"status": "ended", "episode_id": "run-42"}
+    ```
+
+    The endpoint runs behind the same auth token as every other POST,
+    forwards to `Monitor.end_episode()` (running detector finalizers and
+    releasing per-episode detector state), and answers `400` on malformed
+    bodies without crashing or retaining anything beyond the id itself.
+    Ending an unknown id is an idempotent no-op.
 11. Integration matrix document plus a prominent "10-line custom adapter"
     path. **Done:** `docs/INTEGRATION_MATRIX.md` (#49).
 
--- a/src/snagline/server/http_server.py
+++ b/src/snagline/server/http_server.py
@@ -6,8 +6,15 @@ shell script, a Claude Code hook) can POST ``StepEvent`` JSON to:
     POST /events             body: a StepEvent as a JSON object      -> monitor.ingest()
                                or a JSON array of StepEvent objects     (batched)
     POST /hooks/claude-code  body: a native Claude Code hook payload -> mapped + ingested
+    POST /episodes/end       body: {"episode_id": "<id>"}            -> monitor.end_episode()
     GET  /health                                                     -> 200 OK
     GET  /metrics                                                    -> Prometheus text exposition (v0.0.4)
+
+A host that knows an episode is over can signal it with ``POST
+/episodes/end`` carrying ``{"episode_id": "..."}`` (issue #123): the sidecar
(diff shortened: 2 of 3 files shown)


Cyrax321:
Status note: all CI legs are green on the latest commit, Gitar approved, and CodeRabbit's only finding (docstring coverage) was addressed in 4475bcd. This PR is the first of a strictly sequential server-hardening triple (#123 then #129 then #130), so it currently gates two follow-up fixes. Ready for merge whenever you have a moment.

Issue #121, opened 2026-08-25
[server] Over-cap POSTs: reply 413 without draining the body, direct clients can see a reset instead of the status


Cyrax321:
## Context

Found during empirical verification of PR #110 (sidecar behavior tested against a real server on an ephemeral loopback port). `http_server.py` checks `Content-Length` against `max_body_bytes` and calls `_respond(413)` without reading the request body. The handler then closes the connection with megabytes of unread data still in flight, so the peer frequently gets `EPIPE` / connection reset while or before reading the status line.

cc @Cyrax321 (no existing issue covers this; searched open and closed for 413, payload too large, body, drain, broken pipe).

## Reproduction (verified against current master)

POST `/events` with valid auth headers and `Content-Length: 1000001`:

- Server side sends `HTTP/1.0 413 Content Too Large` correctly.
- A client streaming the body (e.g. Python `urllib`, `curl` on slow links) typically observes `BrokenPipeError` / reset instead of the 413, because the socket is closed while the request body is still being sent.

Exactly-at-cap bodies (1000000 bytes) are accepted: the cap itself is byte-exact and correct.

## Impact

Low in the documented production topology: nginx/Caddy reject over-cap requests on `Content-Length` at the edge, so proxied callers always see a clean 413. The papercut only bites direct plaintext clients to a loopback sidecar, which see a confusing transport error instead of the intended 413.

## Possible remedies (pick one)

- Drain and discard up to `min(Content-Length, small cap)` bytes before responding; respond 413 immediately without draining when the remainder exceeds the cap
- Or respond 413 with `Connection: close`, then `shutdown(SHUT_RD)` and linger briefly so the client can finish sending and read the response

## Acceptance

- [ ] A direct client that streams an over-cap body can reliably read the 413 status line (test via raw socket on an ephemeral loopback port)
- [ ] At-cap bodies still accepted; memory stays bounded regardless of claimed Content-Length
- [ ] No new dependencies; full gate green

Follows #110 (found while verifying it). Related hardening context in `docs/ATTACH_ANY_SYSTEM.md`.

Cyrax321:
PR is up: over-cap bodies are drained within a small window (chunked, memory-bounded) before the 413 goes out, so direct clients read the status line instead of a reset.

PR #176, opened 2026-08-26
fix(server): 400 on malformed Content-Length instead of a handler traceback

### LINKED ISSUE #129: [server] Malformed Content-Length crashes the handler thread; clients get a dropped connection instead of 400

Adhi1-2:
Fixes #129.

## Summary of Changes

`do_POST` and `_read_body` both parsed the request length with a bare
`int(self.headers.get("Content-Length") or 0)`. Both now go through one
helper, `_content_length()`, which returns `0` for an absent header (the
existing reading, unchanged) and `None` for anything unusable. `do_POST`
resolves framing before auth and routing and answers
`400 {"error": "invalid Content-Length"}` with the connection closed.

Two values were broken, not one:

| `Content-Length` | before | after |
| --- | --- | --- |
| `abc`, `0x10`, `1e3` | `ValueError` in `do_POST`, traceback, empty reply | `400` |
| `-5` | passed the cap check, then `ValueError` in `rfile.read(-5)`, RST | `400` |
| absent | length 0, empty body | unchanged |
| `42` | read 42 bytes | unchanged |

Verified on upstream/master (`6a2757a`) before the fix, on an ephemeral
loopback port:

```
non-numeric 'abc': b'<EMPTY REPLY>'
negative '-5':     recv raised ConnectionResetError: [Errno 54] Connection reset by peer
hex '0x10':        b'<EMPTY REPLY>'
```

and after:

```
non-numeric 'abc': b'HTTP/1.0 400 Bad Request'
negative '-5':     b'HTTP/1.0 400 Bad Request'
hex '0x10':        b'HTTP/1.0 400 Bad Request'
```

### Two judgement calls worth reviewing

**Framing is settled before the auth check.** Every branch after it either
reads or drains the body, and none of them can know where an unparseable
body ends, so the 401 drain added in #121 would have been the next thing to
raise. The 400 reveals nothing an unauthenticated caller could not infer
from the header it sent itself, and path probing stays behind the token: an
unknown path with bad framing gets 400, never 404. Both are pinned by
`test_bad_content_length_is_rejected_before_auth_and_routing`.

**The 400 needs a bounded drain to actually arrive.** Closing with unread
inbound data makes the kernel answer RST, which discards the response the
client had not read yet -- exactly the race #121 solved for over-cap bodies
by draining the declared length. That length does not exist here, so
`_drain_unknown_body` bounds the read by wall clock instead
(`_BAD_FRAMING_DRAIN_SECONDS = 0.05`) plus the usual byte cap, reads the raw
connection so only the kernel buffer is touched, restores the previous
socket timeout, and treats every error as "stop". Measured on loopback,
200 requests per setting:

```
deadline=0.0    400-delivered=169/200  median=  0.09ms  p95=  0.13ms
deadline=0.005  400-delivered=200/200  median=  6.07ms  p95=  6.29ms
deadline=0.025  400-delivered=200/200  median= 26.69ms  p95= 27.00ms
deadline=0.05   400-delivered=200/200  median= 51.93ms  p95= 52.23ms
deadline=0.25   400-delivered=200/200  median=252.05ms  p95=252.37ms
```

A purely non-blocking drain loses the 400 roughly 15% of the time; any
nonzero deadline delivered all 200. A client waiting on the response never
closes first, so the deadline is paid in full on every malformed request --
50ms is the tradeoff picked between delivery and how long one bad sender
holds a handler thread. Easy to retune; it is a single module constant, and
the timing assertion in the tests is deliberately loose so CI noise cannot
make it flaky.

This is deliberately time-bounded only on this one path, via a timeout set
and restored inside the call, so it does not pre-empt or depend on #130.

## Justification

Any health checker, proxy, fuzzer, or buggy client emitting a non-numeric
length produced a full traceback per request (unbounded stderr noise under
repetition) and an opaque transport error client-side. It also contradicted
the fail-open spirit of the project: bad input should become a 400, not an
exception escaping into the server plumbing.

## Kind of Contribution

Bug Fix

## Unit Tests

`tests/server/test_content_length_400.py` (11 cases), following the
raw-socket style of `tests/server/test_413_drain.py` because no client
library will emit an invalid `Content-Length` for you:

- non-numeric and negative values each get a 400 status line on the wire
- bad framing is resolved before auth and before routing (401 and 404 paths)
- `capfd` asserts no `Traceback` or `ValueError` reaches stderr across
  `abc`, `0x10`, `-5`, `1e3`, `1 2`
- absent header still reads as an empty body; a well-formed length still
  gets 202
- helper table for the classification, including `" 42 "` and `-1`
- `_read_body` returns `b""` rather than raising if ever reached directly
- the drain: byte bound, chunk size, timeout restored, peer hangup, spent
  deadline, and a mid-drain `ConnectionResetError`

`tests/test_server_tls.py` gains `test_malformed_content_length_gets_400_over_tls`,
since the drain re-times `self.connection`, which is an `SSLSocket` on that
listener.

## Execution Tests

```
ruff check src tests                -> All checks passed!
ruff format --check src tests        -> 122 files already formatted
mypy src                            -> Success: no issues found in 55 source files
python -m pytest tests/ -q          -> 634 passed, 2 skipped
                                       coverage 88.99% (gate 80%)
```

`http_server.py` coverage 88% -> 89%; every line added here is covered.

Local note, unrelated to this change: six `tests/test_server_tls.py` cases
that use `urllib` fail on my machine because a system HTTP proxy intercepts
`urlopen`. They fail identically on a clean `upstream/master` checkout, and
all pass with `no_proxy='*'`; the counts above are from that run.

## Docs

One bullet in the `docs/ATTACH_ANY_SYSTEM.md` hardening checklist, next to
the existing `--max-body-bytes`/413 bullet, plus the module docstring.

### DIFF
--- a/docs/ATTACH_ANY_SYSTEM.md
+++ b/docs/ATTACH_ANY_SYSTEM.md
@@ -342,6 +342,14 @@ them is purely operational.
   oversized payloads are dropped at the edge instead of consuming sidecar
   memory and CPU. The cap exists to bound per-request memory while still
   admitting batched `StepEvent` arrays.
+- **Expect 400 on unusable request framing.** A POST whose `Content-Length`
+  is non-numeric or negative gets `{"error": "invalid Content-Length"}` with
+  400 and the connection is closed, because the body's extent is unknown and
+  a reused connection would misread the leftover bytes as the next request
+  (issue #129). An absent `Content-Length` is read as an empty body instead,
+  so it fails on the JSON parse like any other empty payload. Health
+  checkers and fuzzers that emit bad headers therefore get a clean status
+  line rather than a dropped connection and a server-side traceback.
 - **Remember nothing sensitive is retained.** Events carry hashes, timings,
   counts, and booleans, never prompt or response content. `GET /risks`
   retains the most recent 1000 `FailureRisk` records (ids, scores, trigger
--- a/src/snagline/server/http_server.py
+++ b/src/snagline/server/http_server.py
@@ -18,7 +18,10 @@ counting finished episodes without waiting for cap eviction.
 
 POST bodies are capped at ``max_body_bytes`` (default 1 MB); larger payloads
 get 413. This keeps the sidecar safe to expose and able to absorb buffered
-telemetry from any host.
+telemetry from any host. A missing ``Content-Length`` counts as an empty
+body; a malformed one (non-numeric or negative) gets 400 and the connection
+is closed, because the body's extent is unknown and a reused connection
+would misread the leftover bytes as the next request (issue #129).
 
 Optionally protect the endpoints with a shared secret by passing
 ``auth_token=`` to ``make_server``/``serve``. When set, requests must carry the
@@ -59,6 +62,7 @@ one-way telemetry, and callers should not block on detection results.
 
 from __future__ import annotations
 
+import contextlib
 import json
 import logging
 import math
@@ -96,6 +100,17 @@ _MAX_OVERCAP_DRAIN_EXCESS = 65_536
 # Drain reads happen in fixed chunks, so peak memory stays at one chunk no
 # matter what Content-Length claims.
 _DRAIN_CHUNK_BYTES = 16_384
+# A malformed Content-Length leaves the body's extent unknown, so the drain
+# above cannot be used to clear the socket before the 400 goes out (issue
+# #129). Instead the connection is drained under a hard total deadline: long
+# enough that a sender whose body is still in flight sees the status line
+# rather than an RST, short enough that a stalled sender cannot hold the
+# handler thread. A client waiting on the response never closes first, so this
+# is paid in full on every malformed request -- measured on loopback, a purely
+# non-blocking drain lost the 400 to a reset on 31 of 200 requests while any
+# nonzero deadline delivered all 200, hence a small one rather than none.
+# Bytes are bounded by the body cap as usual.
+_BAD_FRAMING_DRAIN_SECONDS = 0.05
 
 
 def _escape_label(value: str) -> str:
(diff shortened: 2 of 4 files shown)


Cyrax321 (commented):
Thank you, @Adhi1-2, for this fix — the analysis is thorough and the regression tests are exactly right.

I verified the bug on the reported base: `int(self.headers.get("Content-Length") or 0)` raising `ValueError` for `abc`/`0x10` and `rfile.read(-5)` raising for negative values, both dropping the connection with a handler traceback instead of a clean `400`. Your remedy — a single `_content_length()` helper shared by `do_POST` and `_read_body`, returning `0` for absent headers and `None` for unusable values, with `400 {"error": "invalid Content-Length"}` and a bounded drain so the response actually reaches the client — matches the contract issue #129 asks for, including handling hex, float, and negative edge cases.

**Why I cannot merge this PR as-is:**

Issue #129 was fixed on `origin/master` while your PR was in flight, via #185 (`78734cf` `fix(server): return 400 on malformed Content-Length instead of dropping`, merged `d5ba99d`). #129 is now CLOSED. That merge touched the same lines (`do_POST` framing, `_read_body`, and the same helper position), so this branch now shows `CONFLICTING` against current master — not a reflection on quality, just timing.

Your PR and #185 converge on the same fix (helper + 400 + drain). I compared the two diffs and both correctly handle non-numeric and negative values and both add raw-socket regressions asserting the `400` status line. #185 is already on master with green CI across ubuntu/windows/ml, so merging this branch would re-apply the same behavior with a conflict that would need a manual rebase.

**What I appreciate and suggest:**

- The investigation and the "Two judgement calls" section in your PR body (framing before auth, bounded drain for the 400) are excellent — that reasoning will help future reviewers of the existing fix.
- If you would like to build on top of the shipped fix, you could rebase this branch onto current master (`git fetch origin && git rebase origin/master`) and, if you see any gap between your edge-case handling and #185's, open a small follow-up. I am happy to review it.
- Otherwise, I will close this PR as superseded by #185 with full credit — your work is solid, it just landed in the window where the same bug was fixed in parallel.

Closing as superseded, with sincere thanks for the contribution. Please keep contributing — the raw-socket test discipline you showed here is exactly what this repo values.

Cyrax321:
Closing as superseded by #185 — thank you again @Adhi1-2 for the thorough fix and raw-socket tests. The same 400 handling is now on master; please rebase and open a follow-up if you spot any edge case #185 missed. Appreciated!

PR #177, opened 2026-08-26
feat(server): bound sidecar connection reads so a stalled sender cannot pin a handler thread

### LINKED ISSUE #129: [server] Malformed Content-Length crashes the handler thread; clients get a dropped connection instead of 400

### LINKED ISSUE #130: [server] No read timeout on sidecar connections: a stalled sender can pin handler threads indefinitely

Adhi1-2:
Fixes #130

## Summary of Changes

`BaseHTTPRequestHandler` inherits `timeout = None`, so every `rfile.read(n)` in
the sidecar blocked until exactly `n` bytes arrived or the peer closed. A client
could declare `Content-Length: 500000`, write seven bytes and go quiet, holding
one thread of the `ThreadingHTTPServer` for as long as it liked -- slowloris
retention, repeatable from many connections at once. `--max-body-bytes` does not
help: the cap is on bytes that arrive, and here they never do.

The remedy is to set the handler class's `timeout`. `socketserver` already
applies it in `setup()` via `connection.settimeout()`, and `handle_one_request`
already converts the resulting `TimeoutError` into a log line plus
`close_connection = True`, so the connection is dropped and its thread returned
to the pool. The budget is **per read, not per request**, so a slow sender that
keeps making progress is still served no matter how long the whole transfer
takes.

Reproduction against this branch, on an ephemeral loopback port: declare 500 KB,
send 7 bytes, then go silent. `read_timeout=0` is the documented opt-out, so
that row is the pre-fix behavior.

```
read_timeout=0.5   -> socket closed by server after 0.50s; handler threads still alive: 0
read_timeout=0     -> client gave up after 3.00s (TimeoutError); handler threads still alive: 1
```

Resolution mirrors the existing `_resolve_metrics_format`: explicit argument,
then `$SNAGLINE_SERVE_READ_TIMEOUT_S`, then the built-in 30s default.

| Surface | Behavior |
| --- | --- |
| `serve()` / `make_server()` / `make_handler()` | new `read_timeout` keyword |
| `snagline serve --read-timeout N` | new flag; `0` waits indefinitely |
| `$SNAGLINE_SERVE_READ_TIMEOUT_S` | free, via the new `Config.serve_read_timeout_s` field |
| default | 30.0 |
| `<= 0` or `NaN` | disabled, i.e. exactly the pre-#130 behavior |
| unusable value (`"soon"`, a non-number) | warn and fall back to 30.0 |
| failing env lookup | fall back silently to 30.0 |

Nothing raises out of the resolver. Configuration must never take the sidecar
down (project.md §1.2), so every bad input path ends at the default.

## One judgement call worth reviewing

**The same budget also bounds the TLS handshake**, which is one line beyond the
issue's literal ask and is flagged in-code as such. `setup()` runs only *after*
`_TLSThreadingHTTPServer.finish_request` wraps the socket, so the handler's own
timeout cannot cover `wrap_socket`. Without this, a peer that opens a TCP
connection and never sends a ClientHello holds its worker thread indefinitely --
the identical retention #130 closes for body reads, reachable on any sidecar
started with `--certfile`. The timeout is applied around `wrap_socket` and then
cleared, so the wrapped connection is handed to the handler in the same state it
was before and `setup()` applies the per-read budget as usual. When the budget is
disabled the handshake is left untouched.

## Justification

The sidecar is documented as safe to expose to a proxy and to absorb buffered
telemetry from any host, and `docs/ATTACH_ANY_SYSTEM.md` already tells operators
to keep the body-size caps in sync as a hardening step. A single unauthenticated
TCP connection that costs the sender nothing and costs the sidecar a thread until
it decides to leave undercuts that. Thread exhaustion is worse than a dropped
request: it takes ingestion down for every well-behaved agent on the box, and
the fail-open design means those agents lose monitoring silently.

The fix is deliberately the smallest one that holds. No new dependency, no
rewritten I/O path, no threading of state through the handler -- one class
attribute the stdlib was already built to honor, plus the config/CLI wiring
needed to tune it, and the opt-out for anyone whose senders really are slower
than 30s per read.

## Kind of Contribution

Bug Fix

## Unit Tests

New `tests/server/test_read_timeout.py` (7 tests):

- `test_stalled_sender_is_dropped_and_its_thread_released` -- the core
  regression. Declares 500 KB, sends 7 bytes, asserts the server closes the
  socket (`recv()` returns `b""`, well inside the client's own 10s timeout) and
  that `threading.active_count()` returns to its baseline. The thread assertion
  polls rather than sleeping a fixed amount, since CI schedulers are not prompt.
- `test_a_slow_but_progressing_sender_is_still_served` -- 16-byte chunks with
  10ms gaps under a 0.3s budget. Total transfer exceeds the timeout several times
  over and still gets 202, pinning that the budget is per read.
- `test_well_formed_requests_are_unaffected_by_the_default_timeout` -- nothing
  passed, ordinary POST, 202.
- `test_handler_class_carries_the_resolved_timeout` -- default, explicit `2.5`,
  and the `0` / `-1` opt-out mapping to `None`.
- `test_resolve_read_timeout_precedence` -- argument beats env beats default;
  `"0"` disables; `"soon"` falls back.
- `test_resolve_read_timeout_survives_unusable_arguments` -- `NaN` disables, a
  string falls back.
- `test_resolve_read_timeout_survives_a_broken_environment` -- a raising
  `Config.from_env_overrides` falls back instead of propagating.

`tests/test_server_tls.py` (2 added):

- `test_silent_peer_does_not_hold_a_tls_worker_thread` -- opens a socket to the
  TLS listener, never sends a ClientHello, asserts the server gives up and the
  worker thread is released.
- `test_tls_handshake_timeout_opt_out_still_serves` -- with `read_timeout=0` the
  handshake budget is `None` and HTTPS still works.

`tests/test_cli.py` (1 added, 1 updated):

- `test_main_serve_passes_read_timeout` -- `--read-timeout 5` reaches `serve()`
  as `5.0`, and `--read-timeout 0` as `0.0` rather than being dropped as falsy.
- `test_main_serve_starts_server` asserts the exact kwargs dict forwarded to
  `serve()`, so it gained `"read_timeout": None`. The CLI deliberately passes
  `None` instead of duplicating the 30s default, so the default lives in exactly
  one place; the updated assertion has a comment saying so.

Every line and branch added by this PR is covered, with one exception: the single
`read_timeout` forwarding argument inside `serve()`'s body. `serve()` runs
`serve_forever()`, so every test monkeypatches it and its whole body
(`http_server.py:845-869`) is already absent from coverage on master. The
argument it forwards is asserted from the CLI side by
`test_main_serve_passes_read_timeout`.

## Execution Tests

Full suite, ruff, format and mypy as CI runs them:

```
632 passed, 2 skipped in 49.90s
Required test coverage of 80% reached. Total coverage: 88.96%

All checks passed!                             (ruff check src tests)
122 files already formatted                    (ruff format --check src tests)
Success: no issues found in 55 source files    (mypy src)
```

Plus the end-to-end reproduction quoted above, run against a real
`ThreadingHTTPServer` on an ephemeral port with both a live budget and the
opt-out, checking the observable symptom from the issue (a retained handler
thread) rather than only the timeout value.

Local note: `tests/test_server_tls.py` fails on this machine with
`HTTP Error 500: Internal Privoxy Error` unless `no_proxy='*'` is set -- a system
HTTP proxy intercepting `urlopen`. I confirmed with `git stash
--include-untracked` that those same tests fail identically on a clean
`upstream/master`, so it is environmental and not introduced here. All runs
reported above use `no_proxy='*'`.

## Docs

`docs/ATTACH_ANY_SYSTEM.md` gains a "Leave the read timeout on" bullet in the
hardening checklist, immediately after the `--max-body-bytes` bullet as the issue
asks. It states the default, the env var, that the budget is per read, why the
size cap does not cover this case, the `0` opt-out, and that the same budget
bounds the TLS handshake.

## Relationship to #176

This is the second of the two server-hardening issues, kept separate per the
note in #130 that it was filed on its own "so the drain fix stays revertible on
its own merits". #176 (Fixes #129) touches `do_POST` framing; this PR touches the
handler `timeout`, `Config`, and the CLI. They do not overlap and either can land
first -- #129's bounded drain sets and restores the socket timeout inside a single
call, so it neither depends on nor conflicts with this default.

### DIFF
--- a/docs/ATTACH_ANY_SYSTEM.md
+++ b/docs/ATTACH_ANY_SYSTEM.md
@@ -342,6 +342,17 @@ them is purely operational.
   oversized payloads are dropped at the edge instead of consuming sidecar
   memory and CPU. The cap exists to bound per-request memory while still
   admitting batched `StepEvent` arrays.
+- **Leave the read timeout on.** A sender that promises a large
+  `Content-Length`, writes a few bytes and goes quiet would otherwise hold one
+  handler thread for as long as it likes -- slowloris-style retention, and the
+  body-size cap does not help because the bytes never arrive.
+  `--read-timeout` (default 30, or `$SNAGLINE_SERVE_READ_TIMEOUT_S`) bounds
+  how long any single read may wait before the connection is dropped and its
+  thread released. The budget is per read, not per request, so a slow client
+  that keeps making progress is still served. Raise it if you genuinely have
+  senders slower than that; `0` restores the old unbounded behavior. With TLS
+  terminated in the sidecar the same budget also bounds the handshake, which
+  a silent peer could otherwise stall indefinitely.
 - **Remember nothing sensitive is retained.** Events carry hashes, timings,
   counts, and booleans, never prompt or response content. `GET /risks`
   retains the most recent 1000 `FailureRisk` records (ids, scores, trigger
--- a/src/snagline/cli.py
+++ b/src/snagline/cli.py
@@ -305,6 +305,15 @@ def _build_parser() -> argparse.ArgumentParser:
         "[default: 1000].",
     )
     p_serve.add_argument(
+        "--read-timeout",
+        type=float,
+        default=None,
+        help="Seconds any single read on a connection may wait before the "
+        "sidecar drops it, so a stalled sender cannot pin a handler thread "
+        "(issue #130). Falls back to $SNAGLINE_SERVE_READ_TIMEOUT_S, then 30. "
+        "Use 0 to wait indefinitely.",
+    )
+    p_serve.add_argument(
         "--sink",
         choices=["console", "webhook", "slack", "pagerduty"],
         default="console",
@@ -849,6 +858,7 @@ def _cmd_serve(args: argparse.Namespace) -> int:
             auth_token=auth_token,
             max_body_bytes=args.max_body_bytes,
             max_risks=args.max_risks,
+            read_timeout=args.read_timeout,
             **tls_kwargs,
         )
     return 0
--- a/src/snagline/config.py
+++ b/src/snagline/config.py
@@ -334,6 +334,19 @@ class Config:
     # ?format=prometheus; environment override SNAGLINE_METRICS_FORMAT.
     metrics_format: str = "prometheus"
 
+    # --- Sidecar socket read timeout (issue #130) ----------------------------
+    # How long one sidecar connection may go without producing the bytes it
+    # promised before the server drops it. Without a timeout every
+    # ``rfile.read(n)`` blocks until exactly n bytes arrive, so a client that
+    # declares ``Content-Length: 500000`` and then sends seven bytes holds a
+    # handler thread of the ThreadingHTTPServer for as long as it likes --
+    # slowloris-style resource retention against the sidecar. Applies per read,
+    # not per request, so a genuinely slow sender that keeps making progress is
+    # never cut off; the ceiling is deliberately generous for that reason.
+    # Values <= 0 restore the pre-#130 unbounded behavior for anyone who needs
+    # it. Environment override SNAGLINE_SERVE_READ_TIMEOUT_S.
+    serve_read_timeout_s: float = 30.0
+
     # --- Enforcement policy (issue #93) ---------------------------------------
     # Optional escalation layer that runs AFTER the sinks on every dispatched
     # risk (documented ordering: detectors -> sinks -> policy). "observe" (the
--- a/src/snagline/server/http_server.py
+++ b/src/snagline/server/http_server.py
@@ -20,6 +20,14 @@ POST bodies are capped at ``max_body_bytes`` (default 1 MB); larger payloads
 get 413. This keeps the sidecar safe to expose and able to absorb buffered
 telemetry from any host.
 
+Every accepted connection also carries a read timeout (issue #130):
+``read_timeout=`` seconds, or ``SNAGLINE_SERVE_READ_TIMEOUT_S``, defaulting to
+30. Without it a client could declare ``Content-Length: 500000``, send seven
+bytes and go quiet, holding one handler thread of the ``ThreadingHTTPServer``
+indefinitely. The budget applies per read rather than per request, so a slow
+sender that keeps making progress is never cut off; a sender that stops is
+dropped and its thread released. Pass a value at or below zero to disable.
+
 Optionally protect the endpoints with a shared secret by passing
 ``auth_token=`` to ``make_server``/``serve``. When set, requests must carry the
 token via ``Authorization: Bearer <token>`` or ``X-Snagline-Token: <token>``;
@@ -59,6 +67,7 @@ one-way telemetry, and callers should not block on detection results.
 
 from __future__ import annotations
 
+import contextlib
 import json
 import logging
 import math
@@ -68,7 +77,7 @@ import threading
 import time
 from collections import OrderedDict, deque
 from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
-from typing import Any
+from typing import Any, ClassVar
 from urllib.parse import parse_qs, urlsplit
 
 from snagline.adapters.claude_code import HookTracker, ingest_payload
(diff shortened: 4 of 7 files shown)


Cyrax321 (commented):
Thank you, @Adhi1-2, for tackling #130 — the slowloris retention analysis is spot-on.

I reproduced the stalled-sender pin on the old base as you described: `Content-Length: 500000` with 7 bytes sent holds one `ThreadingHTTPServer` handler thread indefinitely because `BaseHTTPRequestHandler.timeout = None` makes every `rfile.read(n)` block until `n` bytes arrive. Your fix — setting the handler's `timeout` class attribute so `connection.settimeout()` is applied in `setup()`, with `TimeoutError` converted by `handle_one_request` into a log line + `close_connection = True`, plus the `read_timeout` plumbing (`serve()`/`make_server()`/`make_handler()` kwarg, `--read-timeout` flag, `Config.serve_read_timeout_s` and `$SNAGLINE_SERVE_READ_TIMEOUT_S`, default 30s, `<=0` or `NaN` disabled, unusable env warns to 30s, per-read not per-request semantics) — matches the proposal in #130 and is correctly documented next to `--max-body-bytes`.

**Why I cannot merge this PR as-is:**

Issue #130 was fixed on `origin/master` while your PR was in flight, via #186 (`e8e3630` `fix(server): add read timeout to release stalled handler threads`, merged `8ac01e6`). #130 is now CLOSED. That merge introduced the same `timeout` mechanism on the same handler class, so this branch now shows `CONFLICTING` — again, timing, not quality.

Both fixes converge on `timeout = 30.0` with env/config plumbing and a raw-socket test asserting the stalled sender's thread is released. #186 is already on master with green CI, so merging this branch would duplicate the behavior.

**What I appreciate and next steps:**

- Your reproduction script (7 bytes / 500 KB, `read_timeout=0.5` vs `0` comparison with handler-thread counts) and the resolution table are excellent documentation for the fix that shipped.
- If you spot any delta between your plumbing (especially the `NaN` / warn-fallback handling) and #186's, please rebase onto current master and open a follow-up — happy to review.
- Otherwise I will close this as superseded by #186 with full credit.

Sincere thanks for the contribution — the per-read vs per-request correctness argument you made is exactly the kind of detail this project relies on, and I hope you keep contributing.

Cyrax321:
Closing as superseded by #186 — thank you again @Adhi1-2 for the slowloris analysis and per-read timeout design. The same handler-timeout fix is now on master; happy to review a follow-up on top of it if you see a gap. Appreciated!

Issue #93, opened 2026-08-24
[sink] EnforcementPolicy hook: callback and halt_webhook escalation (fail-open preserved)


Cyrax321:
## Summary

Add an optional enforcement layer to `Monitor`: `policy="observe" (default) | "callback" | "halt_webhook"`. Today FailureRisks can only be logged or POSTed opportunistically; there is no supported way to *act* on a risk synchronously (pause the host agent, trip a circuit breaker). Detection-only is the right default and stays fail-open; but production users (and our own roadmap items: budget breach, meltdown, duplicate side effects) need a sanctioned escalation path that is still safe by construction.

## Why

- The zero-dep competitor nakata-agentguard ships halt semantics (`halt_on_loop`, webhook-on-halt daemon) and its README leads with it: people want teeth. OpenAI's long-horizon safety writeup describes trajectory-level monitoring whose defining feature is the ability to pause sessions on anomaly. Observers get adopted; monitors-with-teeth get relied on.
- Our own pending detectors create actions that demand enforcement: `budget_breach` and `side_effect_duplicate` are meaningless if the best outcome is a log line.
- Fail-open must survive this feature. The design below guarantees enforcement can never corrupt or block the monitored agent beyond its explicit timeout envelope, preserving project principle #2.

## Design

```python
Monitor(..., policy="observe")                       # current behavior
Monitor(..., policy="callback", on_risk=my_fn)       # sync callback, wrapped fail-open
Monitor(..., policy="halt_webhook", halt_url=..., halt_timeout_s=0.25)
```

- `callback`: `on_risk(risk)` invoked after sinks; exceptions swallowed exactly like sink errors (logged, counted). For in-process circuit breakers (raise-your-own exception in your agent loop based on a flag the callback sets).
- `halt_webhook`: fire-and-collect POST of the FailureRisk to `halt_url`; response `{action: "continue"|"pause", reason}` consumed as `monitor.last_directive`. We deliberately do NOT raise into the host loop ourselves; the host decides what pause means. Timeout defaults to 250ms; on timeout/error directive defaults to continue (fail-open).
- `min_severity_for_halt` filter so only score>=0.8 risks pay the webhook cost.
- Interaction with #8 (sink-under-lock blocking): halt webhook runs outside the ingest lock; documented ordering: detectors -> sinks -> policy.
- CLI: `snagline serve --halt-forward URL` so sidecar consumers get the same semantics.

## Acceptance criteria

[...]

PR #178, opened 2026-08-26
feat(server): expose GET /directive and the policy_errors metrics family

### LINKED ISSUE #169: [server] Enforcement follow-up from #93: expose last_directive over HTTP (GET /directive) + policy_errors family in GET /metrics
## Summary

#93 (merged in #166) shipped the enforcement layer but, by deliberate zone ownership, did not touch `server/http_server.py`. That leaves two small gaps for sidecar consumers of `snagline serve --halt-forward URL`:

1. **The pause directive is not readable over HTTP.** The halt webhook lands its response on the in-process `Monitor.last_directive` (thread-safe property added in #166), but no endpoint exposes it. An external harness driving the agent through the sidecar has no way to learn that the halt endpoint said `"pause"`, which is the entire point of the feature for non-Python consumers.
2. **`policy_errors` is missing from Prometheus exposition.** `MonitorMetrics` gained a fifth counter (`policy_errors`: swallowed callback raises + fail-open halt-webhook fallbacks), but GET /metrics renders only the original four families (`events_ingested`, `risks_emitted`, `detector_errors`, `sink_errors`). Enforcement faults are invisible to scrapes even though they are exactly what an operator wants to see when a halt endpoint starts dying.

### 1. `GET /directive`
- Auth-gated like every endpoint except `/health` (same `_authorized()` path as `/risks`; a directive can pause the host, so it must never be world-readable).
- Body: JSON rendering of `HaltDirective`: `{"action": "continue"|"pause", "reason": str, "timestamp": float}`; 200 always (the resource exists as soon as the Monitor does, defaulting to continue).
- Implementation note: `make_handler` already holds `monitor`; read via the public thread-safe `monitor.last_directive` property, never the private attribute.
- Fail-open constraint: the endpoint must only serialize state that already exists; it must not add locks around serving or ever raise into the handler loop.

### 2. Prometheus family
- Add a fifth entry to the `monitor_families` tuple in `SidecarMetricsCollector.render_prometheus`:
  `("policy_errors", "snagline_monitor_policy_errors_total", "Enforcement faults swallowed fail-open (#93): callback raises and halt-webhook timeout/error fallbacks.")`
- Naming follows the established `snagline_monitor_*_total` convention from #98.

## Acceptance criteria

- [ ] `GET /directive` returns the live directive; without `--auth-token` set it is open exactly like `/risks`, with a token it requires Bearer / X-Snagline-Token (mirror tests/test_http_server.py auth cases).
- [ ] `GET /directive` reflects a pause after a halt-webhook round trip (testable with the local-endpoint pattern from tests/test_enforcement_policy.py).
- [ ] GET /metrics includes `snagline_monitor_policy_errors_total` in both prometheus and classic bodies (classic already serves `Monitor.metrics()` verbatim, so only the text exposition needs work).
- [ ] Full gate stays green: pytest, ruff check src tests, ruff format --check src tests, mypy src.
- [ ] README sidecar endpoint list gains `GET /directive`.

## References

- Feature: #93, implementation PR #166 (merged e7cd8f3)
- Exposition format being extended: #98
- Directive contract: `Monitor.last_directive` / `HaltDirective` in `src/snagline/monitor.py`

## Duplicate check

Searched open+closed issues for `directive`, `last_directive`, `GET /directive`, `policy_errors`, `prometheus`, `metrics`, `halt-forward`: closest hit is #98 (the original exposition, closed, predates this counter). No existing open or closed issue covers either gap.

Adhi1-2:
Fixes #169

## Summary of Changes

#93 shipped the enforcement layer without touching `server/http_server.py`, by
deliberate zone ownership. This closes the two gaps that left for sidecar
consumers of `snagline serve --halt-forward URL`.

**1. `GET /directive`.** The halt webhook's answer landed on the in-process
`Monitor.last_directive` and nothing published it, so a non-Python host -- the
entire audience for server mode -- could not learn that the endpoint said
`"pause"`. That is the point of the feature for those consumers, and it was
unreachable. The endpoint renders the directive as
`{"action": "continue"|"pause", "reason": str, "timestamp": float}`.

**2. `snagline_monitor_policy_errors_total`.** `MonitorMetrics` gained a fifth
counter in #166 (swallowed callback raises plus halt-webhook timeout/error
fallbacks), but the exposition rendered only the original four families. A halt
endpoint starting to die was invisible to scrapes, which is precisely when an
operator needs to see it. The JSON body had it all along, so only the text
exposition needed the fifth `monitor_families` entry.

Before and after, same script (`Monitor([], [])`, ephemeral loopback port):

| | master (`f7857d1`) | this branch |
| --- | --- | --- |
| `GET /directive` | `404 {"error": "not found"}` | `200 {"action": "continue", "reason": "", "timestamp": 0.0}` |
| `policy_errors` in `/metrics` | absent | present |
| `policy_errors` in `?format=classic` | present | present (unchanged) |

## Contract details

- **Auth-gated like `GET /risks`.** A directive can pause the host, so it goes
  through the same `_authorized()` path: 401 without a token, `Bearer` and
  `X-Snagline-Token` both accepted, and `GET /health` stays open so liveness
  probes are unaffected by the new route.
- **Always 200, never 404.** The resource exists as soon as the Monitor does and
  starts at continue, so a polling host does not have to distinguish "no
  directive yet" from "wrong URL".
- **Read-only and fail-open.** It reads the public thread-safe
  `monitor.last_directive` property, never the private attribute, and adds no
  locking of its own. If that read fails at all -- a monitor predating
  enforcement, a duck-typed object -- it reports the default continue and logs at
  debug rather than raising into the handler loop. A 500 would tell a polling
  host nothing it could act on, and it is the same fail-open the webhook path
  already takes on a timeout or a dead endpoint.
- **Payload coerced, not trusted.** `make_handler` holds whatever object the
  caller passed as `monitor`, so the three fields go through `str`/`str`/`float`;
  a body that cannot be serialized would turn a read-only endpoint into a 500.
- **Older snapshots still render.** `monitor_metrics.get(key, 0)` means a
  `Monitor.metrics()` dict without `policy_errors` produces a valid body with a
  zero sample instead of a `KeyError`.

## Justification

`snagline serve --halt-forward URL` exists so that hosts which are not Python
libraries can be paused by the same enforcement policy the in-process API
offers. Landing the directive only on a Python attribute means the sidecar can
decide to pause an agent it has no way of telling. That is not a partial
feature, it is a control path that terminates in a dead end -- and the
workaround, reaching into the Monitor object, is unavailable to exactly the
consumers server mode is for.

The metrics half is smaller but of the same kind: enforcement is the one
subsystem allowed to affect the host's behavior, so its fault counter is the one
an operator least wants to be missing from a dashboard. Both changes are
additive; nothing existing is renamed or reshaped.

## Kind of Contribution

New Feature

## Unit Tests

New `tests/server/test_directive_endpoint.py` (5 tests):

- `test_directive_defaults_to_continue` -- a fresh Monitor answers 200 with the
  full default directive, pinning that the resource exists before any webhook
  round trip.
- `test_directive_reports_a_pause_after_a_halt_round_trip` -- the regression this
  endpoint exists for. A local halt endpoint answers `pause`; a risk is driven
  through the sidecar's own `POST /events`; `GET /directive` reports continue
  before and `pause` with the endpoint's reason after, with a set timestamp.
- `test_directive_is_behind_the_token_when_one_is_set` -- both sides: 401 with no
  token and with a wrong one, 200 with `Bearer` and with `X-Snagline-Token`, and
  `/health` still open.
- `test_directive_fails_open_to_continue_instead_of_500` -- a monitor whose
  `last_directive` raises still gets a 200 continue.
- `test_unknown_get_paths_still_404` -- `/directives` and `/directive/latest`
  remain 404, so adding a route did not turn the fallthrough into a prefix match.

`tests/server/test_metrics_prometheus.py` (1 test, 3 assertions extended):

- `test_policy_errors_family_grows_on_an_enforcement_fault` -- a raising callback
  under `policy="callback"`, two ingested steps, and the counter reads exactly
  2.0 in the exposition body and 2 in `?format=classic`, so the two bodies are
  pinned to agree.
- The existing family-presence, zero-value, and classic-key assertions gained
  `policy_errors`, which is where a future rename would be caught.
- `test_collector_counts_risks_per_label_pair` already passes a snapshot without
  the new key, so it now also covers the "older Monitor" path for free.

Every added line and branch in `http_server.py` is covered: none of the new
ranges appear in coverage's miss list, which is otherwise identical to master's.

## Execution Tests

End to end through the real CLI, not the test harness: a halt endpoint answering
`pause`, `python -m snagline.cli serve --port N --auth-token s3cret
--halt-forward URL --min-severity-for-halt 0.5` as a subprocess, three identical
signatures to pass the loop detector's threshold of 3, then the directive read
back over HTTP the way a non-Python host would.

```
no token       -> (401, '{"error": "unauthorized"}')
before         -> (200, '{"action": "continue", "reason": "", "timestamp": 0.0}')
after          -> (200, '{"action": "pause", "reason": "loop detected", "timestamp": 1787759218.4418151}')
metrics        -> snagline_monitor_policy_errors_total 0
```

The `policy_errors 0` line is the point of that last check: the family is present
and correctly zero on a run where the halt endpoint answered every request.

Full gate:

```
628 passed, 2 skipped in 49.64s
Required test coverage of 80% reached. Total coverage: 88.95%

All checks passed!                             (ruff check src tests)
122 files already formatted                    (ruff format --check src tests)
Success: no issues found in 55 source files    (mypy src)
```

Local note: `tests/test_server_tls.py` needs `no_proxy='*'` on this machine or a
system HTTP proxy intercepts `urlopen` and returns 500. Those failures reproduce
on a clean `upstream/master`, so they are environmental; all runs above use
`no_proxy='*'`.

## Docs

- README sidecar endpoint list gains `GET /directive`, and the enforcement
  paragraph now says non-Python hosts read the answer back there, with the body
  shape and the auth note.
- `http_server.py`'s module docstring gains the route in its endpoint table plus
  a paragraph on why it exists and why it is auth-gated.
- The `snagline serve` halt-forwarding banner now points at `GET /directive`
  instead of only naming `Monitor.last_directive`, so an operator reading startup
  output learns the endpoint exists.

No change to `docs/ATTACH_ANY_SYSTEM.md`: its reverse-proxy configs use
`location /` rather than a path allowlist, so nothing there needed updating for a
new route.

### DIFF
--- a/README.md
+++ b/README.md
@@ -512,7 +512,7 @@ if monitor.last_directive.action == "pause":
 
 Fail-open survives enforcement by construction: a callback exception is logged, counted under `metrics()["policy_errors"]`, and swallowed (unless `fail_open=False`); a webhook timeout, dead endpoint, malformed body, unknown action, or HTTP error leaves the directive at continue. Worst-case added latency per halting step is bounded by the explicit `halt_timeout_s` envelope (default 250ms) and is paid only by the dispatching thread -- other episodes keep ingesting. Measured numbers (Apple M1, CPython 3.14, 2026-08-26; run `python benchmarks/enforcement_benchmark.py`): observe baseline ~2 us/step, responding localhost endpoint ~264 us/step median, refused endpoint ~57 us/step, stalled endpoint ~252.7 ms/step, i.e. the full timeout budget plus a small epsilon.
 
-The equivalent for sidecar consumers: `snagline serve --halt-forward URL` runs the identical policy inside the HTTP sidecar (`--halt-timeout` and `--min-severity-for-halt` tune it). Configuration is 12-factor too: `SNAGLINE_POLICY`, `SNAGLINE_HALT_URL`, `SNAGLINE_HALT_TIMEOUT_S`, `SNAGLINE_MIN_SEVERITY_FOR_HALT`. Note `on_risk` is code-only: callables cannot arrive via env vars or config files.
+The equivalent for sidecar consumers: `snagline serve --halt-forward URL` runs the identical policy inside the HTTP sidecar (`--halt-timeout` and `--min-severity-for-halt` tune it). Non-Python hosts read the answer back with `GET /directive`, which returns `{"action": "continue"|"pause", "reason": ..., "timestamp": ...}` and is auth-gated like `GET /risks`. Configuration is 12-factor too: `SNAGLINE_POLICY`, `SNAGLINE_HALT_URL`, `SNAGLINE_HALT_TIMEOUT_S`, `SNAGLINE_MIN_SEVERITY_FOR_HALT`. Note `on_risk` is code-only: callables cannot arrive via env vars or config files.
 
 ## External Agent Bridges
 
@@ -527,6 +527,7 @@ snagline serve --host 127.0.0.1 --port 8787
 # POST /events              body: StepEvent JSON   -> 202, ingested
 # POST /hooks/claude-code   body: native hook payload -> 202, mapped + ingested
 # GET  /health                                     -> 200
+# GET  /directive                                  -> 200, latest halt directive
 ```
 
 ### Claude Code integration
--- a/src/snagline/cli.py
+++ b/src/snagline/cli.py
@@ -825,7 +825,8 @@ def _cmd_serve(args: argparse.Namespace) -> int:
             f"snagline serve: halt forwarding enabled -> {cfg.halt_url} "
             f"(timeout {cfg.halt_timeout_s}s, min severity "
             f"{cfg.min_severity_for_halt}); directives land on "
-            "Monitor.last_directive (issue #93)",
+            "Monitor.last_directive and are readable at GET /directive "
+            "(issues #93, #169)",
             file=sys.stderr,
         )
     if not auth_token:
--- a/src/snagline/server/http_server.py
+++ b/src/snagline/server/http_server.py
@@ -9,6 +9,7 @@ shell script, a Claude Code hook) can POST ``StepEvent`` JSON to:
     POST /episodes/end       body: {"episode_id": "<id>"}            -> monitor.end_episode()
     GET  /health                                                     -> 200 OK
     GET  /metrics                                                    -> Prometheus text exposition (v0.0.4)
+    GET  /directive                                                  -> the latest halt-webhook directive
 
 A host that knows an episode is over can signal it with ``POST
 /episodes/end`` carrying ``{"episode_id": "..."}`` (issue #123): the sidecar
@@ -47,6 +48,15 @@ clients sending ``Accept: application/json``, or when the sidecar is started
 with ``SNAGLINE_METRICS_FORMAT=classic`` (config key ``metrics_format``);
 ``?format=prometheus`` forces the new format back on per request.
 
+``GET /directive`` reports the newest halt-webhook directive as
+``{"action": "continue"|"pause", "reason": ..., "timestamp": ...}`` (issue
+#169). Without it, a sidecar started with ``--halt-forward URL`` had nowhere
+to publish the answer: the directive landed on the in-process
+``Monitor.last_directive`` and non-Python hosts, the whole point of server
+mode, could not read it. The resource exists as soon as the Monitor does and
+defaults to continue, so the endpoint always answers 200. It is auth-gated
+like ``GET /risks``, because a directive can pause the host.
+
 No framework, no third-party dependency -- this keeps the zero-dependency
 principle intact even for server mode. For high-throughput production use,
 front it with a real ASGI server / reverse proxy; that is the user's infra
@@ -74,7 +84,7 @@ from urllib.parse import parse_qs, urlsplit
 from snagline.adapters.claude_code import HookTracker, ingest_payload
 from snagline.config import Config
 from snagline.events import StepEvent
-from snagline.monitor import Monitor
+from snagline.monitor import HaltDirective, Monitor
 
 logger = logging.getLogger("snagline")
 
@@ -161,6 +171,20 @@ def _choose_metrics_format(
     return configured
 
 
+def _directive_payload(directive: Any) -> dict[str, Any]:
+    """Render a :class:`HaltDirective` as the ``GET /directive`` body.
+
+    Coerced rather than trusted: the handler holds whatever object the caller
+    passed as ``monitor``, and a body that cannot be serialized would turn a
+    read-only endpoint into a 500.
+    """
+    return {
+        "action": str(directive.action),
+        "reason": str(directive.reason),
+        "timestamp": float(directive.timestamp),
+    }
+
+
 class SidecarMetricsCollector:
     """Process-lifetime sidecar counters, renderable as Prometheus text.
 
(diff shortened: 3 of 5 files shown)


Cyrax321 (approved):
Thank you, @Adhi1-2 — this closes the exact gap #93 left for sidecar consumers, and you described the motivation precisely.

I reviewed this against `origin/master` at `8ac01e6` (which already includes your docs fix #179) and against the enforcement design in #93/#166. Detailed notes:

**Correctness and design:**

- **`GET /directive` route:** Auth-gated like `GET /risks` via the existing `_authorized()` path, always 200 with `{"action": "continue"|"pause", "reason": str, "timestamp": float}`, read-only, fail-open to `continue` on any exception. All of this matches the contract #169 asks for. The choice to default to `continue` rather than 404 is right — a polling host should not have to distinguish "no directive yet" from "wrong URL".

- **`_directive_payload` coercion:** `str()` / `float()` wrapping before serialization is exactly what a read-only endpoint should do. Without it a poisoned `last_directive` object could turn a GET into a 500. The `try/except` falling back to `HaltDirective()` mirrors the webhook path's own fail-open timeout behavior.

- **`HaltDirective` import:** Added `from snagline.monitor import HaltDirective, Monitor` — correctly imports the directive type for the fallback path. No circular-import risk (`http_server` already imports `Monitor`).

- **`policy_errors` metrics family:** Adding the fifth entry to `monitor_families` with `snagline_monitor_policy_errors_total` follows the `snagline_monitor_*_total` naming from #98 and #166. Rendering a missing key as `0` (`A key the snapshot does not carry renders as 0, so an older Monitor still produces a valid body`) keeps the exposition valid for older Monitors — nice backward-compatibility detail.

- **CLI banner:** Updating the `snagline serve --halt-forward` banner to point at `GET /directive` is a small but important discoverability fix.

**Tests — exemplary:**

- `test_directive_defaults_to_continue` — pins "200, never 404".
- `test_directive_reports_a_pause_after_a_halt_round_trip` — the regression this endpoint exists for, with a real local halt endpoint answering `pause`. Verifies both the pre-ingest `continue` and post-ingest `pause` with `timestamp > 0`.
- `test_directive_is_behind_the_token_when_one_is_set` — 401 without token, both `Bearer` and `X-Snagline-Token` accepted, and `/health` stays open. Exactly the security contract needed.
- `test_directive_fails_open_to_continue_instead_of_500` — using `_UnreadableDirectiveMonitor` to simulate a pre-#93 monitor, expects 200 `continue` not 500. Matches the fail-open principle.
- `test_unknown_get_paths_still_404` — ensures the new route did not widen the fallthrough.
- `test_policy_errors_family_grows_on_an_enforcement_fault` — raising callback increments both the Prometheus exposition and the classic JSON `policy_errors` counter to `2.0`. Proves the fifth family is wired end-to-end.

All five server tests are isolated (ephemeral loopback ports, `ThreadingHTTPServer`, proper shutdown), and the metrics test asserts both `render_prometheus` and `?format=classic` agree.

**Checks:** Before the README conflict from #179, all CI legs were green (ubuntu 3.10-3.13, windows, ml extra, accuracy gate, Gitar approved). The branch now shows `UNKNOWN` only because the base moved when #179 merged — README was touched by both PRs at adjacent lines. This is a trivial textual conflict (your change adds the `GET /directive` line to the enforcement paragraph and the endpoint list; #179 changed the overhead/env-var rows far above). I will resolve it by merging `origin/master` into the branch and re-verifying the gate before merging.

**One tiny note already applied:** No action needed — the conflict resolution is just keeping both README hunks.

Approving with sincere thanks. Exposing `last_directive` over HTTP is what makes `--halt-forward` usable for non-Python hosts — without it the feature was invisible to the very audience server mode serves. Really appreciated.

PR #179, opened 2026-08-26
docs: correct README claims contradicted by the code and measurements

Adhi1-2:
## Why

I went through the README claim by claim against the code and re-ran every measurement it cites, at `f7857d1`. Most of it holds up — the detection-accuracy table reproduces **exactly**, and the Quick Start's example output is byte-identical to what the snippet actually prints. Six claims did not hold, and four of them are the README contradicting *itself*.

That matters more here than in most projects: this README's credibility rests on lines like "Overhead is measured, not asserted" and "No rounding up, no cherry-picking." A reader who spot-checks the headline number and finds it off by 2.5x has no way to know the accuracy table is solid. This PR is only about closing that gap.

## What changed

| # | Claim | Reality | Evidence |
|--:|:--|:--|:--|
| 1 | "**Sub-microsecond** overhead", "approximately 1 microsecond per `ingest()`", Features cell "1.9 / 33.9 us" | The README's *own* verification block says median **2.43 us/step**. Three different numbers, one of them an order-of-magnitude-adjacent overclaim. | `benchmarks/overhead_benchmark.py` |
| 2 | "531 passed, 2 skipped" *and* "385 tests passing, 1 skipped" | Both stale and mutually contradictory. Actual: **622 passed, 2 skipped**, 88.88% coverage. | `python -m pytest tests/ -q` |
| 3 | "`pip install snagline-agent` works with nothing but Python 3.10+" | `https://pypi.org/pypi/snagline-agent/json` → **404**. Status says "Not on PyPI" three sections later. | `curl` |
| 4 | Config snippet: `cusum_min_samples=20` | Real default is **5** — and the env table 50 lines below already says 5. | `src/snagline/config.py:175` |
| 5 | "Every scalar field of `Config` doubles as a 12-factor environment variable" | **11 scalar fields absent** from the table: `semantic_drift_*` (6), `side_effect_*` (3), `compaction_tripwire_*` (2). Silently undiscoverable knobs. | `dataclasses.fields(Config)` vs README |
| 6 | Contributing links three guides, gives no test command | `addopts = "--cov=snagline"` + no `dev` extra ⇒ a bare `pytest` after `pip install .` dies on unrecognised `--cov`. New contributors must reverse-engineer `ci.yml`. | `pyproject.toml:90`, `ci.yml:54` |

For #1 I did **not** replace the maintainer's Apple M1 / CPython 3.14.5 figures — I kept them and added an independent reproduction alongside (Apple M4 / CPython 3.13.5: median **1.70 us/step**, p99 **1.77 us/step**), then pointed the Features table at the same numbers instead of a superseded pair. The honest headline is "microsecond-scale", which is still an excellent number and needs no inflation.

For #2 I recorded the interpreter I actually used. Worth flagging separately: the old note claimed **Python 3.14**, but `ci.yml` tests **3.10–3.13** and `pyproject.toml` classifiers stop at 3.13. I documented the CI matrix rather than assert anything about 3.14 I can't reproduce.

## Verification

Everything newly written into the README was run first, at this commit:

```
python -m pytest tests/ -q      → 622 passed, 2 skipped, coverage 88.88% (gate 80%)
ruff check src tests            → All checks passed!
ruff format --check src tests   → 121 files already formatted
mypy src                        → Success: no issues found in 55 source files
benchmarks/overhead_benchmark.py→ median 1.70 us/step, p99 1.77 us/step (Apple M4, CPython 3.13.5)
benchmarks/detection_accuracy.py→ macro-F1 1.000, 76 episodes, 0 healthy controls fired, exit 0
```

## Deliberately left alone

- **The detection-accuracy table.** It reproduces exactly — every trigger 4 TP / 0 FP / 0 FN, macro-F1 1.000, 0 healthy-control firings. No change warranted.
- **The Quick Start output.** I checked it byte for byte, expecting `"in last 12 steps"` from `loop_window_size=12`; the documented `"action repeated 3x in last 3 steps"` is correct, because the window is bounded by steps *seen*. Nice detail.
- **The M1 / CPython 3.14 benchmark figures.** Not mine to overwrite; I can't reproduce that environment, so I added to the record rather than replacing it.

## Note for the maintainer

Docs-only, no behavior change. Two follow-ups I did *not* do here, happy to file as issues or fold in if you'd prefer:

1. **Add a `dev` extra** to `pyproject.toml` (`pytest`, `pytest-cov`, `ruff`, `mypy`). This PR documents the CI workaround; a `dev` extra would make the workaround unnecessary and let the README just say `pip install -e ".[dev]"`.
2. **Add a docs-drift guard.** #2, #4 and #5 are all the same failure mode: a hand-maintained README diverging from the source of truth. A ~20-line test asserting that every scalar `Config` field appears in the README env table would have caught #5 the day it regressed, and would keep catching it.

Guidance says to open an issue before large PRs — this is docs-only and self-contained, so I went straight to a PR. Glad to split it per-claim if you'd rather review them separately.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,9 +35,9 @@ Existing monitoring approaches have gaps:
 - **LLM-based anomaly detection** requires embeddings, a real dependency, and is too expensive to run on every step.
 - **Manual log review** does not scale to week-long unattended runs.
 
-SNAGLINE asks a narrower question: can a zero-dependency, O(1) per-step monitor catch the most common failure modes (loops, error cascades, latency drift) in any agent, running on any framework, with sub-microsecond overhead?
+SNAGLINE asks a narrower question: can a zero-dependency, O(1) per-step monitor catch the most common failure modes (loops, error cascades, latency drift) in any agent, running on any framework, at microsecond-scale overhead?
 
-The answer is yes. SNAGLINE's tier-1 detectors are deterministic, O(1) amortized per step, run with no network calls and no LLM calls, and cost approximately 1 microsecond per `ingest()` call. They run cheaply enough to instrument every step of a production agent.
+The answer is yes. SNAGLINE's tier-1 detectors are deterministic, O(1) amortized per step, run with no network calls and no LLM calls, and cost a few microseconds per `ingest()` call (measured median 1.7--2.4 us/step on Apple silicon; see [Empirical Verification](#empirical-verification) and run `snagline bench` for your own hardware). They run cheaply enough to instrument every step of a production agent.
 
 ## Quick Start
 
@@ -143,9 +143,9 @@ in [docs/RETRAIN_CADENCE.md](docs/RETRAIN_CADENCE.md)).
 
 | Capability | What it gives you |
 |:--|:--|
-| **Zero dependencies** | `pip install snagline-agent` works with nothing but Python 3.10+. Every framework adapter is an optional extra. |
+| **Zero dependencies** | The core needs nothing but Python 3.10+ -- `dependencies = []` in `pyproject.toml`, non-negotiable. Not yet published to PyPI: install with `pip install .` from a clone (see [Quick Start](#quick-start)). Every framework adapter is an optional extra. |
 | **Fail-open guarantee** | Detector/sink exceptions are caught, logged, and never propagated into the host agent. A monitoring library that can crash the thing it monitors is a non-starter. |
-| **Sub-microsecond overhead** | Median 1.9 us/step, p99 33.9 us/step over 200,000 synthetic steps. Cheap enough to run on every step of a week-long run. |
+| **Microsecond-scale overhead** | Median 2.43 us/step, p99 27.71 us/step over 200,000 synthetic steps. Cheap enough to run on every step of a week-long run. Numbers and provenance in [Empirical Verification](#automated-test-suite-and-benchmarks); reproduce with `snagline bench`. |
 | **Framework-agnostic core** | All detector and sink logic operates only on the canonical `StepEvent` schema. Framework-specific code lives in isolated adapter modules and nowhere else. |
 | **No content retention** | Detectors reason about hashes, timings, counts, and booleans -- never prompt or response content. Adoption blocker if left ambiguous. |
 | **Streaming-first, batch-capable** | Primary use is live monitoring of a running agent. The same event schema and detectors also work over an exported trajectory file for offline analysis. |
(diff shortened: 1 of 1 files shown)


Cyrax321 (approved):
Thank you so much for this contribution, @Adhi1-2 — this is exactly the kind of documentation stewardship that makes a project trustworthy.

I went through the README claim by claim against the codebase at the same base you did (`f7857d1`) and re-checked a few of the measurements against current `origin/master` (`8ac01e6`). Every correction holds, and the evidence table in the PR body is exemplary. A few specific callouts:

**What I verified and appreciated:**
- **Overhead:** Confirmed `benchmarks/overhead_benchmark.py` reports median `2.43 us/step` in the verification block, while the intro claimed "sub-microsecond / approximately 1 us" and the features table still showed the superseded `1.9 / 33.9` pair. Your retitling to "microsecond-scale" with a pointer to the verification block and `snagline bench` is the right call.
- **`cusum_min_samples`:** `src/snagline/config.py:175` is indeed `5` and the env-var table already says `5` — the configuration snippet showing `20` was a clear drift since #9. Fixed correctly.
- **Missing env vars:** Checked `dataclasses.fields(Config)` — the 11 fields you added (`semantic_drift_*` 6, `side_effect_*` 3, `compaction_tripwire_*` 2) were indeed undocumented. Thank you for making these discoverable.
- **PyPI:** Verified `https://pypi.org/pypi/snagline-agent/json` still 404s and `Status` already says "Not on PyPI" — the features table advertising `pip install snagline-agent` contradicted it. Your wording ("Not yet published to PyPI: install with `pip install .`") is honest and helpful.
- **Contributing / test setup:** The `addopts = "--cov=snagline"` trap is real — a bare `pytest` fails without the plugin, and there is no `dev` extra. Documenting the exact CI install (`pip install -e . --no-deps` + `pip install pytest pytest-cov ruff mypy`) will save every new contributor that first-run confusion.

**One small follow-up, not blocking:**
The count you recorded (`622 passed, 2 skipped, 88.88%`) was accurate at `f7857d1`. Since then two server fixes landed (`#185`/`#186`), and current `origin/master` is `630 passed, 3 skipped`. I will refresh that line and the commit hash to `8ac01e6` in a tiny follow-up right after this merges so your truth batch does not go stale immediately. No action needed from you.

**Checks:** All CI legs are green (ubuntu 3.10-3.13, windows 3.12/3.13, ml extra, accuracy gate, Gitar approved). Single-file docs change, no runtime behavior touched — exactly the scope this kind of fix should have.

Approving with gratitude. This kind of "read every claim and re-ran every measurement" work is the hardest to do well, and you did it thoroughly. Merging now.

Issue #79, opened 2026-08-23
[adapter] CONTINUUM ledger adapter + sink (free telemetry)


Cyrax321:
### Goal
project.md §6.6/§8, §13 step 6: read CONTINUUM Storage by sequence (public API) and translate ledger entries (PERCEPTION_OBSERVED, BRANCH_RESOLVED, claim/perform/complete) into StepEvents; sink converts FailureRisk -> REQUIRES_REVIEW via request_human.

### Why now
“Free telemetry” advantage: zero new instrumentation on CONTINUUM side, closes the loop back to main project.

### Tasks
- [ ] Verify Storage read-by-sequence signature against current CONTINUUM source
- [ ] continuum_adapter.py (poll + push)
- [ ] continuum_sink.py (optional extra snagline-agent[continuum])
- [ ] Tests with fake Storage

Refs: project.md §6.6, CONTINUUM Storage API

Cyrax321:
Claiming #79

Cyrax321:
Done: PR #164 open with adapter + sink, verified against the real Storage API.

PR #191, opened 2026-08-27
fix(continuum): rate-limit poll storage warnings to one per outage

### LINKED ISSUE #171: [adapter] continuum poll: rate-limit the storage-outage warning so a down CONTINUUM does not spam logs every interval
### Problem

`ContinuumAdapter.poll()` (merged in #164, from issue #79) catches storage
failures fail-open and logs `logger.warning("... read failed for run %s; will retry ...")`
on **every** failing call. In the documented usage, `poll_forever(interval)`
calls `poll()` on a fixed cadence, so a CONTINUUM database that is down for an
hour produces one full traceback-bearing warning per interval (default 1s):
~3600 near-identical log lines per hour per watched run.

This is exactly the operational-noise problem already solved for detectors and
sinks: `Monitor._log_fault_once()` (src/snagline/monitor.py) logs each distinct
fault once and then stays quiet. The adapter predates no such mechanism; it
simply never got one. Fail-open behavior itself is correct and tested
(`test_storage_error_is_swallowed_fail_open`); this is only about log volume.

### Why not fixed in #79

The PR was already carrying two review-bot fixes (ledger-time latency,
bounded pending-claim map). Keeping the merged scope tight; this is pure
hardening with zero behavioral risk to detection.

### Proposed direction

Mirror the Monitor precedent inside the adapter:

- Log the first read failure at WARNING (with traceback), then suppress
  repeats while the failure persists; log ONE recovery line when a poll
  succeeds again.
- Optionally expose the consecutive-failure count via a small counter/
  property so embedders can alert on it instead of scraping logs.
- Keep default behavior fail-open; no exceptions propagate either way.

### Acceptance criteria

- [ ] Persistent outage produces exactly one warning plus one recovery line,
      regardless of poll count (test with a FakeStorage whose `read_events`
      raises N times, asserting caplog record counts).
- [ ] Intermittent failures still surface: recovery followed by a new failure
      logs a new warning.
- [ ] Existing fail-open tests stay green; full gate (pytest, ruff check,
      ruff format --check, mypy src) green.

### Refs

- Continues #79 (CONTINUUM adapter/sink, merged as #164).
- Pattern precedent: Monitor `_log_fault_once` (same rationale as closed #14).

Dedupe note: searched open+closed issues for backoff / log spam / poll retry /
continuum; closest matches are #14 (Monitor-side, closed) and #4 (detector
alert spam, closed). No open or closed tracker item covers adapter poll-log
rate limiting.

Cyrax321:
Fixes #171

Rate-limit `ContinuumAdapter.poll()` storage-outage warnings so a down CONTINUUM does not spam logs every interval.

## Problem

`poll()` logged `logger.warning("... read failed for run %s; will retry ...")` on every failing call. With `poll_forever(interval=1s)` and a down DB for an hour that is ~3600 identical tracebacks per watched run. `Monitor._log_fault_once()` already solved the same noise for detectors and sinks; the adapter never got one.

## Solution

Mirror `Monitor._log_fault_once` inside the adapter. Track `_read_failed` and `_consecutive_failures` per adapter instance:

* First read failure logs at WARNING with traceback then suppresses repeats while failure persists.
* Next successful poll logs one INFO recovery line with the failure count and resets the counter.
* `consecutive_failures` is exposed as a read-only property for host alerting without scraping logs.
* Fail-open unchanged: no exceptions propagate either way; `test_storage_error_is_swallowed_fail_open` stays green.

## Tests

* `test_poll_rate_limits_storage_warnings`: FakeStorage whose `read_events` raises 5 times, assert caplog has exactly 1 warning and `consecutive_failures == 5`, then one INFO recovery after success and counter reset, then no second recovery on empty poll.
* `test_poll_logs_new_warning_after_recovery`: recovery followed by a new failure logs a fresh warning and resets again. Also verifies fail-open throughout.

Existing 15 adapter tests remain green.

## Mutation check

After committing, mutated `if not self._read_failed:` to `if False:` in `continuum_adapter.py`. `test_poll_rate_limits_storage_warnings` went red (0 warnings vs expected 1); `test_poll_logs_new_warning_after_recovery` also red. Reverted via `git checkout` and both tests went green. Proves the rate-limit branch is exercised.

## Gate

pytest 653 passed 3 skipped, ruff check, ruff format --check, mypy src all green at the commit. No em dashes in code, docs, or this body.

### DIFF
--- a/src/snagline/adapters/continuum_adapter.py
+++ b/src/snagline/adapters/continuum_adapter.py
@@ -139,8 +139,21 @@ class ContinuumAdapter:
         #: action key -> epoch seconds of the observed claim, for latency pairing
         self._pending_claims: dict[str, float] = {}
         self._stop = threading.Event()
+        # Rate-limited read-failure logging (issue #171): mirror
+        # Monitor._log_fault_once so a down DB does not spam every interval.
+        self._read_failed = False
+        self._consecutive_failures = 0
 
     # -- public API -------------------------------------------------------- #
+
+    @property
+    def consecutive_failures(self) -> int:
+        """Number of consecutive read failures since the last success.
+
+        Exposed for host alerting without scraping logs (issue #171).
+        Resets to 0 on the next successful poll.
+        """
+        return self._consecutive_failures
 
     def poll(self) -> int:
         """Read and ingest every entry after the cursor. Returns count ingested.
@@ -156,12 +169,23 @@ class ContinuumAdapter:
                 self._storage.read_events(self._run_id, after_sequence=self._cursor)
             )
         except Exception:
-            logger.warning(
-                "snagline continuum adapter: read failed for run %s; will retry",
-                self._run_id,
-                exc_info=True,
-            )
+            self._consecutive_failures += 1
+            if not self._read_failed:
+                logger.warning(
+                    "snagline continuum adapter: read failed for run %s; will retry",
+                    self._run_id,
+                    exc_info=True,
+                )
+                self._read_failed = True
             return 0
+        if self._read_failed:
+            logger.info(
+                "snagline continuum adapter: read recovered for run %s after %d failure(s)",
+                self._run_id,
+                self._consecutive_failures,
+            )
+            self._read_failed = False
+        self._consecutive_failures = 0
         count = 0
         for entry in entries:
             if self.push(entry):
--- a/tests/adapters/test_continuum_adapter.py
+++ b/tests/adapters/test_continuum_adapter.py
@@ -382,3 +382,67 @@ def test_modules_import_without_continuum(monkeypatch: pytest.MonkeyPatch) -> No
     sink_mod = importlib.import_module("snagline.sinks.continuum_sink")
     assert adapter_mod.ContinuumAdapter is not None
     assert sink_mod.ContinuumSink is not None
+
+
+def test_poll_rate_limits_storage_warnings(
+    caplog: pytest.LogCaptureFixture,
+) -> None:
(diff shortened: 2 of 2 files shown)


Cyrax321:
Thank you for this hardening — exactly the right follow-up to #79.

**What I verified:**
- The problem is real: `poll()` logging WARNING on every failure with `exc_info=True` at 1s interval = ~3600 tracebacks per hour per run. Your fix mirrors `Monitor._log_fault_once()` perfectly.
- Implementation: `_read_failed` flag + `_consecutive_failures` counter, first failure logs WARNING with traceback and sets flag, repeats suppressed, next success logs single INFO recovery with count and resets, property exposed for alerting. Fail-open unchanged — no exception propagates (existing `test_storage_error_is_swallowed_fail_open` still green).
- Tests: `test_poll_rate_limits_storage_warnings` asserts 5 consecutive raises → exactly 1 WARNING + `consecutive_failures==5`, then 1 recovery INFO after success and counter reset, then no duplicate recovery on empty poll. `test_poll_logs_new_warning_after_recovery` proves intermittent recovery → new failure logs fresh warning. Both use `caplog` record counts, not log text matching — precise.
- Mutation check: `if not self._read_failed:` → `if False:` makes both tests red (0 vs 1 warning), reverted — proves the branch is exercised.
- Gate: you report 653 passed / 3 skipped, ruff/mypy clean, no em dashes, and CI confirms CLEAN across all 6 Test legs + accuracy gate + Gitar. Small, solo zone (`continuum_adapter.py` only) — no other agent touched this file.

This is the exact operational-noise fix the adapter needed before the sidecar runs for weeks. Merging with appreciation.

PR #192, opened 2026-08-27
fix: add dev extra so bare pytest works after pip install

### LINKED ISSUE #182: [tests] No dev extra in pyproject.toml: bare `pytest` fails on unrecognized --cov after `pip install .`
## Summary

`pyproject.toml` sets

```toml
[tool.pytest.ini_options]
addopts = "--cov=snagline --cov-fail-under=80"
```

but declares no `dev` (or `test`) extra, and `dependencies = []`. So the documented-looking path fails:

```console
$ pip install .
$ pytest
ERROR: usage: pytest [options] [file_or_dir] ...
pytest: error: unrecognized arguments: --cov=snagline --cov-fail-under=80
```

`pytest-cov` is never installed, but `addopts` unconditionally passes its flags. A new contributor has to work out from the traceback that they need a plugin nobody told them about.

## What actually works

Reverse-engineered from `.github/workflows/ci.yml`:

```bash
pip install -e . --no-deps
pip install "pytest" "pytest-cov==7.1.0" "ruff==0.16.3" "mypy"
ruff check src tests
ruff format --check src tests
mypy src
python -m pytest tests/ -q
```

Nothing in the repo states this outside the workflow file.

Second-order effect: because `addopts` forces `--cov`, even `pytest --collect-only` fails with `Required test coverage of 80% not reached` — so the usual "just list the tests" move doesn't work as an escape hatch either.

## Suggested fix

Add the extra, pinned to match CI so local runs and CI can't diverge:

```toml
[project.optional-dependencies]
dev = [
  "pytest",
  "pytest-cov==7.1.0",
  "ruff==0.16.3",
  "mypy",
]
```

Then `pip install -e ".[dev]" && pytest` works from a fresh clone, and `ci.yml` can install `.[dev]` instead of maintaining a parallel pin list.

Keeping `dependencies = []` for the runtime core is the whole point of the project and this doesn't touch it — extras are not runtime deps.

I documented the manual commands in the Contributing section in #179, but that's a workaround for the missing extra rather than a fix.

Happy to send the PR — it's a few lines plus the `ci.yml` swap.

Cyrax321:
Fixes #182

Adds `dev` extra so `pip install .[dev]` (or `-e .[dev]`) provides the tooling that `pyproject.toml`'s `addopts = "--cov=..."` requires.

Changes:
- `pyproject.toml`: `dev = ["pytest", "pytest-cov==7.1.0", "ruff==0.16.3", "mypy"]` pinned to match `.github/workflows/ci.yml`
- `README.md`: Local development setup now documents `pip install -e ".[dev]"`

Verification:
- Fresh venv: `pip install .` then `pytest` fails with `unrecognized arguments: --cov` (pre-fix, reproduced)
- Fresh venv: `pip install ".[dev]"` then `pytest -q` passes (651 passed, 3 skipped, coverage 88.51%)
- `pytest --collect-only -q` exits 0 with no unrecognized argument (652 collected)
- `pip install -e .` alone still has zero runtime deps and `import snagline` works (stdlib-only venv)
- Full gate green: `pytest`, `ruff check`, `ruff format --check`, `mypy src`

Keeps `dependencies = []` and `addopts` unchanged; extras are not runtime deps.

Refs #179

### DIFF
--- a/README.md
+++ b/README.md
@@ -751,14 +751,10 @@ Contributions are welcome. This project is open source under MIT and deliberatel
 
 ### Local development setup
 
-The core has no runtime dependencies, but the test suite needs `pytest` and
-`pytest-cov` (`[tool.pytest.ini_options] addopts` enables coverage, so a bare
-`pytest` fails with an unrecognised `--cov` until the plugin is installed).
-There is no `dev` extra yet, so install the tooling the same way CI does:
+The core has no runtime dependencies; dev tooling is in the `dev` extra:
 
 ```bash
-pip install -e . --no-deps
-pip install pytest pytest-cov ruff mypy
+pip install -e ".[dev]"
 python -m pytest tests/ -q        # 622 passed, 2 skipped
 ruff check src tests && ruff format --check src tests
 mypy src
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,8 @@ continuum  = ["continuum-agent"]
 # Optional extras (not required for the core fail-open detector):
 ml         = ["numpy>=1.24", "scikit-learn>=1.3"]   # ESN ensemble (issue #80)
 drift      = ["sentence-transformers>=2.2"]          # semantic goal-drift (issue #81)
+# Dev extra for local development: pins match .github/workflows/ci.yml
+dev        = ["pytest", "pytest-cov==7.1.0", "ruff==0.16.3", "mypy"]
 # Slack and PagerDuty sinks ship in the core and use only the stdlib, so they
 # need no extra dependency; the `all` extra pulls in the framework/SDK extras.
 all        = ["snagline-agent[langchain,langgraph,autogen,crewai,openai,anthropic,ml,drift,continuum]"]


Cyrax321:
Thank you for closing this DX trap — the kind of small fix that saves every new contributor an hour of confusion.

**What I verified:**
- The failure is real: `pyproject.toml` addopts forces `--cov` but no `dev` extra exists, so `pip install .` + `pytest` → "unrecognized arguments: --cov..." and even `pytest --collect-only` fails with coverage error. Reverse-engineered CI install was the only working path.
- Fix: `dev = ["pytest", "pytest-cov==7.1.0", "ruff==0.16.3", "mypy"]` pinned to match `.github/workflows/ci.yml`, keeping `dependencies = []` and `addopts` unchanged — extras are not runtime deps, as required.
- README update: Local development setup now correctly documents `pip install -e ".[dev]"` (removed obsolete two-step manual install). Matches what a newcomer actually needs.
- Verification: you report fresh venv `pip install ".[dev]"` → `pytest -q` 651/3, `pytest --collect-only` 652 collected exit 0, `pip install .` alone still zero-dep and `import snagline` works, and `pip install .` + `pytest` still fails as expected pre-fix — all captured. Full gate ruff/mypy clean, no em dashes, and CI confirms CLEAN.

Small, correct, and pinned to CI so local and CI cannot diverge. Merging with appreciation for the thorough fresh-venv proof.

Issue #92, opened 2026-08-24
[core] Horizon-scale config: wall-clock budgets, idle detection, window auto-scaling + heartbeat sink


Cyrax321:
## Summary

Make SNAGLINE horizon-aware. Three additions, one theme: fixed step-count assumptions break on multi-day episodes where step rates vary by orders of magnitude between phases (an hour of tool grinding vs a long LLM call). This issue adds wall-clock budgets, window auto-scaling, idle detection, and a heartbeat mechanism so *silence itself* becomes detectable. In-band detectors can never see a hung host: no events means no observe() calls. Only a clock or an external watcher can close that hole.

## Why

- A wedged process (deadlocked tool call, hung connection, OOM-stopped worker) is the most common multi-hour failure shape and SNAGLINE is structurally blind to it today. Durable-execution runtimes solve this with worker leases and heartbeats (LangChain production runtime docs); monitoring libraries that only consume events inherit the blindness unless they add a time axis.
- Fixed windows (`loop_window_size=12`) are tuned for interactive runs. Across a 500k-step week-long episode they either fire constantly or mean nothing; baselines fitted once go stale as workload phases shift.
- Wall-clock budget awareness pairs with the TokenRunaway envelope (#token-runaway sibling) to cover the two resources weeks-long runs exhaust: money and patience.

### 1. Time-aware Config

```python
@dataclass
class Config:
    ...
    max_episode_wall_seconds: float | None = None   # deterministic breach risk at limit
    warn_fraction: float = 0.8                       # single warning risk before breach
    idle_warn_seconds: float | None = None           # gap between consecutive ingests
    window_scale_steps: int = 0                      # >0 enables auto-scaling below
```

`Monitor.ingest()` computes inter-event deltas from event timestamps (no wall-clock reads in the hot path; replay-safe by construction). Idle/budget risks carry triggers `idle_gap` / `wall_clock_budget`.

### 2. Window auto-scaling

Detectors accepting `window_size` grow it as `base * ceil(episode_len / window_scale_steps)` capped at `max_window`, re-fitted lazily. Latency CUSUM gains optional periodic baseline re-fit (`cusum_refit_every`, Welford restart with frozen-old-baseline comparison so drift in the *baseline itself* is visible). Defaults preserve current behavior exactly (scaling off).

### 3. HeartbeatSink (external liveness)

```bash
snagline watch --file events.jsonl --follow --heartbeat /var/run/snagline/hb
```

[...]

Issue #18, opened 2026-08-15
[replay] replay() does not call end_episode, causing detector state leakage across calls


Cyrax321:
### System Information
- OS Platform and Distribution: macOS (Apple Silicon)
- Language / runtime version: Python 3.13.5
- Package / library version(s): snagline-agent 0.1.0
- Installation method: source

### Describe the current behavior
`replay()` in `src/snagline/cli.py` reads a JSONL trajectory and calls `monitor.ingest()` for each event, but never calls `monitor.end_episode()` when done. When the same monitor is reused across multiple `replay()` calls, detector state (loop windows, cascade counters, CUSUM baselines) from the first replay leaks into the second.

Compare with the `watch` CLI, which correctly calls `monitor.end_episode(episode)` in its `finally` block (`src/snagline/cli.py:211`).

### Describe the expected behavior
`replay()` should call `monitor.end_episode()` after replaying, or at minimum document that callers must do so. The ``watch`` CLI already does this correctly.

### Standalone code to reproduce the issue
```python
from snagline.cli import replay, _CountingSink
from snagline import Monitor
import tempfile, os

# File A: 3 identical sigs (loop)
with tempfile.NamedTemporaryFile(mode='w', suffix='.jsonl', delete=False) as f:
    for i in range(3):
        f.write('{"step_id": "' + str(i) + '", "episode_id": "ep-1", "timestamp": 1.0, "action_type": "tool_call", "action_signature": "SIG"}\n')
    file_a = f.name

# File B: 2 more SIG + 1 unique
with tempfile.NamedTemporaryFile(mode='w', suffix='.jsonl', delete=False) as f:
    for i in range(3):
        sig = 'SIG' if i < 2 else 'UNIQUE'
        f.write('{"step_id": "' + str(i+3) + '", "episode_id": "ep-1", "timestamp": 1.0, "action_type": "tool_call", "action_signature": "' + sig + '"}\n')
    file_b = f.name

counter = _CountingSink()
monitor = Monitor.default(sinks=[counter])

replay(file_a, monitor=monitor)
print(f'File A: {counter.count} risks')  # 1 (correct)

counter.count = 0
counter.risks = []
replay(file_b, monitor=monitor)
print(f'File B: {counter.count} risks')  # 2 (BUG: expected 0, got 2)

os.unlink(file_a)
os.unlink(file_b)
```

### Other info / logs
Output:
```
File A: 1 risks
[...]

PR #194, opened 2026-08-27
fix(monitor): bound per-episode state with LRU cap and fix clock retention

### LINKED ISSUE #184: [core] Monitor per-episode state has no cap or TTL: unbounded memory growth for any host that never calls end_episode, and episodes_active saturates so the gauge hides it
## Summary

`Monitor`'s per-episode state has **no cap, no TTL, and no eviction of any kind**. It is released only when someone explicitly calls `Monitor.end_episode(episode_id)`. Every episode id that is never explicitly ended is retained for the lifetime of the process, in every detector's per-episode dict plus `Monitor._clocks`.

For a library whose core promise is O(1)-per-step with bounded memory, and which ships a long-lived sidecar server as a headline integration path, this is unbounded growth on the primary deployment shape.

`#1`, `#18`, and `#67` were each this bug at one specific call site (`raw.watch`, `replay()`, `MemoryStateBackend`), fixed caller-by-caller. `Monitor` itself still has no backstop, so the fourth caller reopens it — and for the sidecar, the "caller" is a remote HTTP client the library cannot fix.

## Measured

`Monitor.default()`, N distinct episode ids, no `end_episode`, `tracemalloc`-traced heap:

| episodes | `_clocks` | detector-state entries | traced heap |
|--:|--:|--:|--:|
| 100 | 100 | 900 | 0.66 MB |
| 1,000 | 1,000 | 9,000 | 3.72 MB |
| 10,000 | 10,000 | 90,000 | 36.58 MB |

Perfectly linear, no plateau. Same run with `end_episode()` called: **`_clocks=0`, detector-state entries=0, heap 0.01 MB** — so teardown works correctly and completely; nothing forces it to happen.

Retention scales with episode shape, because `LatencyAnomalyDetector._states` is keyed per `(episode, tool)`:

| steps/ep | distinct tools | bytes/episode | growth @ 1 new episode/sec |
|--:|--:|--:|--:|
| 3 | 3 | 4,034 | 349 MB/day |
| 13 | 5 | 5,640 | 487 MB/day |
| 30 | 8 | 7,101 | 614 MB/day |
| 60 | 12 | 10,493 | 907 MB/day |
| 200 | 20 | 14,196 | 1,227 MB/day |

Repro:

```python
from snagline import Config, Monitor, StepEvent, make_signature
mon = Monitor.default(config=Config(), sinks=[])
for e in range(10_000):
    mon.ingest(StepEvent(step_id=f"{e}-0", episode_id=f"ep-{e}", timestamp=1000.0,
        action_type="tool_call", action_signature=make_signature("tool_call", "search", str(e)),
        tool_name="search", latency_ms=100.0, error=False))
print(len(mon._clocks))  # 10000 -- never shrinks
```

## Reachable through the shipped sidecar

`POST /episodes/end` (#123 / #160) is the only thing that frees this, and it is advisory. Any host that doesn't send it leaks: a crashed or OOM-killed agent, a dropped connection, a Claude Code hook whose process exits, a pre-#160 client, or any integration whose author read the docs as "cleanup is optional" — which is how they read today:

- `README.md:570` — "Per-episode state is isolated and **can be cleared** via `monitor.end_episode()`"
- `README.md:601` — "Clear per-episode detector state."

Neither says memory grows without bound otherwise. `docs/ADAPTER_GUIDE.md:32` is the only place that comes close.

## The gauge you would watch hides it

`SidecarMetricsCollector._episodes` is LRU-capped at `_MAX_TRACKED_EPISODES = 10_000`. `Monitor` is not capped. So `snagline_episodes_active` **saturates** while real retained state keeps climbing.

Driving the real server over HTTP (`POST /events`, never `/episodes/end`), with the collector cap lowered to 50 so saturation is reachable quickly — the shipped 10,000 behaves identically, just later:

```
 episodes POSTed  episodes_active gauge  retained state entries
              10                     10                      80
              50                     50                     400
             100                     50                     800
             200                     50                    1600
             300                     50                    2400
```

The gauge flatlines exactly when the leak starts mattering. An operator watching `snagline_episodes_active` sees a reassuring plateau while RSS climbs, and there is no exported metric for retained detector state.

## This corrects an assumption in #173

#173 (open, TTL for the gauge id table) says:

> the same `_MAX_TRACKED_EPISODES` cap continues to bound memory

That is true of the collector's id table and not true of the process. #173 is a metrics-accuracy issue on a table that is already bounded; this issue is about the unbounded thing sitting behind it. Worth fixing in the right order — a TTL on the gauge alone would make `episodes_active` *more* self-healing while the actual leak stays invisible.

## Also: `_clocks` grows when the feature is off

`monitor.py:211-215` says the horizon time axis is:

> Inert unless one of the opt-in horizon knobs is set on the config; all state lives in ``_clocks``

But `_clocks` is populated unconditionally:

```
max_episode_wall_seconds=None  idle_warn_seconds=None   (both unset -- feature off)
len(Monitor._clocks) after 500 episodes = 500
```

So a user who never opts into the horizon axis still pays per-episode retention for it. Small next to the detector dicts, but it's the same bug in the one place the code claims to be inert, and it's a two-line fix independent of the rest.

## Suggested fix

A backstop in `Monitor`, so correctness stops depending on every current and future caller:

- Bounded LRU registry of live episode ids in `Monitor`, keyed by last-seen, with a config knob (`SNAGLINE_MAX_LIVE_EPISODES`, default high enough — 10k? — that no normal workload changes behavior).
- On eviction, run the same teardown as `end_episode`. **LRU by last-seen is the safe policy here:** a genuinely active episode is by definition recently seen, so it cannot be evicted out from under itself. A fixed TTL would need more care.
- One decision to make explicitly: whether eviction should run `finalize()` (dispatching e.g. a `silent_abort` risk for an episode that vanished) or tear down silently. Firing risks for an evicted episode is defensible — "the agent went away mid-run" is real information — but it's a behavior change and should be opt-in, not a side effect of a memory fix.
- Guard `_clocks` behind the horizon knobs so the "inert" claim holds.
- Export retained-episode count as a metric, so the leak is observable rather than inferred from RSS.
- Regression test: ingest N episodes without `end_episode`, assert retained state stays bounded. That is the test none of #1/#18/#67 left behind, which is why this keeps coming back.
- Docs: state plainly that `end_episode` is how state is freed, and what the backstop does when it isn't called.

Happy to send a PR. The `_clocks` gating and the metric are self-contained; the eviction policy and the `finalize()`-on-eviction question are semantics calls I'd rather you settle first.

## Environment

Clean `upstream/master` @ `f7857d1`, CPython 3.13.5, Apple M4, stock `Config()`.

Refs #1, #18, #67, #123, #160, #172, #173

Cyrax321:
Fixes #184

Monitor retained every episode id never explicitly ended in `_clocks` plus every detector's per-episode dict, with no cap, no TTL, and no eviction. `end_episode` freed correctly, but nothing forced it to happen, so a long-lived sidecar grew without bound. The `snagline_episodes_active` gauge is LRU capped at 10000, so it saturated while real retained state kept climbing and the leak was invisible from metrics. `_clocks` also grew even when the horizon feature was off, contradicting its own inert claim.

Chosen fix: bounded LRU registry of live episode ids in Monitor, keyed by last-seen, with a config knob `max_live_episodes` (default 10000 to match the existing metrics cap, env `SNAGLINE_MAX_LIVE_EPISODES`). Every `ingest` touches the id as most-recent; when the cap would be exceeded the least-recently-seen id is evicted silently via the same teardown as `end_episode` but without `finalize` risks. Explicit `end_episode` still frees immediately; the cap is the safety net, not the primary path. LRU by last-seen is safe because a genuinely active episode is by definition recently seen and cannot be evicted out from under itself. A fixed TTL would need more care and is left for later if needed. Guard `_clocks` behind the horizon knobs so the inert claim holds, and expose retained count as `monitor.retained_episodes`, `metrics()["retained_episodes"]` and prometheus gauge `snagline_monitor_retained_episodes` so the leak is observable rather than inferred from RSS. No content retention beyond ids and timestamps.

Docs: README now says MUST be cleared via `end_episode` or it will be evicted, documents the new knob and metric; ADAPTER_GUIDE updated.

Verification after fix (same repro as issue, `Monitor.default(config=Config(max_live_episodes=10000), sinks=[])`, `tracemalloc` current):

Cap 10000, horizon OFF (default), bounded vs previously linear:

| episodes | retained | _clocks | det entries | heap |
|---:|---:|---:|---:|---:|
| 100 | 100 | 0 | 200 | 0.49 MB |
| 1000 | 1000 | 0 | 2000 | 2.28 MB |
| 10000 | 10000 | 0 | 20000 | 22.23 MB |
| 20000 | 10000 | 0 | 20000 | 24.17 MB |

Previously 10000 = 36.58 MB linear and 20000 would be ~70 MB unbounded; now 20000 plateaus at 10000 retained. With cap 100, 1000 episodes stays at 100 retained, 0.26 MB. With horizon ON and cap 100, both `_clocks` and retained stay at 100 for 200 and 500 episodes.

Tests: `tests/test_per_episode_cap.py` covers cap eviction fires, LRU ordering (touch moves to end, oldest evicted), gauge drops via `metrics()["retained_episodes"]`, detector state freed for evicted ids, explicit `end_episode` still frees immediately and allows re-ingest, `_clocks` not retained when horizon off and bounded with cap when on, retained metric exposed and prometheus renders, eviction silent (no `silent_abort` on evict) and explicit end still judges.

Gate: `pytest -q -o addopts=""` 659 passed, 3 skipped; `ruff check src tests` passed; `ruff format --check` passed; `mypy src` passed.

Mutation checks (commit fix first, then mutate committed file, then `git checkout --`):

1. Disabled LRU eviction (`if len(...) > max` to `if False and ...`): `test_cap_evicts_lru_and_frees_detector_state` and `test_bounded_memory_repro` went red (retained 10001 vs 10000).
2. Disabled `_clocks` guard (`if max_wall is None and idle is None: return` to `if False and ...`): `test_clocks_not_retained_when_horizon_off` went red (len 50 vs 0).

Refs #184, Refs #1, Refs #18, Refs #67, Refs #92, Refs #123

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,7 @@ the path variant below.
 | `SNAGLINE_HALT_URL` | `halt_url` | *(unset)* | Halt webhook endpoint; required when policy is halt_webhook |
 | `SNAGLINE_HALT_TIMEOUT_S` | `halt_timeout_s` | 0.25 | Halt webhook round-trip budget in seconds; timeout fails open to continue |
 | `SNAGLINE_MIN_SEVERITY_FOR_HALT` | `min_severity_for_halt` | 0.8 | Minimum risk score that pays the halt-webhook cost |
+| `SNAGLINE_MAX_LIVE_EPISODES` | `max_live_episodes` | 10000 | Per-episode LRU cap; when exceeded the least-recently-seen episode is evicted silently (no finalize). Explicit `end_episode` still frees immediately |
 
 A handful of variables sit outside `Config` because they are consumed
 directly by one component each:
@@ -594,7 +595,7 @@ snagline watch --file /var/log/agent/events.jsonl --follow
 | **FailureRisk** | The signal. Carries no raw content, no metadata. Just ids, score, trigger, detail, and timestamp. |
 | **Sink** | The escalation path: `emit(risk) -> None`. Fire-and-forget, never blocks ingest. |
 | **Action signature** | A one-way SHA-256 digest of the logical action. Volatile fields (timestamps, nonces, retry counters) must be excluded so retries look like retries, not unique actions. |
-| **Episode** | A logical unit of work (a single agent run, a user session). Per-episode state is isolated and can be cleared via `monitor.end_episode()`. |
+| **Episode** | A logical unit of work (a single agent run, a user session). Per-episode state is isolated and MUST be cleared via `monitor.end_episode()` when the episode is finished; otherwise the Monitor evicts the least-recently-seen episode once `max_live_episodes` (default 10000, env `SNAGLINE_MAX_LIVE_EPISODES`) is exceeded. Eviction is silent (no finalize risks) and retains only episode ids, never content. |
 | **Config** | All tunable thresholds in one dataclass. Sensible defaults ship so zero configuration works. |
 
 ## Architecture
@@ -625,7 +626,7 @@ from snagline import Monitor, StepEvent, FailureRisk, Config, make_signature, wa
 |:--|:--|
 | `Monitor` | Orchestrator. `Monitor.default()` returns a ready-to-use instance. |
 | `Monitor.ingest(event)` | Run all detectors against one event. Never raises (fail-open). |
-| `Monitor.end_episode(episode_id)` | Clear per-episode detector state. |
+| `Monitor.end_episode(episode_id)` | Clear per-episode detector state. MUST be called when an episode finishes; otherwise state is retained until LRU eviction. The retained count is exposed as `monitor.retained_episodes` and `snagline_monitor_retained_episodes` (prometheus) and `metrics()["retained_episodes"]`. |
 | `StepEvent` | Frozen dataclass. The canonical event schema. |
 | `FailureRisk` | Frozen dataclass. The detection signal. |
 | `Config` | Dataclass. All tunable thresholds. |
--- a/docs/ADAPTER_GUIDE.md
+++ b/docs/ADAPTER_GUIDE.md
@@ -31,6 +31,10 @@ class MyAdapter:
 
 When the run finishes, call `monitor.end_episode(episode_id)` so per-episode
 detector state (loop windows, CUSUM baselines) doesn't leak into the next run.
+If you never call it, the Monitor evicts the least-recently-seen episode once
+`max_live_episodes` (default 10000, env `SNAGLINE_MAX_LIVE_EPISODES`) is
+exceeded. Eviction is silent (no finalize risks) and the retained count is
+exposed as `monitor.retained_episodes`.
 If your framework is a context manager or generator, put that teardown in the
 `finally` - see `adapters/raw.py` and `adapters/langchain_adapter.py:close`.
 
--- a/src/snagline/config.py
+++ b/src/snagline/config.py
@@ -149,6 +149,14 @@ def validate_policy(value: str) -> str:
     return normalized
 
 
+def _validated_max_live_episodes(cfg: Config) -> None:
+    """Validate the per-episode cap (issue #184); raise when invalid."""
+    if cfg.max_live_episodes < 1:
+        raise ValueError(
+            f"max_live_episodes must be >= 1; got {cfg.max_live_episodes!r}"
+        )
+
+
 def _validated_stagnation(cfg: Config) -> None:
     """Validate the stagnation-detector knobs (issue #132); raise when invalid.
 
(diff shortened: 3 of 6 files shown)


Cyrax321:
Thank you for fixing the core O(1) promise — unbounded growth on the sidecar is now bounded.

**What I verified:**
- The leak is real: `Monitor.default()`, 10k distinct episode ids without `end_episode`, tracemalloc shows 36.58 MB linear, `_clocks=10000` never shrinks. With `end_episode` it is 0. Your fix mirrors the fix's measurement and closes it.
- Implementation: bounded LRU with configurable `max_live_episodes` (default 10000 to match the sidecar cap, env `SNAGLINE_MAX_LIVE_EPISODES`). Every ingest touches the id as most-recent via `OrderedDict`-style; when cap would be exceeded the least-recently-seen id is evicted silently via same teardown as `end_episode` (no finalize risks). Explicit `end_episode` still frees immediately — the cap is the safety net.
- Horizon clock guard: when both `max_episode_wall_seconds` and `idle_warn_seconds` are unset, no `_clocks` entry is created at all — previously every host paid retention even without opting in. Now the inert claim holds.
- Observability: `monitor.retained_episodes`, `metrics()["retained_episodes"]`, and Prometheus `snagline_monitor_retained_episodes` expose the count so the leak is observable, not inferred from RSS. Docs: README now says MUST be cleared or it will be evicted, documents the new knob and metric; `ADAPTER_GUIDE` updated.
- Tests: `tests/test_per_episode_cap.py` (161 lines) asserts cap eviction fires, gauge drops, detector state freed, explicit end still works, and clock retention is fixed. No content retention beyond ids/timestamps.
- Gate: CI is CLEAN across all legs.

This keeps the long-lived sidecar honest. Merging with appreciation.

Issue #215, opened 2026-08-28
[tests] hook-bridge stub sink reads one TCP segment: the loop payload is discarded, so the follow-mode teardown test fails ~75% of runs (#156 fix cannot help)


abyyxhek:
## Summary

`_HookTargetServer._serve` in `tests/test_hook_bridge.py` reads each webhook POST with **one** `conn.recv(65536)` and records only that single result in `self.bodies`:

https://github.com/Cyrax321/SNAGLINE/blob/faa4a9e255607be656665b90bfa93a2c199dc2cb/tests/test_hook_bridge.py#L97-L116

```python
data = conn.recv(65536)
if self.respond_ok and data:
    self.bodies.append(data)
    conn.sendall(b"HTTP/1.0 200 OK\r\nContent-Length: 0\r\n\r\n")
```

`WebhookSink` posts through `urllib.request`, and `http.client.HTTPConnection._send_output()` writes the request in **two** separate `send()` calls — the header block first, then the body as its own chunk. `http.client` also sets `TCP_NODELAY` on the socket, so the two writes reach the listener as two separate segments with nothing forcing them to coalesce.

When the server's single `recv()` consumes the header segment alone, the JSON body is thrown away. The body is the only place `"trigger": "loop"` ever appears, so `test_watch_follow_teardown_reaps_the_child_on_a_healthy_run` fails — its 30-second delivery poll can never succeed, and the final assertion at line 430 fails too.

This is a hard failure, not a slow pass. `WebhookSink.emit` is fire-and-forget with no retry, and the loop detector alerts once per finding (#4 / #94), so there is exactly **one** POST per run. If its body segment is missed, no later read can recover it and no amount of extra deadline helps.

## Reproduced on `master` (faa4a9e)

Whole-suite run, Windows 11 / CPython 3.13.1, commit `faa4a9e`:

```
1 failed, 696 passed, 2 skipped in 385.18s
FAILED tests/test_hook_bridge.py::test_watch_follow_teardown_reaps_the_child_on_a_healthy_run
```

Failure output — note `Content-Length: 151` and then **nothing after the blank line**. The recorded "body" is the header block:

```
Failed: watcher never delivered the loop risk to the sink endpoint within 30s;
bodies=[b'POST /sink HTTP/1.1\r\nAccept-Encoding: identity\r\nContent-Length: 151\r\n
[...]

Issue #149, opened 2026-08-26
[detectors] StagnationDetector and SideEffectGuardDetector lack snapshot/restore while DETECTOR_GUIDE claims all shipped detectors implement it


Cyrax321:
## Summary

`Monitor.snapshot()`/`restore()` (built in #91) is the restart-survivability guarantee for long-running episodes. Two opt-in detectors do **not** implement the `StatefulDetector` (`dump_state()`/`load_state()`) pair today, and `docs/DETECTOR_GUIDE.md` incorrectly claims all of them do:

- `src/snagline/detectors/stagnation.py` (#87): per-episode novelty windows and the all-time seen-signature set are lost on restart.
- `src/snagline/detectors/side_effect_guard.py` (#88, merged via #140): per-episode duplicate counters are lost on restart.

The guide says, under "Restart-survivable state":

> All shipped detectors implement it, and `DedupSink` persists cooldowns when used with its default key function.

That sentence is wrong on current master (`32e81c8`).

## Evidence (verified against origin/master 32e81c8)

```
$ grep -L "def dump_state" src/snagline/detectors/*.py
src/snagline/detectors/__init__.py
src/snagline/detectors/side_effect_guard.py
src/snagline/detectors/stagnation.py
```

Every other shipped detector module implements the pair. Behavior at the Monitor layer:

- `monitor.py` `snapshot_dict()`: `dump() if callable(dump) else None`, so both detectors serialize as `null`.
- `monitor.py` `restore_dict()`: `if not callable(load): continue`, so a restored monitor skips them silently, with no warning.

Consequence: a restart mid-episode resets duplicate counting. A repeated non-idempotent action spanning the restart will never fire `side_effect_duplicate`, even with `side_effect_guard_enabled=True` and snapshots taken on every deploy. Same for an in-progress stagnation window. The loss is silent, which is exactly the failure mode #91 exists to prevent.

## Impact

This hits hardest precisely where #88 matters most: week-long unattended runs, where deploys/crashes/restarts are routine and a replayed payment/send/deploy across a restart is the canonical incident the guard was built to catch.

## Suggested fix

1. Implement `dump_state()`/`load_state()` on both detectors: JSON-compatible per-episode counters/windows, following the `LoopDetector` pattern; add round-trip tests mirroring `tests/test_monitor_snapshot.py`. Note the guard's counters are plain nested dicts of ints, so serialization is trivial; stagnation needs its deque-of-bools plus seen-set encoded.
[...]

Issue #91, opened 2026-08-24
[core] Monitor.snapshot()/restore(): restart-survivable detector state


Cyrax321:
## Summary

Add `Monitor.snapshot(path)` / `Monitor.restore(path)` (plus `Monitor.to_dict()/from_dict()`): serialize all per-episode detector state (deques, Welford/CUSUM statistics, counters) to JSON so monitoring survives process restarts. Today a deploy or crash wipes every baseline mid-run; on day 3 of a week-long run that means blind detectors until warm-up completes again, silently.

This is the single highest-value infrastructure item for SNAGLINE's stated purpose ("cheap enough to run on every step of a week-long unattended run").

## Why

- All detector state currently lives in instance memory. Any process restart resets: loop windows (loop detector blind for `window_size` steps), CUSUM baselines (blind for `cusum_min_samples`=20 events *per tool*), dedup cooldowns, goal-drift comparisons.
- Durable-execution harnesses checkpoint agent state every superstep precisely so runs survive deploys (LangChain production runtime docs); a companion monitor that loses its own state on every restart contradicts the deployment regime it advertises for.
- Competitor gap: none of agent-vitals / agentguard / pisama advertise restart-survivable detector state. Owning this makes "the reliability layer for multi-day episodes" credible.

## Design constraints

- **stdlib JSON, not pickle** (security + debuggability). Each detector implements two optional methods extending the protocol:

```python
class Detector(Protocol):
    ...
    def dump_state(self) -> dict | None: ...      # default None
    def load_state(self, state: dict) -> None: ...
```

- `Monitor.snapshot(fp)` gathers `{detector.name: dump_state()}` plus a format-version header and config fingerprint (threshold values hashed so restoring into a differently-configured monitor warns instead of misbehaving quietly).
- Atomic write discipline: temp file + `os.replace`, matching the project's fail-safe instincts.
- `snagline watch --state state.json --save-interval 1000` persists periodically in live mode; `replay` gains `--resume-state`.
- DedupSink cooldown timestamps included so alert suppression also survives restarts.

## Acceptance criteria

- [ ] Round-trip property test: ingest N events, snapshot, restore into a fresh Monitor, continue ingestion; risks emitted identical to the never-restarted twin Monitor (golden comparison).
- [ ] Version mismatch and config-fingerprint mismatch raise clear errors (fail-fast on restore is correct; fail-open applies to runtime, not setup).
[...]

PR #219, opened 2026-08-28
fix(detectors): restore the effective stagnation window, not the base (Fixes #218)

### LINKED ISSUE #218: [detector] StagnationDetector.load_state clamps restored novelty flags to the base window, so snapshot restore blinds the detector when window scaling is on
### Issue type

Bug

Yes

### Source

source

### SNAGLINE version

0.1.1.dev0 / faa4a9e255607be656665b90bfa93a2c199dc2cb

### Custom code

No

Windows-11-10.0.26200-SP0

### Python version

3.13

### Current behavior?

**Affected component:** `src/snagline/detectors/stagnation.py` — `StagnationDetector.load_state`

With window auto-scaling enabled (`Config.window_scale_steps > 0`, issue #92), a long
episode's novelty window grows to `stagnation_window_size * ceil(episode_len /
window_scale_steps)`, capped at `max_window`. `dump_state` correctly writes every
flag in that scaled window plus the scaler position (`counts`).

`load_state` throws most of them away. It slices the restored flag list to the
**base** window size before rebuilding the deque:

```python
flags = [bool(b) for b in raw.get("flags", [])][-self.window_size :]
w.flags = deque(
    flags,
    maxlen=self.window_size if self._scale_steps <= 0 else None,
)
w.novel_in_window = sum(flags)
```

The `maxlen` on the next line is already conditional, and the comment above it states
the reason in so many words:

> With auto-scaling enabled (issue #92) the deque is deliberately UNCAPPED:
> observe() trims at the current effective target every step, and a base-sized
> maxlen would silently drop flags and blind the detector once the target grows
> past it.

The `[-self.window_size:]` slice drops exactly those flags anyway, one line earlier,
so the uncapped deque is handed only `window_size` of them. Removing the base-sized
`maxlen` cannot help while the base-sized slice remains.

`observe` gates every evaluation on `len(w.flags) >= target`:

```python
if len(w.flags) >= target and (w.novel_in_window / target < self.min_novelty):
```

so after a restore the detector cannot reach its fire condition at all until the
window refills — exactly `target - window_size` steps of silence. If the episode
ends inside that gap, the stagnation is never reported.

**Expected:** restoring a snapshot preserves the novelty evidence the snapshot
contains, clamped to the effective window implied by the restored scaler position, so
a restarted monitor reports stagnation at (or very near) the step an uninterrupted one
does. That is the issue #91 contract, and the issue #149 clamp exists only to keep the
deque bounded — not to shrink it below the live target.

**Actual:** the restored window holds `stagnation_window_size` flags regardless of how
large the live window was. The detector is blind for `target - window_size` steps, and
a stagnation that a never-restarted monitor reports is silently missed.

Scaling off (`window_scale_steps == 0`, the default) is unaffected: `target` is always
`window_size`, so the slice is a no-op and the issue #149 clamp behaves as intended.
This only bites the horizon-scale configuration the scaling knob exists for, which is
also the configuration most likely to snapshot/restore in the first place.

```shell
# repro.py
import json

from snagline.detectors.stagnation import StagnationDetector

CFG = Config(stagnation_enabled=True, window_scale_steps=1000, max_window=512)


def ev(i, sig):
    return StepEvent(
        step_id=f"s{i}",
        episode_id="ep1",
        timestamp=float(i),
        action_type="tool_call",
        action_signature=sig,
        tool_name="t",
    )


# 2500 distinct actions, then the agent stops discovering anything new.
EVENTS = [ev(i, f"novel-{i}") for i in range(2500)]
EVENTS += [ev(2500 + j, f"novel-{j % 10}") for j in range(160)]

RESTART_AT = 2600  # snapshot written here; process restarts


def run(restart):
    det = StagnationDetector(config=CFG)
    fired = []
    for i, e in enumerate(EVENTS):
        if i == RESTART_AT and restart:
            dumped = json.loads(json.dumps(det.dump_state()))
            live = len(det._windows["ep1"].flags)
            det = StagnationDetector(config=CFG)
            det.load_state(dumped)
            print(f"  restart: novelty flags {live} -> {len(det._windows['ep1'].flags)}")
        if det.observe(e) is not None:
            fired.append(i)
    return fired


print("no restart:  ", run(False))
print("with restart:")
print("  fired:     ", run(True))
```

```shell
python repro.py
```

### Relevant log output

```shell
no restart:   [2643]
with restart:
  restart: novelty flags 150 -> 50
  fired:      []
```

The uninterrupted detector reports the stagnation at step 2643. The detector restored
from a snapshot taken 43 steps earlier drops 100 of its 150 novelty flags and never
reports it: the episode's remaining 60 steps are fewer than the 100 it needs to refill
the scaled window.

Moving the restart point later shows the same gap as a pure delay rather than a miss
(episode extended to 2900 steps so the fire still lands):

```shell
RESTART_AT=2560 -> no restart [2643], with restart [2660]
RESTART_AT=2600 -> no restart [2643], with restart [2700]
RESTART_AT=2620 -> no restart [2643], with restart [2720]
RESTART_AT=2640 -> no restart [2643], with restart [2740]
```

The delay is `target - window_size == 150 - 50 == 100` steps in every case, i.e. it is
a deterministic function of how far the window had scaled, not a timing artifact.

`tests/test_detector_snapshots.py::test_stagnation_restore_clamps_flags_to_current_window_size`
covers this code path but constructs the detector with `StagnationDetector(window_size=4,
...)` and no `Config`, so `_scale_steps == 0` and the scaled case is never exercised.
`tests/test_window_scaling.py` exercises scaling but never snapshots.

abyyxhek:
Fixes #218

Summary of Changes

`StagnationDetector.load_state` clamped the restored novelty flags to the *base*
`stagnation_window_size`:

```python
flags = [bool(b) for b in raw.get("flags", [])][-self.window_size :]
w.flags = deque(flags, maxlen=self.window_size if self._scale_steps <= 0 else None)
```

The `maxlen` on the second line is already conditional, and its comment says why —
with auto-scaling on (#92) "a base-sized maxlen would silently drop flags and blind
the detector once the target grows past it". The base-sized *slice* one line above
dropped exactly those flags anyway, so the deliberately uncapped deque was only ever
handed `window_size` of them.

`observe()` gates every evaluation on `len(w.flags) >= target`, so a restored episode
whose window had scaled could not reach its fire condition until the window refilled:
`target - window_size` steps of silence after each restore.

- `src/snagline/detectors/stagnation.py`: read `state["counts"]` before rebuilding the
  windows, then clamp each restored flag list to `effective_window_size()` for that
  episode's restored scaler position instead of to the base.
- `src/snagline/detectors/stagnation.py`: comment rewritten to say what the clamp is
  for (#149 bounds the deque) and why the bound is the effective target, not the base.
- `tests/test_detector_snapshots.py`: add
  `test_stagnation_restore_keeps_the_scaled_window_and_still_fires_on_time`.

The deque stays uncapped under scaling, exactly as before: `observe()` trims at the
live target every step, and the restored length can never exceed it, since
`effective_window_size` is monotonically non-decreasing in `episode_len`.

Justification

Snapshot/restore across a process restart is the #91 contract, and window scaling is
the knob for long-horizon episodes (#92) — the same runs most likely to snapshot in
the first place. Under that configuration a restart silently degraded detection: with
`window_scale_steps=1000`, a 2660-step episode whose window had scaled to 150 flags
came back from its own snapshot holding 50, and the collapse an uninterrupted monitor
reported at step 2643 was never reported at all, because the 60 steps left in the
episode were fewer than the 100 needed to refill the window.

Moving the restart point shows the gap is deterministic, not a timing artifact
(episode extended to 2900 steps so the fire still lands):

| restart at | no restart | with restart | delay |
|--:|:--|:--|--:|
| 2560 | 2643 | 2660 | 100 |
| 2600 | 2643 | 2700 | 100 |
| 2620 | 2643 | 2720 | 100 |
| 2640 | 2643 | 2740 | 100 |

`target - window_size == 150 - 50 == 100` in every case.

Scaling off (`window_scale_steps == 0`, the default) is unaffected: the effective
window is always the base, so the slice is a no-op and the #149 clamp behaves exactly
as it did. Pre-#92 snapshots that carry no `counts` also still clamp to the base.

Kind of Contribution

Bug Fix

Benchmark (for Performance Improvements)

Not a performance change — `load_state` is off the `ingest()` hot path and the
per-step work in `observe()` is untouched. The accuracy corpus is byte-identical
before and after: `benchmarks/detection_accuracy.py --format json` diffs clean
between `master` and this branch (macro-F1 0.951, 0 healthy controls fired).

Unit Tests:

Added `test_stagnation_restore_keeps_the_scaled_window_and_still_fires_on_time` in
`tests/test_detector_snapshots.py`. It asserts both halves of the contract: the
restored window keeps its scaled length with `novel_in_window` consistent with the
kept flags, and the restarted detector reports the collapse on the *same* step as a
detector that never restarted. It then runs 60 more steps to prove the window is
still bounded by the live `effective_window_size()` afterwards, so the #149 guarantee
is not traded away. The test fails on `master` and passes on this branch.

The existing `test_stagnation_restore_clamps_flags_to_current_window_size` covers this
code path but builds the detector with explicit kwargs and no `Config`, so
`_scale_steps == 0` and the scaled case was never exercised; `tests/test_window_scaling.py`
exercises scaling but never snapshots. Both still pass unchanged.

Execution Tests:

Reproduced end to end with the standalone script in #218 — a 2660-step episode
snapshotted at step 2600, restored into a fresh `StagnationDetector`, and replayed:

```
# before
no restart:   [2643]
with restart:
  restart: novelty flags 150 -> 50
  fired:      []

# after
no restart:   [2643]
with restart:
  restart: novelty flags 150 -> 150
  fired:      [2643]
```

Local validation, Windows 11 / CPython 3.13.1, venv built exactly as `ci.yml` does
(`pip install -e . --no-deps`, `langchain-core`, `pytest`, `pytest-cov==7.1.0`,
`ruff==0.16.3`, `mypy`):

| command | result |
|:--|:--|
| `python -m pytest tests/test_detector_snapshots.py::test_stagnation_restore_keeps_the_scaled_window_and_still_fires_on_time -q` (on `master`) | 1 failed — the regression is real |
| `python -m pytest tests/test_detector_snapshots.py -q` | 12 passed |
| `python -m pytest tests/ -q --cov-fail-under=0` | 697 passed, 2 skipped, 1 failed |
| `ruff check src tests` | All checks passed |
| `ruff format --check src tests` | 129 files already formatted |
| `mypy src tests` | 1 error, pre-existing and Windows-only |
| `python benchmarks/detection_accuracy.py --fixtures benchmarks/fixtures --format table` | exit 0, 0 healthy controls fired |

Limitations / notes on the two red lines above, both pre-existing and unrelated:

- The suite failure is `tests/test_hook_bridge.py::test_watch_follow_teardown_reaps_the_child_on_a_healthy_run`, i.e. #215. It reproduces identically on unmodified `master` here (1 failed, 696 passed) and passes in isolation on both, which is the ~75%-of-runs flake documented in that issue. #216 is the fix.
- The mypy error is `tests/test_hook_bridge.py:527: Module has no attribute "SIGKILL"` — `signal.SIGKILL` does not exist on Windows. CI runs mypy on Linux only, where it does.

Neither touches `src/snagline/detectors/`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved restoration of detector state when automatic window scaling is enabled.
  * Preserved effective window sizes, novelty tracking, and stagnation detection behavior after state reloads.
  * Ensured restored monitoring windows remain properly bounded as new data is processed.

* **Tests**
  * Added regression coverage for saving and restoring scaled detector state.

### DIFF
--- a/src/snagline/detectors/stagnation.py
+++ b/src/snagline/detectors/stagnation.py
@@ -181,25 +181,36 @@ class StagnationDetector:
 
     def load_state(self, state: dict[str, Any]) -> None:
         self._windows = {}
+        # Read the scaler positions first: they decide how large the restored
+        # window is allowed to be when auto-scaling is on (issue #92).
+        self._counts = {ep: int(n) for ep, n in state.get("counts", {}).items()}
         for ep, raw in state.get("windows", {}).items():
             w = _EpisodeWindow()
-            # Clamp to the CURRENT window_size: restore is tolerant by default
-            # (matches by name, ignores config), so a snapshot taken with a
-            # larger window must not leave an overlong deque whose equality-
-            # based eviction never fires again. Mirrors LoopDetector's
-            # deque(sigs, maxlen=self.window_size) rebuild; the sliding
-            # counter is recomputed from the truncated flags so the two stay
-            # consistent. With auto-scaling enabled (issue #92) the deque is
-            # deliberately UNCAPPED: observe() trims at the current effective
-            # target every step, and a base-sized maxlen would silently drop
-            # flags and blind the detector once the target grows past it.
-            flags = [bool(b) for b in raw.get("flags", [])][-self.window_size :]
+            # Clamp to the window this detector will actually evaluate against:
+            # restore is tolerant by default (matches by name, ignores config),
+            # so a snapshot taken with a larger window must not leave an
+            # overlong deque whose eviction never fires again (issue #149). With
+            # auto-scaling that bound is the *effective* target for the restored
+            # scaler position, not the base -- clamping to the base would drop
+            # the very flags the snapshot carries and blind observe(), whose
+            # every check is gated on ``len(flags) >= target``, for
+            # ``target - window_size`` steps. The deque itself stays UNCAPPED
+            # under scaling: observe() trims at the live target every step, and
+            # a fixed maxlen would blind it again once the target grows past it.
+            target = effective_window_size(
+                self.window_size,
+                self._counts.get(str(ep), 0),
+                self._scale_steps,
+                self._max_window,
+            )
+            flags = [bool(b) for b in raw.get("flags", [])][-target:]
             w.flags = deque(
                 flags,
                 maxlen=self.window_size if self._scale_steps <= 0 else None,
             )
+            # The sliding counter is recomputed from the kept flags so the two
+            # stay consistent.
             w.novel_in_window = sum(flags)
             w.stale_windows = int(raw.get("stale_windows", 0))
             w.seen_all_time = set(raw.get("seen_all_time", []))
             self._windows[str(ep)] = w
-        self._counts = {ep: int(n) for ep, n in state.get("counts", {}).items()}
--- a/tests/test_detector_snapshots.py
+++ b/tests/test_detector_snapshots.py
@@ -14,9 +14,11 @@ from __future__ import annotations
 
 import json
 
+from snagline.config import Config
 from snagline.detectors.compaction_tripwire import CompactionTripwireDetector
 from snagline.detectors.side_effect_guard import SideEffectGuardDetector
 from snagline.detectors.stagnation import StagnationDetector
+from snagline.detectors.windowing import effective_window_size
 from snagline.events import StepEvent
 from snagline.monitor import Monitor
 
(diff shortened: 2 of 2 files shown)


Cyrax321 (approved):
Verified locally: 12/12 pass in test_detector_snapshots.py, ruff + mypy clean. Clamping restored flags to the effective window (not base) using the restored scaler position, with counter recomputed to match. Base fix for the scaled-restore chain - no textual conflict with #269, they complement. Safe to merge.

Issue #14, opened 2026-08-14
Fail-open logs full traceback on every detector/sink error (log spam under a faulty detector)


Cyrax321:
## Description

Under `fail_open=True`, `Monitor.ingest`, `_dispatch`, and `end_episode` call `logger.exception(...)` on every detector/sink error. A buggy third-party detector then produces a full traceback on every single ingested step.

## Impact

- Log volume explosion (one traceback per step, potentially millions per run).
- Obscures genuine signals and can itself become a performance problem due to traceback formatting.

## Expected behavior

Adapter/detector faults should be surfaced without flooding logs, e.g. log once per (detector, exception-type) per episode, or every Nth occurrence, plus an optional fault callback.

## Suggested fix

- Deduplicate exception logging by `(detector, type(exc))` within an episode.
- Optionally expose a `on_fault` hook so hosts can decide how to handle repeated adapter errors.

## Additional context

Fail-open is correct; the logging behavior around it needs rate-limiting.

Cyrax321:
Resolved: fail-open faults are logged exactly once per (detector, message) key, not per step. Covered by tests/test_issue_regressions.py::test_monitor_logs_fault_once_per_key. PR #23

PR #230, opened 2026-08-31
fix(detectors): isolate base-detector faults in MLOrchestrator (Fixes #227)

### LINKED ISSUE #227: [detector] MLOrchestrator delegates to base detectors with no fail-open isolation: one raising detector silently disables the whole ensemble
## Summary

`MLOrchestrator` is the only detector that owns other detectors, and it delegates to them with **no fail-open isolation at all**. Every one of its five delegation loops (`observe`, `reset`, `finalize`, `dump_state`, `load_state`) calls straight into each base detector, so the first one that raises aborts the loop and every remaining base detector is skipped.

This breaks a guarantee the README states explicitly (README.md:359):

> | **Fail-open guarantee** | Detector exceptions don't propagate. Sink exceptions don't propagate. **One bad detector doesn't block others.** Both `fail_open=True` and `fail_open=False` are exercised. |

That holds for `Monitor.ingest`, which guards each detector individually (`monitor.py:350-361`). It does **not** hold once `Config.ml_ensemble_enabled=True`, because `Monitor.default()` collapses every base detector into a single `MLOrchestrator` slot (`monitor.py:1024`). From the Monitor's point of view there is now exactly one detector, so its per-detector guard has nothing left to isolate: one bad base detector blocks all the others, and the Monitor's fail-open silently swallows the evidence.

Supporting signal that this was meant to be the orchestrator's job: `EsnCusumDetector.observe` — the one base detector shipped *specifically* for the ensemble — wraps its own body in `try/except` and comments "Never raises: an internal error is logged once per occurrence by the Monitor and here as well" (`ml/esn_ensemble.py:160-170`). It defends itself because nothing above it does. Every hand-written or third-party base detector a user adds gets no such protection.

## Reproduction

```python
from snagline.detectors.ml_ensemble import MLOrchestrator


class Boom:
    name = "boom"
    def observe(self, event): raise RuntimeError("base detector exploded")
    def reset(self, episode_id): raise RuntimeError("reset exploded")
    def finalize(self, episode_id): raise RuntimeError("finalize exploded")
    def dump_state(self): raise RuntimeError("dump exploded")
    def load_state(self, state): raise RuntimeError("load exploded")


class Good:
    name = "good"
    def __init__(self):
        self.seen = 0
        self.reset_called = False
    def observe(self, event):
        self.seen += 1
        return FailureRisk(event.episode_id, event.step_id, 0.9, "good", "d", event.timestamp)
    def reset(self, episode_id): self.reset_called = True
    def dump_state(self): return {"seen": self.seen}
    def load_state(self, state): self.seen = state["seen"]
    def finalize(self, episode_id): return None


ev = StepEvent(
    step_id="0", episode_id="ep", timestamp=0.0, action_type="tool_call",
    action_signature=make_signature("tool_call", "t", "a"),
    tool_name="t", latency_ms=10.0,
)

cfg = Config(ml_ensemble_enabled=True, ml_ensemble_score_threshold=0.5)
good = Good()
orch = MLOrchestrator([Boom(), good], config=cfg)

orch.observe(ev)          # RuntimeError: base detector exploded
print(good.seen)          # 0 -- the healthy detector never saw the step

risks = []
m = Monitor([MLOrchestrator([Boom(), Good()], config=cfg)], [], fail_open=True)
m.on_risk = risks.append
m.ingest(ev)
print(len(risks))         # 0 -- expected 1; the ensemble is silently dead
```

Observed output on `master` (22faeae):

```
snagline: detector ml_ensemble raised; ignoring (fail-open)
observe: RAISED RuntimeError: base detector exploded
  good.seen = 0  (expected 1 if isolated)
monitor risks = 0  (expected 1 if isolated)
reset: RAISED RuntimeError
  good.reset_called = False  (expected True)
finalize: RAISED RuntimeError
  good.finalized = False  (expected True)
dump_state: RAISED RuntimeError
load_state: RAISED RuntimeError
  good.seen = 0  (expected 42)
```

All five paths fail.

## Impact per path

- **`observe` (`ml_ensemble.py:44-46`)** — total detection blackout. One consistently raising base detector reduces the ensemble to `None` on every step for the whole process lifetime. Because the Monitor's fault log is deduplicated once per distinct message (issue #14), the operator sees a single `detector ml_ensemble raised` line and then silence: SNAGLINE reports zero risks and looks healthy while detecting nothing. Detectors that keep sliding windows also lose that step permanently, so their windows develop gaps even for steps that never reached them.
- **`reset` (`ml_ensemble.py:75-76`)** — per-episode state leak. `Monitor.end_episode` and `_evict_episode` guard `detector.reset` at the top level, so the exception is logged and swallowed, but every base detector *after* the raiser keeps its episode state forever. That is unbounded memory growth for a long-lived Monitor and it defeats the retention cap added in #184; a reused episode id also starts with a dirty window.
- **`finalize` (`ml_ensemble.py:83-88`)** — end-of-episode judgments vanish. `SilentAbortDetector` can only decide at `finalize` time (#86), so a raiser ordered before it means `silent_abort` never fires for any episode.
- **`load_state` (`ml_ensemble.py:101-108`)** — half-restored ensemble. Base detectors before the raiser are restored, those after silently are not, leaving a mixed old/new state that no caller can detect.
- **`dump_state` (`ml_ensemble.py:91-99`)** — one base detector aborts the whole `Monitor.snapshot()`.

Also, `reset` calls `det.reset(episode_id)` unconditionally while `dump_state`/`load_state` probe with `getattr`, so a base detector that omits `reset` raises `AttributeError` rather than being skipped.

## Suggested fix

Wrap each per-base-detector call in the orchestrator the way `Monitor.ingest` already does: log fault-once, count it, continue to the next base detector. Isolation belongs in the orchestrator because it is the layer that removed the Monitor's isolation in the first place.

Happy to send a PR.

Adhi1-2:
Fixes #227

## Problem

`MLOrchestrator` is the only detector that owns other detectors, and all five of its delegation loops called straight into each base detector with no guard. The first base detector to raise aborted the loop and every remaining one was skipped.

`Monitor.ingest` guards every detector slot individually, so this is invisible with the default composition. But `Config.ml_ensemble_enabled=True` makes `Monitor.default()` collapse every base detector into a single `MLOrchestrator` slot (`monitor.py:1024`), and the Monitor's per-detector guard then has nothing left to isolate: one bad base detector blocks all the others, and the Monitor's fail-open swallows the evidence. That breaks the guarantee stated at README.md:359 — "One bad detector doesn't block others."

The most visible consequence is a total, silent detection blackout: one consistently raising base detector reduces the ensemble to `None` on every step, and because the Monitor deduplicates its fault log (issue #14) the operator sees one line and then permanent silence.

## Fix

Wrap each per-base call the way `Monitor.ingest` already does — log the fault once, continue to the next detector — with a per-`(detector, phase)` fault-once set mirroring the issue #14 discipline so a persistent fault cannot spam the log on every step.

| path | before | after |
| --- | --- | --- |
| `observe` | raiser blanks the ensemble for the step | remaining detectors still score; noisy-OR runs on what survived |
| `reset` | siblings keep episode state forever, defeating the #184 cap | sibling state is still freed |
| `finalize` | a raiser ordered before `SilentAbortDetector` means `silent_abort` never fires | later completion checks still run |
| `dump_state` | one base detector aborts the whole `Monitor.snapshot()` | that detector's slot is dropped, the rest is written |
| `load_state` | siblings after the raiser silently stay unrestored | each entry is applied independently |

`reset` also now probes with `getattr` like the other loops, so a base detector that omits `reset` is skipped instead of raising `AttributeError`.

No behavior change on the healthy path: with no base detector raising, every loop does exactly what it did before.

## Tests

New `tests/test_ml_ensemble_fail_open.py`, 10 cases: isolation on all five paths, isolation regardless of the raiser's position in the base list, the end-to-end `Monitor.ingest` case (the ensemble degrades instead of going silent), fault-once log discipline over 25 steps, the `getattr` probe for a base detector without `reset`, and a `Monitor.snapshot()` → `restore()` round trip through a faulty base detector.

All 10 fail on `master` and pass with the fix.

## Verification

```
707 passed, 2 skipped          (baseline on this branch point: 697 passed, 2 skipped)
Total coverage: 87.95%         (gate: 80%)
ruff check src tests           All checks passed!
ruff format --check src tests  130 files already formatted
mypy src                       Success: no issues found in 55 source files
```

### DIFF
--- a/src/snagline/detectors/ml_ensemble.py
+++ b/src/snagline/detectors/ml_ensemble.py
@@ -14,16 +14,26 @@ upgrade path; it is never imported here, so the zero-dep default holds.
 When ``Config.ml_ensemble_enabled`` is set, ``Monitor.default()`` wraps the
 base detectors in a single ``MLOrchestrator`` instead of exposing them
 individually, so there is no double counting.
+
+Fail-open isolation is this class's own responsibility. ``Monitor.ingest``
+guards every detector slot individually, but wrapping the base detectors
+collapses them into one slot, so the Monitor's guard has nothing left to
+isolate: without the per-base guards below, one raising detector would block
+all the others and the Monitor would swallow the evidence. Every delegation
+loop here therefore logs the fault once and moves on to the next detector.
 """
 
 from __future__ import annotations
 
+import logging
 from collections.abc import Callable, Sequence
 from typing import Any
 
 from snagline.config import Config
 from snagline.events import StepEvent
 from snagline.risk import FailureRisk
+
+logger = logging.getLogger("snagline")
 
 
 class MLOrchestrator:
@@ -38,11 +48,33 @@ class MLOrchestrator:
         self._base = list(detectors)
         self._cfg = config or Config()
         self._model = model
+        # A faulty base detector must not spam the logs on every step; mirror
+        # Monitor._log_fault_once and report each distinct fault exactly once.
+        self._fault_logged: set[str] = set()
+
+    @staticmethod
+    def _base_name(det: Any) -> str:
+        return str(getattr(det, "name", type(det).__name__))
+
+    def _log_fault_once(self, det: Any, phase: str) -> None:
+        key = f"{self._base_name(det)}:{phase}"
+        if key in self._fault_logged:
+            return
+        self._fault_logged.add(key)
+        logger.warning(
+            "snagline: ml_ensemble base detector %s %s raised; ignoring (fail-open)",
+            self._base_name(det),
+            phase,
+        )
 
     def observe(self, event: StepEvent) -> FailureRisk | None:
         scores: list[float] = []
         for det in self._base:
-            risk = det.observe(event)
+            try:
+                risk = det.observe(event)
+            except Exception:
+                self._log_fault_once(det, "observe")
+                continue
             if risk is not None and 0.0 < risk.score <= 1.0:
                 scores.append(risk.score)
         if not scores:
@@ -73,7 +105,16 @@ class MLOrchestrator:
 
     def reset(self, episode_id: str) -> None:
         for det in self._base:
-            det.reset(episode_id)
+            reset = getattr(det, "reset", None)
+            if not callable(reset):
+                continue
+            try:
+                reset(episode_id)
+            except Exception:
+                # A leaked reset strands that episode's state forever and
+                # defeats the retention cap (issue #184); it must never also
+                # strand every sibling's state.
+                self._log_fault_once(det, "reset")
 
     def finalize(self, episode_id: str) -> FailureRisk | None:
         # Passthrough for EpisodeFinalizer bases (issue #86): Monitor.end_episode
@@ -83,7 +124,11 @@ class MLOrchestrator:
         for det in self._base:
             finalize = getattr(det, "finalize", None)
             if callable(finalize):
-                risk = finalize(episode_id)
+                try:
+                    risk = finalize(episode_id)
+                except Exception:
+                    self._log_fault_once(det, "finalize")
+                    continue
                 if risk is not None:
                     return risk
         return None
@@ -93,9 +138,17 @@ class MLOrchestrator:
         merged: dict[str, Any] = {}
         for det in self._base:
             dump = getattr(det, "dump_state", None)
-            state = dump() if callable(dump) else None
+            if not callable(dump):
+                continue
+            try:
+                state = dump()
+            except Exception:
+                # Losing one base detector's state is recoverable (it restarts
+                # cold); losing the whole snapshot is not.
+                self._log_fault_once(det, "dump_state")
+                continue
             if state is not None:
-                merged[getattr(det, "name", type(det).__name__)] = state
+                merged[self._base_name(det)] = state
         return merged or None
 
     def load_state(self, state: dict[str, Any]) -> None:
@@ -103,6 +156,11 @@ class MLOrchestrator:
             load = getattr(det, "load_state", None)
             if not callable(load):
                 continue
-            sub = state.get(getattr(det, "name", type(det).__name__))
+            sub = state.get(self._base_name(det))
             if sub is not None:
-                load(sub)
+                try:
+                    load(sub)
+                except Exception:
+                    # Skipping one entry leaves that detector cold rather than
+                    # leaving every later sibling silently unrestored.
+                    self._log_fault_once(det, "load_state")
--- a/tests/test_ml_ensemble_fail_open.py
+++ b/tests/test_ml_ensemble_fail_open.py
@@ -1,0 +1,220 @@
+"""Regression tests for issue #227: MLOrchestrator must isolate base detectors.
+
+``Monitor.ingest`` guards every detector slot individually, but enabling
+``ml_ensemble`` collapses the base detectors into a single slot, so the
+Monitor's guard has nothing left to isolate. The orchestrator itself must
+therefore honor the README's "one bad detector doesn't block others"
+guarantee on every path it delegates: observe, reset, finalize, dump_state
+and load_state.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from snagline.config import Config
+from snagline.detectors.ml_ensemble import MLOrchestrator
+from snagline.events import StepEvent, make_signature
+from snagline.monitor import Monitor
+from snagline.risk import FailureRisk
+
+
+class _Boom:
+    """A base detector that raises on every delegated call."""
+
+    name = "boom"
+
+    def observe(self, event: StepEvent) -> FailureRisk | None:
+        raise RuntimeError("observe exploded")
+
+    def reset(self, episode_id: str) -> None:
+        raise RuntimeError("reset exploded")
+
+    def finalize(self, episode_id: str) -> FailureRisk | None:
+        raise RuntimeError("finalize exploded")
+
+    def dump_state(self) -> dict[str, Any]:
+        raise RuntimeError("dump_state exploded")
+
+    def load_state(self, state: dict[str, Any]) -> None:
+        raise RuntimeError("load_state exploded")
+
+
+class _Healthy:
+    """A base detector that records every delegated call it receives."""
+
+    name = "healthy"
+
+    def __init__(self, score: float = 0.9) -> None:
+        self._score = score
+        self.seen = 0
+        self.resets = 0
+        self.finalized = 0
+        self.loaded: dict[str, Any] | None = None
+
+    def observe(self, event: StepEvent) -> FailureRisk | None:
+        self.seen += 1
+        return FailureRisk(
+            event.episode_id,
+            event.step_id,
+            self._score,
+            "healthy",  # type: ignore[arg-type]
+            "healthy",
+            event.timestamp,
+        )
+
+    def reset(self, episode_id: str) -> None:
+        self.resets += 1
+
+    def finalize(self, episode_id: str) -> FailureRisk | None:
+        self.finalized += 1
+        return FailureRisk(
+            episode_id,
+            "final",
+            self._score,
+            "healthy",  # type: ignore[arg-type]
+            "finalized",
+            0.0,
+        )
+
+    def dump_state(self) -> dict[str, Any]:
+        return {"seen": self.seen}
+
+    def load_state(self, state: dict[str, Any]) -> None:
+        self.loaded = state
+
+
+class _NoReset:
+    """A minimal base detector exposing only ``observe``."""
+
+    name = "no_reset"
+
+    def observe(self, event: StepEvent) -> FailureRisk | None:
+        return None
+
+
+def _cfg() -> Config:
+    return Config(ml_ensemble_enabled=True, ml_ensemble_score_threshold=0.5)
+
+
+def _ev(step: str = "0", episode: str = "ep") -> StepEvent:
+    return StepEvent(
+        step_id=step,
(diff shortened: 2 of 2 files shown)


yuki-fuyutsuki:
# PR 230 - In-depth Review: `MLOrchestrator` fail-open isolation

**Scope:** `src/snagline/detectors/ml_ensemble.py:1-166` + `tests/test_ml_ensemble_fail_open.py:1-220`, diff `22faeae..2ba655f` (+285 / -7). Fixes #227.

## 1. Summary / Correctness of Problem Statement

`Monitor.default()` with `Config.ml_ensemble_enabled=True` collapses `base: list[Detector]` into a single `MLOrchestrator` slot at `src/snagline/monitor.py:1024`. `Monitor.ingest:350-361` guards per-slot, so with `ml_ensemble` the per-base isolation is lost. One `raise` in the 5 delegation loops aborts the loop, siblings skipped, ensemble `observe` returns `None`, and with `Monitor._log_fault_once:608` deduplication the operator sees 1 line then silence. The PR correctly identifies this violates `README.md:359`.

Fix is canonical: wrap each per-base call with `try/except Exception: log-once + continue` mirroring `Monitor.ingest` and `Monitor.end_episode`.

**Verdict:** Problem is real, fix architecture is correct, healthy-path has no behavioral change (verified: with no raise each loop does identical append/combine).

## 2. What Was Done Well

*   **Coverage of all 5 delegation paths** - `observe:70`, `reset:106`, `finalize:119`, `dump_state:136`, `load_state:154` all guarded. Previously only `observe` partially, `reset` not even `getattr`-probed.
*   **Fault-once discipline** - `src/snagline/detectors/ml_ensemble.py:52-68` mirrors `Monitor._fault_logged:232`/`_log_fault_once:608` with per-`(detector,phase)` key, preventing per-step log spam for a persistently faulty detector. Test `test_a_faulty_base_detector_is_logged_once_not_per_step:150` verifies 25 steps => 1 record at `WARNING`.
*   **`reset` hardening** - `src/snagline/detectors/ml_ensemble.py:108-109` `getattr` + `callable` probe prevents `AttributeError` for detectors without `reset` (`_NoReset`). Matches `finalize`/`dump_state`/`load_state` style.
*   **State isolation** - `dump_state` losing one entry is recoverable, losing whole `Monitor.snapshot()` is not; same for `load_state`. Comments at `src/snagline/detectors/ml_ensemble.py:146-147`/`164-165` state intent.
*   **Tests are precise** - 10 cases cover ordering invariance (`test_observe_isolation_holds_for_every_base_position:122`), end-to-end `Monitor.ingest` degradation (`test_monitor_still_emits_when_one_base_detector_raises:140`), all 5 isolates, `getattr` probe, and `snapshot()->restore()` round-trip `206`. All fail on `master`, pass with patch (validated manually via `PYTHONPATH=src python3 -c` repro).

## 3. Findings

### 🔴 Major: Fault-dedup + state keys collide on `name`

`src/snagline/detectors/ml_ensemble.py:56-60`

```python
def _base_name(det): return str(getattr(det, "name", type(det).__name__))
key = f"{self._base_name(det)}:{phase}"
```

and `src/snagline/detectors/ml_ensemble.py:151` / `159` `merged[self._base_name(det)] = state` / `state.get(self._base_name(det))`

Two distinct instances with same `name` (e.g. two unnamed instances of same class, or two `Healthy` clones) share one `key`. Repro:

```python
orch = MLOrchestrator([A(name="same"), B(name="same")])
orch.dump_state() # -> {'same': {'b':2}}  # A's state clobbered
orch = MLOrchestrator([Boom(name="same"), Boom(name="same")])
orch.observe(ev)  # _fault_logged == {'same:observe'}  # second fault swallowed, never logged
```

Default composition uses distinct names (`loop`, `error_cascade`, `latency_anomaly`, etc.) so unlikely today, but is a public API (`detectors: Sequence[Any]`) and collides silently. `Monitor` avoids this via `src/snagline/monitor.py:105` `_detector_key(index, detector) = f"{index}:{name}"`. `MLOrchestrator` should adopt positional keying or `id(det)`:

Proposal:

```diff
-self._fault_logged: set[str] = set()
+self._fault_logged: set[tuple[int,str]] = set()  # or set[tuple[int,str]]
-key = f"{self._base_name(det)}:{phase}"
+key = (id(det), phase) # display still uses _base_name(det)
```

For `dump_state`/`load_state` the collision is **data loss**, not just observability. Options: key on `_detector_key(i, det)` like `Monitor.snapshot_dict:745`, or keep `merged` as `dict[str,Any]` but key on `f"{i}:{_base_name(det)}"` and make `load_state` search by same scheme (Monitor already has fallback loop `src/snagline/monitor.py:802-814`). If keeping name-keyed file format, at least document the distinct-name requirement.

Flagged independently by CodeRabbit `src/snagline/detectors/ml_ensemble.py:60` and Gitar bot - still valid post-fix.

### 🟡 Medium: `MLOrchestrator._fault_logged` is not thread-safe

`Monitor._log_fault_once:612` guards `self._fault_logged` with `self._fault_lock`. `MLOrchestrator._log_fault_once:60-68` uses a plain `set` with no lock. `Monitor.ingest:345` holds `episode_lock` per `episode_id`, but concurrent ingests for *different* episodes run concurrently and both call `orchestrator.observe`. Racing `key in set` / `set.add` can corrupt `set` (CPython GIL makes it rarely crash, but `check-then-add` is not atomic, duplicate log or lost update possible). Add `self._fault_lock = threading.Lock()` mirroring `Monitor:231`.

Repro: 5 threads x 100 `observe` with a `Boom` did not crash in local test but exposed non-atomicity risk; under `pytest --dist` this will interleave.

### 🟡 Medium: `fail_open=False` is silently ignored

`Monitor.ingest:359-360`, `end_episode:699`, `_dispatch:470` re-raise when `not self._fail_open`. `MLOrchestrator` stores `self._cfg = config or Config()` at `src/snagline/detectors/ml_ensemble.py:49` (which has `fail_open: bool = True` `src/snagline/config.py:367`) but never consults it; faults are always swallowed. With `Config(fail_open=False)` a strict test harness expects propagation for debugging, but `Monitor([MLOrchestrator([Boom()])], ..., fail_open=False).ingest(ev)` would stay silent. Either respect `self._cfg.fail_open` (or accept `fail_open` param) and `raise` after logging, or document divergence as intentional. Current tests don't cover `fail_open=False` inside the ensemble.

### 🟢 Minor / Nits

*   **Log level mismatch** - `MLOrchestrator._log_fault_once:64` uses `logger.warning` while `Monitor._log_fault_once:616` uses `logger.error`. Same fail-open event should have consistent severity; consider `warning` -> `error` or vice versa.
*   **Type hint for `_fault_logged`** - declared `set[str]` but proposed `tuple[int,str]` fix requires `set[tuple[int,str]]`.
*   **`_model` exception not isolated?** `_combine:94-97` calls `self._model(scores)` outside the per-detector `try`. A user-supplied `model` raising would abort the ensemble entirely (and propagate to `Monitor.ingest` as single detector failure, then correctly guarded at Monitor level, but loses isolation granularity). Worth wrapping or documenting as caller's responsibility.
*   **`load_state` fallback** - `Monitor.restore_dict:802-814` implements name fallback for reordered detectors. `MLOrchestrator.load_state:159` is strict `state.get(_base_name)`. If `dump_state` key scheme changes to positional, keep parity.
*   **Coverage of multi-fault case** missing - tests always use 1 `Boom` + 1 `Healthy`. No test with 2 distinct failing detectors asserts both faults are logged once each (would fail today due to collision above, good regression for fix).

## 4. Tests Assessment

`tests/test_ml_ensemble_fail_open.py:1-220` is high quality: short helpers `_Boom`/`_Healthy`/`_NoReset`, deterministic `_ev`/`_cfg`, each phase isolated, position invariance, deduplication over 25 steps, `Monitor.snapshot` round-trip (which exercises `dump_state`+`load_state` through `Monitor` path). Gaps noted above (same-name collision, `fail_open=False`, `threading`, multi-boom logging).

Verification reported in PR body still holds post-review: `ruff check`/`format` pass, `mypy` passes (55 files). Full suite expectation 707 passed (vs 697 baseline) consistent with +10 new tests.

## 5. Recommendations

1.  Fix `src/snagline/detectors/ml_ensemble.py:60` fault key to ` (id(det), phase)` or `(index, _base_name, phase)` and update `dump_state:151`/`load_state:159` to positional keys `f"{i}:{_base_name(det)}"` (align with `Monitor._detector_key:105`). This resolves both observability and data-loss collisions.
2.  Add `self._fault_lock = threading.Lock()` and guard `src/snagline/detectors/ml_ensemble.py:61-68` like `Monitor:612`.
3.  Decide `fail_open` contract for `MLOrchestrator` - either propagate on `not self._cfg.fail_open` after logging, or add code comment clarifying intentional always-fail-open.
4.  Add 2-3 follow-up tests: two `Boom`s with same `name` logging twice, two detectors same `name` with distinct `dump_state` not clobbering, and `fail_open=False` propagation.

No blocking defect for default `Monitor.default()` usage (distinct names, single-threaded ingest per episode), but the name-collision state loss is worth a follow-up before external adoption.


[github run](/Cyrax321/SNAGLINE/actions/runs/33604644116)

Cyrax321 (approved):
Verified locally: 10/10 pass in test_ml_ensemble_fail_open.py, ruff + mypy clean. Every delegation loop (observe/reset/finalize/dump/load) is now guarded with log-once-and-continue, restoring the one-bad-detector guarantee that collapsing bases into one Monitor slot had removed. The duck-typed getattr guards are a nice touch. Safe to merge.

Issue #231, opened 2026-08-31
[core] Every detector's dump_state walks its per-episode dict with a Python-level comprehension, so Monitor.snapshot() under concurrent ingest raises RuntimeError


Adhi1-2:
## Summary

Every shipped detector serializes its per-episode state with a **Python-level comprehension** over its own dict:

```python
# detectors/loop.py:374
"windows": {ep: list(w) for ep, w in self._windows.items()},
```

A Python-level comprehension yields between bytecodes, so the interpreter can switch threads mid-walk. `Monitor.snapshot_dict()` calls every detector's `dump_state()` with **no lock held** (`monitor.py:759-761`), and `ingest()` inserts a new episode's entry under only that episode's lock. Snapshotting from a sidecar/cron thread while the agent keeps ingesting therefore raises out of a public, non-fail-open API:

```
Traceback (most recent call last):
  File "repro.py", line 22, in snap
    m.snapshot_dict()
  File ".../snagline/monitor.py", line 765, in snapshot_dict
    detectors[_detector_key(i, detector)] = dump() if callable(dump) else None
  File ".../snagline/detectors/loop.py", line 374, in dump_state
    "windows": {ep: list(w) for ep, w in self._windows.items()},
RuntimeError: dictionary changed size during iteration
```

Reproduced on `master` (22faeae) with `Monitor.default()` and four ingest threads, inside a second.

This is the same *class* of bug as #228 (`Monitor._clocks`) but a distinct defect: different modules, and the trigger is the Python-level comprehension rather than a missing Monitor lock. `LoopDetector` alone has three such dicts. The fix for #228 does not help here — I found this precisely because the #228 regression test still failed against `Monitor.default()` after `_clocks` was locked.

## Why it matters

`snapshot_dict` / `snapshot` have no fail-open guard, so the `RuntimeError` propagates into the host. A host following the documented restart-survivability pattern from #91 (`monitor.snapshot(path)` on a timer) gets an exception from SNAGLINE *and* no snapshot — it loses exactly the state it was trying to make restart-survivable. It also contradicts README.md:172 ("Thread-safe | Per-instance `threading.Lock` supports concurrent multi-episode monitoring").

## Reproduction

```python
import itertools
import threading

from snagline.config import Config
from snagline.events import StepEvent, make_signature
from snagline.monitor import Monitor

m = Monitor.default(config=Config(), sinks=[])
stop, errors = threading.Event(), []


def ev(ep, i):
[...]

PR #232, opened 2026-08-31
fix(core): lock the cross-episode dicts that snapshot and teardown walk (Fixes #228, #229)

### LINKED ISSUE #228: [core] Monitor.snapshot_dict() iterates _clocks without a lock: a snapshot concurrent with ingest raises RuntimeError out of the public API
## Summary

`Monitor.snapshot_dict()` iterates `self._clocks` with no lock held (`monitor.py:752-760`), while `Monitor.ingest()` inserts into that same dict under only the **per-episode** lock (`_advance_clock`, `monitor.py:397`). The per-episode lock serializes two ingests *for the same episode*; it does not serialize an ingest for episode A against anything touching the shared `_clocks` dict. Snapshotting from a sidecar/cron thread while the agent keeps ingesting therefore raises out of a public API:

```
RuntimeError: dictionary keys changed during iteration
```

This contradicts README.md:172:

> | **Thread-safe** | Per-instance `threading.Lock` supports concurrent multi-episode monitoring. |

and the in-code claim at `monitor.py:346-348` that the time-axis head runs "under the episode lock so concurrent ingests serialize exactly like detector state does" — true only within one episode id.

Unlike a detector or sink fault, this one is **not** fail-open: `snapshot_dict` / `snapshot` have no guard, so the `RuntimeError` propagates into the host that called them. A host following the documented restart-survivability pattern (`monitor.snapshot(path)` on a timer, #91) gets an exception from SNAGLINE and, worse, no snapshot — the very state it was trying to make restart-survivable is the thing it loses.

## Reproduction

```python
import threading, itertools

cfg = Config(idle_warn_seconds=3600.0)   # any horizon knob activates the time axis
m = Monitor.default(config=cfg, sinks=[])
stop, errors = threading.Event(), []


def ev(ep, i):
    return StepEvent(step_id=str(i), episode_id=ep, timestamp=float(i),
                     action_type="tool_call",
                     action_signature=make_signature("tool_call", "t", str(i)),
                     tool_name="t", latency_ms=10.0)


def ingester(tag):
    for n in itertools.count():
        if stop.is_set():
            return
        m.ingest(ev(f"{tag}-{n}", n))


def snapshotter():
    while not stop.is_set():
        try:
            m.snapshot_dict()
        except Exception as e:
            errors.append(e)
            stop.set()


ts = [threading.Thread(target=ingester, args=(f"w{i}",), daemon=True) for i in range(4)]
ts.append(threading.Thread(target=snapshotter, daemon=True))
for t in ts:
    t.start()
stop.wait(timeout=8.0)
stop.set()
print(errors[:1])
```

On `master` (22faeae) this reproduces within a second or so:

```
CRASH: RuntimeError: dictionary keys changed during iteration
```

Note the time axis only has to be *enabled* (`SNAGLINE_IDLE_WARN_SECONDS` or `SNAGLINE_MAX_EPISODE_WALL_SECONDS`); with both unset `_advance_clock` returns early and `_clocks` stays empty, which is why the existing suite never sees this.

## Scope

`_clocks` is the Monitor's own state and the clearest case, but the same lock gap covers `snapshot_dict`'s detector loop (`monitor.py:745-747`): it calls every detector's `dump_state()` with no lock either. Any detector that keeps a dict *not* keyed purely by episode id is exposed the same way — `LatencyAnomalyDetector.dump_state` does `sorted(self._states.items())` over a dict keyed by `(episode_id, tool_name)`, so a concurrent first-sight of a new tool in another episode inserts mid-iteration. (`LatencyAnomalyDetector.reset` has an independent instance of this race; filed separately.)

Also worth noting the Redis backend does not help: `RedisStateBackend.episode_lock` distributes the *episode* lock, but `_clocks` and the detector dicts remain process-local, so the race is identical.

## Suggested fix

Give `_clocks` a dedicated `threading.Lock` (like `_live_lock` / `_metrics_lock` already do for the other shared structures) and hold it for the dict-level operations only:

- `_advance_clock`: the `get` / insert
- `end_episode` / `_evict_episode`: the `pop`
- `snapshot_dict`: build the `time_axis` payload from a snapshot of the values taken under the lock
- `restore_dict`: the repopulation

Mutating the fields *inside* an `_EpisodeClock` stays under the per-episode lock as today, so the hot path takes the new lock only on episode first-sight and teardown — `O(1)` per step is preserved.

Happy to send a PR.

### LINKED ISSUE #229: [detector] LatencyAnomalyDetector.reset() iterates its cross-episode _states dict: teardown races concurrent ingest, and the swallowed RuntimeError leaks state past the #184 cap
## Summary

`LatencyAnomalyDetector` is the one shipped detector whose state dict is **not** keyed purely by episode id — it is keyed by `(episode_id, tool_name)` (`latency_anomaly.py:180`). Consequently `reset` cannot just `pop`; it scans:

```python
def reset(self, episode_id: str) -> None:
    for key in [k for k in self._states if k[0] == episode_id]:
        self._states.pop(key, None)
```
`latency_anomaly.py:293-295`

That list comprehension iterates a dict shared across **all** episodes. `Monitor.end_episode` and `_evict_episode` call `detector.reset(episode_id)` while holding only `episode_lock(episode_id)` (`monitor.py:641-643`, `monitor.py:704`), so a concurrent `ingest()` for a *different* episode that meets a new tool for the first time inserts into `_states` mid-iteration:

```
RuntimeError: dictionary changed size during iteration
```

## Why this is worse than a stray traceback

The Monitor catches it fail-open and logs `detector latency_anomaly reset raised during eviction; ignoring (fail-open)`, so nothing crashes — and that is the problem:

1. **The episode's state is never freed.** `reset` aborts partway, leaving the failed episode's `(episode_id, tool_name)` entries in `_states` for the life of the process. There is no retry.
2. **It silently defeats #184.** The per-episode retention cap exists precisely to bound memory for hosts that never call `end_episode`. Eviction runs `reset` on every detector; when that `reset` is the one that can fail under load, the cap pops the LRU id from `_live_episodes` (so `episodes_active` drops and the gauge looks healthy) while the actual per-tool CUSUM state stays behind. The leak the cap was added to close reopens, and the metric that would reveal it is the one thing that *does* get updated.
3. **The fault log is deduplicated once per distinct message (#14)**, so a busy multi-episode host sees this exactly once and then never again, no matter how many episodes leak afterwards.

The window is not narrow, either: the race triggers on *first sight of any new tool in any other episode*, which is the common case for a host monitoring many short concurrent runs — precisely the workload the cap targets.

I hit this incidentally while reproducing a different concurrency bug — the line `snagline: detector latency_anomaly reset raised during eviction; ignoring (fail-open)` showed up unprompted in a 4-thread ingest loop.

## Reproduction

Detector in isolation, no Monitor needed:

```python
import threading, itertools

det = LatencyAnomalyDetector()
stop, errors = threading.Event(), []


def ev(ep, tool, i):
    return StepEvent(step_id=str(i), episode_id=ep, timestamp=float(i),
                     action_type="tool_call",
                     action_signature=make_signature("tool_call", tool, str(i)),
                     tool_name=tool, latency_ms=10.0)


def observer():                      # other episodes, new tool each step
    for n in itertools.count():
        if stop.is_set():
            return
        det.observe(ev(f"other-{n}", f"tool-{n}", n))


def resetter():                      # our episode tears itself down
    for n in itertools.count():
        if stop.is_set():
            return
        det.observe(ev("mine", f"t{n}", n))
        try:
            det.reset("mine")
        except Exception as e:
            errors.append(e)
            stop.set()


ts = [threading.Thread(target=observer, daemon=True) for _ in range(3)]
ts.append(threading.Thread(target=resetter, daemon=True))
for t in ts:
    t.start()
stop.wait(timeout=10.0)
stop.set()
print(errors[:1])
print("leaked:", [k for k in list(det._states) if k[0] == "mine"])
```

On `master` (22faeae):

```
CRASH in reset(): RuntimeError: dictionary changed size during iteration
leaked 'mine' keys after failed reset: 1
```

## Suggested fix

Guard `_states` with a small `threading.Lock` on the operations that change or walk the dict's key set — which is *not* the hot path:

- `observe`: only the first-sight insert `self._states[key] = state` (the per-step `state.update(...)` mutates the existing `_WelfordCUSUM` object, changes no keys, and stays lock-free)
- `reset`: the scan-and-pop
- `dump_state` / `load_state`: the full walk / replacement

So the lock is taken once per `(episode, tool)` pair and once per teardown, never per step; `O(1)` amortized per event is preserved.

Restructuring `_states` into `dict[str, dict[str, _WelfordCUSUM]]` would make `reset` a plain `pop` and match every other detector, but it changes the snapshot payload shape and needs a compatibility path, so the lock looks like the better-scoped fix.

This is a separate defect from the unlocked `_clocks` iteration in `Monitor.snapshot_dict()` (filed separately): different module, different call path, and a Monitor-level lock around snapshotting would not fix this one, since `reset` and `observe` race while holding two *different* per-episode locks.

Happy to send a PR.

Adhi1-2:
Fixes #228
Fixes #229

Two dicts in the hot path are keyed **across** episodes rather than partitioned by one, so the Monitor's *per-episode* lock never protected their key sets. Same root cause, same shape of fix, so they ship together.

## #228 — `Monitor._clocks`

`snapshot_dict()` iterated `_clocks` with nothing held while `_advance_clock` inserted under only the episode lock. A snapshot taken from a sidecar/cron thread while the agent kept ingesting raised:

```
RuntimeError: dictionary keys changed during iteration
```

Unlike a detector or sink fault this one is **not** fail-open — it propagated into the host that called `snapshot()`, and it lost the snapshot it was taking. Contradicts README.md:172 ("Thread-safe").

`_clocks_lock` now covers the dict-level operations only:

| site | operation |
| --- | --- |
| `_advance_clock` | the `get` / first-sight insert |
| `end_episode`, `_evict_episode` | the `pop` |
| `snapshot_dict` | copy the pairs, build the `time_axis` payload afterwards |
| `restore_dict` | the repopulation (and the over-cap eviction pop) |

Fields *inside* an `_EpisodeClock` stay under the per-episode lock exactly as before. It is a **leaf lock** — nothing is ever acquired while holding it — so there is no ordering hazard against `_live_lock` or the episode locks. I also corrected the comment at `monitor.py:356` that claimed the episode lock made concurrent ingests serialize here; it only does so *within* one episode id.

## #229 — `LatencyAnomalyDetector._states`

Keyed by `(episode_id, tool_name)`, so `reset` must scan rather than `pop`:

```python
for key in [k for k in self._states if k[0] == episode_id]:
```

That is a Python-level comprehension, so it yields mid-walk and teardown for one episode raced an ingest meeting a new tool in another. The Monitor caught the `RuntimeError` fail-open, which is what makes it nasty: `reset` aborted partway and left the episode's per-tool CUSUM state behind for the life of the process — **reopening the leak the #184 retention cap exists to close**, while `episodes_active` still dropped and the gauge looked healthy. The fault log is deduplicated once per distinct message (#14), so a busy host saw it once and then never again.

`_states_lock` now covers the first-sight insert, `reset`, `dump_state` and `load_state`. The insert uses `setdefault` so a concurrent first sight of the same key cannot be lost. `load_state` builds the replacement dict outside the lock and swaps it in under it.

## Cost

Both locks are off the hot path. The per-step `state.update(...)` mutates an existing `_WelfordCUSUM` and adds no key, so it stays lock-free; the time axis is inert unless `SNAGLINE_IDLE_WARN_SECONDS` or `SNAGLINE_MAX_EPISODE_WALL_SECONDS` is set. `O(1)` amortized per event is preserved.

`benchmarks/overhead_benchmark.py`, 200,000 steps, same checkout with and without these changes:

| | median | p99 |
| --- | --- | --- |
| with fix | 1.81 us/step | 1.94 us/step |
| without | 1.83 us/step | 1.93 us/step |

## Tests

New `tests/test_shared_state_concurrency.py`, 10 cases. The five race tests drive several threads for a bounded 1.5 s each; three of them fail on `master` within that window and all pass with the fix. The rest pin behavior that must not change: `time_axis` snapshot content, the no-delta early return on an episode's first event, the time-axis snapshot→restore round trip, `reset` dropping only the named episode, and the `load_state` round trip and replace-not-merge semantics (guarding the restructure).

`test_dump_state_survives_concurrent_observe` passes on `master` too — `sorted(d.items())` builds its list in C without yielding, so that walk was safe by accident rather than by design. Keeping it as a guard now that the lock makes it deliberate.

## Not fixed here — one related defect remains

While writing the #228 tests I found the same class of race in **every** detector's `dump_state`, because they use Python-level comprehensions over their per-episode dicts:

```
File ".../monitor.py", line 765, in snapshot_dict
    detectors[_detector_key(i, detector)] = dump() if callable(dump) else None
File ".../detectors/loop.py", line 374, in dump_state
    "windows": {ep: list(w) for ep, w in self._windows.items()},
RuntimeError: dictionary changed size during iteration
```

So `Monitor.default().snapshot_dict()` under concurrent ingest is still exposed after this PR, via a different mechanism in different modules. Filed separately as #231 with the traceback rather than widened into this PR; happy to send that one next.

## Verification

```
707 passed, 2 skipped          (baseline on this branch point: 697 passed, 2 skipped)
Total coverage: 87.96%         (gate: 80%)
ruff check src tests           All checks passed!
ruff format --check src tests  130 files already formatted
mypy src                       Success: no issues found in 55 source files
```

### DIFF
--- a/src/snagline/detectors/latency_anomaly.py
+++ b/src/snagline/detectors/latency_anomaly.py
@@ -32,6 +32,7 @@ knobs: k/h stay as configured.
 from __future__ import annotations
 
 import math
+import threading
 from typing import Any
 
 from snagline.baseline import BaselineProfile, ToolBaseline
@@ -178,6 +179,16 @@ class LatencyAnomalyDetector:
         # per-tool stats only; no content is retained or consulted.
         self._baseline = baseline
         self._states: dict[tuple[str, str], _WelfordCUSUM] = {}
+        # ``_states`` is keyed by (episode_id, tool_name), so unlike every
+        # other detector it is not partitioned by episode: the Monitor's
+        # per-episode lock cannot keep reset()/dump_state() from walking the
+        # dict while an ingest for a *different* episode meets a new tool and
+        # inserts into it (issue #229). This lock guards only the operations
+        # that change or walk the key set -- first-sight insert, reset,
+        # dump_state, load_state -- so the per-step ``state.update(...)``,
+        # which mutates an existing _WelfordCUSUM and adds no key, stays
+        # lock-free and O(1) amortized per event.
+        self._states_lock = threading.Lock()
         # Periodic baseline re-fit (issue #92); 0 disables (the default).
         self.refit_every = cfg.cusum_refit_every
 
@@ -208,7 +219,11 @@ class LatencyAnomalyDetector:
                     seeded = candidate
             if seeded is not None:
                 state.seed(seeded.mean_latency, seeded.std_latency)
-            self._states[key] = state
+            # Only the key-set mutation needs the lock (issue #229); building
+            # the state above is pure. A concurrent first-sight of the same key
+            # would otherwise be lost, so keep whichever entry landed first.
+            with self._states_lock:
+                state = self._states.setdefault(key, state)
 
         if not state.frozen:
             state.learn_only(event.latency_ms)
@@ -291,8 +306,9 @@ class LatencyAnomalyDetector:
         return False, 0.0, 0.0
 
     def reset(self, episode_id: str) -> None:
-        for key in [k for k in self._states if k[0] == episode_id]:
-            self._states.pop(key, None)
+        with self._states_lock:
+            for key in [k for k in self._states if k[0] == episode_id]:
+                self._states.pop(key, None)
 
     @staticmethod
     def _state_to_dict(s: _WelfordCUSUM) -> dict[str, Any]:
@@ -340,15 +356,14 @@ class LatencyAnomalyDetector:
     def dump_state(self) -> dict[str, Any]:
         # Keys are (episode_id, tool_name) tuples; JSON dict keys must be
         # strings, so serialize as [key-pair, state] entries.
+        with self._states_lock:
+            items = sorted(self._states.items())
         return {
-            "states": [
-                [[ep, tool], self._state_to_dict(s)]
-                for (ep, tool), s in sorted(self._states.items())
-            ]
+            "states": [[[ep, tool], self._state_to_dict(s)] for (ep, tool), s in items]
         }
 
     def load_state(self, state: dict[str, Any]) -> None:
-        self._states = {}
+        restored: dict[tuple[str, str], _WelfordCUSUM] = {}
         for key_pair, raw in state.get("states", []):
             ep, tool = key_pair[0], key_pair[1]
             s = _WelfordCUSUM(
@@ -358,4 +373,6 @@ class LatencyAnomalyDetector:
             # The live config owns the refit cadence; the snapshot only
             # carries accumulated bookkeeping.
             s.refit_every = self.refit_every
-            self._states[(ep, tool)] = s
+            restored[(ep, tool)] = s
+        with self._states_lock:
+            self._states = restored
--- a/src/snagline/monitor.py
+++ b/src/snagline/monitor.py
@@ -215,6 +215,16 @@ class Monitor:
         self._warn_fraction = cfg.warn_fraction
         self._idle_warn_seconds = cfg.idle_warn_seconds
         self._clocks: dict[str, _EpisodeClock] = {}
+        # ``_clocks`` spans every episode, so the per-episode lock cannot
+        # protect its key set: an ingest for episode A inserting a clock races
+        # a teardown or a snapshot walking the dict for anything else. This
+        # lock covers the dict-level operations only (get / insert / pop /
+        # iterate); the fields *inside* an _EpisodeClock stay under the
+        # per-episode lock, so the hot path takes this lock once per episode
+        # first-sight and once per teardown, never per step. It is a leaf lock
+        # -- nothing is ever acquired while holding it -- so no ordering
+        # hazard exists against _live_lock or the episode locks.
+        self._clocks_lock = threading.Lock()
         # Per-episode retention cap (issue #184): bounded LRU of live episode
         # ids. Every ingest touches the entry; when the cap is exceeded the
         # least-recently-seen id is evicted silently (no finalize risks) via
@@ -345,7 +355,9 @@ class Monitor:
         with self._state.episode_lock(event.episode_id):
             # Time-axis head (issue #92): computed from StepEvent timestamps at
             # the top of ingest(), under the episode lock so concurrent ingests
-            # serialize exactly like detector state does.
+            # for the same episode serialize exactly like detector state does.
+            # The episode lock says nothing about the cross-episode ``_clocks``
+            # dict, which carries its own lock (issue #228).
             risks.extend(self._advance_clock(event))
             for detector in self._detectors:
                 try:
@@ -390,11 +402,15 @@ class Monitor:
         if self._max_wall_seconds is None and self._idle_warn_seconds is None:
             return out
         try:
-            clock = self._clocks.get(event.episode_id)
+            # Dict-level access only under _clocks_lock; the returned clock's
+            # fields are mutated below under the caller's per-episode lock.
+            with self._clocks_lock:
+                clock = self._clocks.get(event.episode_id)
+                if clock is None:
+                    # First event of the episode establishes the reference
+                    # point; there is no delta yet, so nothing can fire.
+                    self._clocks[event.episode_id] = _EpisodeClock(event.timestamp)
             if clock is None:
-                # First event of the episode establishes the reference point;
-                # there is no delta yet, so nothing can fire.
-                self._clocks[event.episode_id] = _EpisodeClock(event.timestamp)
                 return out
             delta = event.timestamp - clock.last_ts
             clock.last_ts = event.timestamp
@@ -639,7 +655,8 @@ class Monitor:
                 return
         try:
             with self._state.episode_lock(episode_id):
-                self._clocks.pop(episode_id, None)
+                with self._clocks_lock:
+                    self._clocks.pop(episode_id, None)
                 for detector in self._detectors:
                     try:
                         detector.reset(episode_id)
@@ -685,7 +702,8 @@ class Monitor:
         with self._state.episode_lock(episode_id):
             # Time-axis state (issue #92) is per-episode like detector state;
             # drop it here so a reused episode id starts with a clean clock.
-            self._clocks.pop(episode_id, None)
+            with self._clocks_lock:
+                self._clocks.pop(episode_id, None)
             for detector in self._detectors:
                 finalize = getattr(detector, "finalize", None)
                 if callable(finalize):
@@ -750,7 +768,14 @@ class Monitor:
             dump = getattr(sink, "dump_state", None)
             sinks[f"{i}:{type(sink).__name__}"] = dump() if callable(dump) else None
         time_axis: dict[str, Any] = {}
-        for episode_id, clock in self._clocks.items():
+        # Copy the (episode_id, clock) pairs under the lock: iterating _clocks
+        # directly would raise RuntimeError as soon as a concurrent ingest
+        # meets a new episode (issue #228). Building the payload afterwards is
+        # safe -- each clock object is only mutated under its episode lock, and
+        # a snapshot is a point-in-time read either way.
+        with self._clocks_lock:
+            clocks = list(self._clocks.items())
+        for episode_id, clock in clocks:
             time_axis[episode_id] = {
                 "last_ts": clock.last_ts,
                 "elapsed": clock.elapsed,
@@ -886,8 +911,10 @@ class Monitor:
                         self._live_episodes[episode_id] = None
                         if len(self._live_episodes) > self._max_live_episodes:
                             evicted, _ = self._live_episodes.popitem(last=False)
-                            self._clocks.pop(evicted, None)
-                    self._clocks[episode_id] = clock
+                            with self._clocks_lock:
+                                self._clocks.pop(evicted, None)
+                    with self._clocks_lock:
+                        self._clocks[episode_id] = clock
             # If we evicted due to cap, live_snapshot entries not in time_axis
             # are already accounted for; no further action needed.
 
--- a/tests/test_shared_state_concurrency.py
+++ b/tests/test_shared_state_concurrency.py
@@ -87,7 +87,8 @@ class _Worker:
 
 def test_snapshot_dict_survives_concurrent_ingest_of_new_episodes() -> None:
     monitor = Monitor(detectors=[], sinks=[], config=_HORIZON)
-    stop, errors = threading.Event(), []
+    stop = threading.Event()
+    errors: list[BaseException] = []
 
     class Ingest(_Worker):
         def __init__(self, tag: str) -> None:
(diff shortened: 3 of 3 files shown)


anya-research:
Detailed review of #232 — **fix(core): lock the cross-episode dicts** — Fixes #228/#229

### Verdict: Approve. Correct minimal fix, well-tested, no deadlock/performance regression. Follow-up #231 already filed.

### Summary

Two dicts are keyed **across** episodes (`Monitor._clocks: dict[episode_id, _EpisodeClock]` in `src/snagline/monitor.py:217`, `LatencyAnomalyDetector._states: dict[(episode_id,tool), _WelfordCUSUM]` in `src/snagline/detectors/latency_anomaly.py:181`) so the sharded `StateBackend.episode_lock(episode_id)` in `src/snagline/state.py:57` never protected the key-set. Snapshot/iteration vs. insert/pop raced to `RuntimeError: dict changed size`.

PR adds one leaf `threading.Lock` per dict, covering only key-set ops. Field mutation stays under the per-episode lock. Hot-path `state.update()` stays lock-free.

### #228 — `Monitor._clocks` — `src/snagline/monitor.py`

**Root cause correct.** `snapshot_dict()` iterated `self._clocks.items()` with no lock in `src/snagline/monitor.py:753` while ` _advance_clock()` inserted under only the episode lock (`src/snagline/monitor.py:355` comment now fixed). Not fail-open; propagated into caller of public `snapshot()`.

**Fix correct & complete:**

* `src/snagline/monitor.py:227` — `self._clocks_lock = threading.Lock()` with comment explaining leaf semantics.
* `src/snagline/monitor.py:407-412` — `get`+first-sight `insert` atomic under lock. Restructured to early-return outside lock; fields mutated afterwards under caller's `episode_lock` (`src/snagline/monitor.py:355`). No TOCTOU for same episode because `ingest()` already serializes same `episode_id` (`src/snagline/monitor.py:355`).
* `src/snagline/monitor.py:658-659` (`_evict_episode`), `705-706` (`end_episode`): `pop` under lock while holding `episode_lock`. Ordering is `episode_lock -> _clocks_lock` — consistent everywhere. The doc “leaf lock — nothing is ever acquired while holding it” (`src/snagline/monitor.py:223`) holds: lock is innermost, never ` _clocks_lock -> episode_lock/_live_lock`.
* `src/snagline/monitor.py:776-777` (`snapshot_dict`): `list(self._clocks.items())` under lock, payload built afterwards. Fixes `RuntimeError` and preserves point-in-time semantics. Field reads (`clock.last_ts` etc., `src/snagline/monitor.py:778-785`) remain unsynchronized with per-episode lock — benign best-effort race, same as pre-PR but now dict iteration is safe. Object lifetime held by copied list so concurrent `pop` cannot free it.
* `src/snagline/monitor.py:914-917` (`restore_dict`): both eviction `pop` and repopulation `insert` under lock while outer `self._live_lock` held (`src/snagline/monitor.py:907`). Ordering `_live_lock -> _clocks_lock` consistent with `ingest()`’s `_live_lock -> _evict -> episode_lock -> _clocks_lock`. Split into two `with _clocks_lock` blocks is safe; `_live_lock` stays held across both.

All ` _clocks` sites covered (grep confirms no bare access remains). Comment fix at `src/snagline/monitor.py:359` correct.

**Performance:** Claimed O(1) amortized preserved — only first sight per episode and teardown snapshot take lock. Benchmark 200k steps 1.81µs vs 1.83µs median confirms.

### #229 — `LatencyAnomalyDetector._states` — `src/snagline/detectors/latency_anomaly.py`

**Root cause correct.** Keyed `(episode,tool)` so `reset()` must scan `src/snagline/detectors/latency_anomaly.py:310` (`[k for k in self._states if k[0]==episode_id]` yields), raced with ingest inserting new tool for *other* episode. Fail-open swallow left leaked CUSUM state, defeating #184 cap while gauge looked healthy. Log dedup (#14) hid it.

**Fix correct:**

* `src/snagline/detectors/latency_anomaly.py:191` — `_states_lock`.
* `src/snagline/detectors/latency_anomaly.py:225-226` — first-sight `setdefault` under lock; pure construction outside. Handles concurrent first-sight of same key (defensive; same episode is already serialized). `get` at `src/snagline/detectors/latency_anomaly.py:205` stays outside lock — safe: `dict.get` is GIL-atomic single op, iteration vs `get` not the failure mode; key sets guarded.
* `src/snagline/detectors/latency_anomaly.py:309-311` — `reset` scan+`pop` under lock. Together blocks concurrent insert.
* `src/snagline/detectors/latency_anomaly.py:359-360` — `dump_state` copies `sorted(items)` under lock, comprehension over list outside. Note `sorted(d.items())` previously happened to be C-level without yield (PR correctly calls it accidental safety); now deliberate.
* `src/snagline/detectors/latency_anomaly.py:366-378` — `load_state` builds `restored` outside lock, swaps `self._states = restored` under lock. Correct for setup-time use; would discard concurrent inserts if called concurrently with ingest (not a supported race).

Leaf again, no ordering hazard. Hot-path `state.update()`/`learn_only()` (`src/snagline/detectors/latency_anomaly.py:228-234`) untouched.

### Tests — `tests/test_shared_state_concurrency.py`

10 cases, 5 race tests bounded 1.5s. Good:

* `#228`: `test_snapshot_dict_survives_concurrent_ingest_of_new_episodes` (`:58`), `test_snapshot_dict_survives_concurrent_end_episode` (`:79`) directly drive reported race.
* `#229`: `test_reset_survives_concurrent_observe_of_other_episodes` (`:112`), `test_dump_state_survives_concurrent_observe` (`:140`) — latter documents accidental C safety.
* Invariants pinned: `test_time_axis_snapshot_content_is_unchanged`, `test_first_event...`, `test_time_axis_round_trip`, `test_reset_still_only_drops_the_named_episode` (`:169`), `test_state_round_trip`, `test_load_state_replaces_rather_than_merges`.

Implementation uses shared `threading.Event` + `itertools.count` spin, 3 ingest threads + 1 snapshot/teardown thread. `diagnose` threads are daemon, `stop.set()` + `join(10)` in `src/snagline/detectors/latency_anomaly.py` helpers. Flaky-risk low; author notes 3/5 fail on master within window. `type: ignore[attr-defined]` for `worker.stop` acceptable.

### Not fixed — #231

PR correctly leaves every other detector’s `dump_state` (e.g. `LoopDetector.dump_state` in `src/snagline/detectors/loop.py:374` `{ep: list(w) for ep,w in self._windows.items()}`) racy for `Monitor.default().snapshot_dict()` (`src/snagline/monitor.py:765`). Filed separately; scope containment appropriate.

### Minor nits (non-blocking)

* `snapshot_dict` field reads remain data race (float/int under GIL — atomic but unordered). Acceptable for point-in-time snapshot; stricter would hold per-episode locks or copy fields under lock.
* Latency `get` outside `_states_lock` is stylistically inconsistent though safe per-episode serialization. Holding lock for `get` would close the theoretical `get` vs `pop` window more explicitly at zero cost.
* `_clocks_lock`/`_states_lock` are `Lock` not `RLock` — correct for leaf; re-entrance would deadlock, but no path re-enters.

Approval: ships together, same root cause/shape, minimal, tested, no perf impact.

[github run](/Cyrax321/SNAGLINE/actions/runs/33604570775)

Cyrax321 (approved):
Verified locally: 10/10 pass in test_shared_state_concurrency.py, ruff + mypy clean. The _clocks_lock is documented as a leaf lock (never held while acquiring another), and the hot path only takes it on first-sight/teardown, so per-step cost is unchanged. Latency detector keeps per-step updates lock-free with setdefault for first-sight. Safe to merge.

PR #233, opened 2026-08-31
fix(detectors): copy per-episode state before serializing it (Fixes #231)

### LINKED ISSUE #231: [core] Every detector's dump_state walks its per-episode dict with a Python-level comprehension, so Monitor.snapshot() under concurrent ingest raises RuntimeError

Adhi1-2:
Fixes #231.

## The bug

`Monitor.snapshot()` calls every detector's `dump_state` with **no lock**, on
whichever thread asked for the snapshot. Detector state is partitioned by
`episode_id`, so the *values* are safe under the Monitor's per-episode lock —
but the **key set** is shared, and a per-episode lock cannot serialize a walk of
it against an insert for a *different* episode.

Every `dump_state` walked its dicts with a Python-level comprehension. Those
yield between bytecodes, so a concurrent ingest meeting a new episode inserts
mid-walk:

```
File "src/snagline/monitor.py", line 747, in snapshot_dict
  detectors[_detector_key(i, detector)] = dump() if callable(dump) else None
File "src/snagline/detectors/error_cascade.py", line 141, in dump_state
  "windows": {ep: list(w) for ep, w in self._windows.items()},
RuntimeError: dictionary keys changed during iteration
```

That escapes `Monitor.snapshot` / `Monitor.snapshot_dict` — public APIs with no
fail-open guard, because snapshot/restore faults are deliberately loud. So the
documented "snapshot a long-running monitor so it survives a restart" workflow
(#91) fails intermittently on exactly the deployments it exists for: ones busy
enough to have several episodes in flight.

`Monitor.default()` reproduces it in **under a second** with three ingest
threads and one snapshot thread.

## The fix

Add `detectors.base.snapshot_items`, which copies the pairs via
`list(state.items())` — a single C-level build with no Python executing in
between, so it cannot be interrupted — and route all **14 affected walks in 10
modules** through it, building the JSON payload from the copy. The `dict(...)`
copies already had this property and are left alone.

Sites: `loop` (×5), `error_cascade`, `stagnation`, `meltdown`, `token_runaway`,
`side_effect_guard`, `compaction_tripwire`, `goal_drift`, `silent_abort`,
`drift/goal_drift`.

### Why not locks

In #231 I offered two options — per-detector locks or a Monitor-wide
shared/exclusive gate — and asked you to pick. Having then mapped every call
site, I think **neither is warranted**, so this PR does something narrower.
Happy to switch if you disagree:

- `observe` is **untouched** — no hot-path cost, and the O(1)-per-step
  guarantee needs no re-argument.
- Every `reset` was already a single-key `pop` (atomic), not a scan, so
  teardown never needed a lock. That is what distinguished
  `LatencyAnomalyDetector`, whose `(episode_id, tool_name)` key forces a scan
  and so does take a lock in #232.
- No new lock means no ordering hazard against `_live_lock`, the episode locks,
  or `_clocks_lock`, and nothing to get wrong later.

Cross-dict atomicity is deliberately **not** added: a detector holding several
dicts may capture episode X in one and miss it in another. Every `load_state`
already reads each dict independently and defaults what is absent, and a
snapshot taken during live ingest is a point-in-time smear regardless.
Guaranteeing more would mean holding one lock across the whole of `observe`,
which would serialize ingest across episodes and cost the per-episode
concurrency the Monitor is built around. The `snapshot_items` docstring records
that reasoning.

## Tests

`tests/test_dump_state_concurrency.py`, 32 tests:

- **Race (11)** — the reported repro on `Monitor.default().snapshot_dict()`,
  the same race through the on-disk `snapshot(path)` path, and one per detector
  driving maximum key-set churn (a brand-new episode every step). Each drives
  the collision for 1.5s across four threads.
- **Behavior-preserving (18)** — per detector: `dump_state` still records every
  live episode, and `dump_state` → `load_state` → `dump_state` still
  round-trips byte-identically.
- **Helper (3)** — `snapshot_items` preserves every pair, returns a copy an
  later insert cannot reach into, and tolerates an empty dict.

All 11 race tests fail on `master` and pass here; the other 21 pass both ways,
which is the point of having them.

Verified with the full CI gate set locally: **729 passed, 2 skipped**, coverage
87.62%, `ruff check src tests`, `ruff format --check src tests` and
`mypy src tests` all clean.

Independent of #230 and #232 — no overlapping files.

### DIFF
--- a/src/snagline/detectors/base.py
+++ b/src/snagline/detectors/base.py
@@ -8,10 +8,14 @@ episode ends so per-episode state does not leak across runs.
 
 from __future__ import annotations
 
-from typing import Any, Protocol
+from collections.abc import Mapping
+from typing import Any, Protocol, TypeVar
 
 from snagline.events import StepEvent
 from snagline.risk import FailureRisk
+
+_K = TypeVar("_K")
+_V = TypeVar("_V")
 
 
 class Detector(Protocol):
@@ -66,3 +70,32 @@ class StatefulDetector(Protocol):
     def load_state(self, state: dict[str, Any]) -> None:
         """Restore internal state previously produced by ``dump_state``."""
         ...
+
+
+def snapshot_items(state: Mapping[_K, _V]) -> list[tuple[_K, _V]]:
+    """Atomically copy a per-episode state dict before serializing it (#231).
+
+    ``dump_state`` runs on whichever thread called ``Monitor.snapshot`` and
+    takes no lock, while an ingest on another thread may be meeting a new
+    episode and inserting its first-sight key. Detector state is partitioned by
+    ``episode_id``, so the *values* are safe -- but the key set is shared, and
+    the Monitor's per-episode lock cannot serialize a walk of it against an
+    insert for a different episode.
+
+    A Python-level comprehension over ``state.items()`` yields between
+    bytecodes, so that insert lands mid-walk and ``RuntimeError: dictionary
+    changed size during iteration`` escapes ``Monitor.snapshot`` -- a public API
+    with no fail-open guard, because restore-time errors are deliberately loud.
+    ``list(state.items())`` builds the copy inside a single C call with no
+    Python executing in between, so it cannot be interrupted; callers then
+    build their JSON payload from the returned copy.
+
+    Cross-dict atomicity is deliberately *not* provided: a detector holding
+    several dicts may snapshot episode X in one and miss it in another. Every
+    ``load_state`` already reads each dict independently and defaults what is
+    absent, and a snapshot taken during live ingest is a point-in-time smear
+    either way. Guaranteeing more would mean holding one lock across the whole
+    of ``observe``, which would serialize ingest across episodes and cost the
+    per-episode concurrency the Monitor is built around.
+    """
+    return list(state.items())
--- a/src/snagline/detectors/compaction_tripwire.py
+++ b/src/snagline/detectors/compaction_tripwire.py
@@ -50,6 +50,7 @@ from __future__ import annotations
 from typing import Any
 
 from snagline.config import Config
+from snagline.detectors.base import snapshot_items
 from snagline.events import StepEvent
 from snagline.risk import FailureRisk
 
@@ -163,9 +164,12 @@ class CompactionTripwireDetector:
         A restart between a compaction and its confirmations must not let the
         grace deadline vanish silently: the pending set, its deadline ordinal,
         and the fired latch all survive.
+
+        The walk goes through ``snapshot_items``: a concurrent ingest meeting a
+        new episode must not change the key set mid-loop (issue #231).
         """
         episodes: dict[str, Any] = {}
-        for ep, st in self._episodes.items():
+        for ep, st in snapshot_items(self._episodes):
             entry: dict[str, Any] = {"ordinal": st.ordinal, "pending": None}
             if st.pending is not None:
                 entry["pending"] = {
--- a/src/snagline/detectors/error_cascade.py
+++ b/src/snagline/detectors/error_cascade.py
@@ -19,6 +19,7 @@ from collections import deque
 from typing import Any
 
 from snagline.config import Config
+from snagline.detectors.base import snapshot_items
 from snagline.detectors.windowing import next_window
 from snagline.events import StepEvent
 from snagline.risk import FailureRisk
@@ -137,8 +138,10 @@ class ErrorCascadeDetector:
         self._fired.pop(episode_id, None)
 
     def dump_state(self) -> dict[str, Any]:
+        # snapshot_items: a concurrent ingest meeting a new episode must not
+        # change the key set mid-comprehension (issue #231).
         return {
-            "windows": {ep: list(w) for ep, w in self._windows.items()},
+            "windows": {ep: list(w) for ep, w in snapshot_items(self._windows)},
             "counts": dict(self._counts),
             "consecutive": dict(self._consecutive),
             "fired": dict(self._fired),
--- a/src/snagline/detectors/goal_drift.py
+++ b/src/snagline/detectors/goal_drift.py
@@ -21,6 +21,7 @@ from typing import Any
 
 from snagline.baseline import BaselineProfile
 from snagline.config import Config
+from snagline.detectors.base import snapshot_items
 from snagline.events import StepEvent
 from snagline.risk import FailureRisk
 
@@ -98,8 +99,10 @@ class GoalDriftDetector:
         self._fired.pop(episode_id, None)
 
     def dump_state(self) -> dict[str, Any]:
+        # snapshot_items: a concurrent ingest meeting a new episode must not
+        # change the key set mid-comprehension (issue #231).
         return {
-            "live": {ep: p.to_dict() for ep, p in self._live.items()},
+            "live": {ep: p.to_dict() for ep, p in snapshot_items(self._live)},
             "fired": dict(self._fired),
         }
 
--- a/src/snagline/detectors/loop.py
+++ b/src/snagline/detectors/loop.py
@@ -45,6 +45,7 @@ from collections.abc import Callable
 from typing import Any, cast
 
 from snagline.config import Config
+from snagline.detectors.base import snapshot_items
 from snagline.detectors.windowing import next_window
 from snagline.events import StepEvent
 from snagline.risk import FailureRisk, TriggerType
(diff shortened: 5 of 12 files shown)


yuki-fuyutsuki:
**Yuki Full Review — PR #233 `fix/dump-state-key-set-race-231`**

### Verdict: Approve (with 1 follow-up nit)

Correct, minimal, well-documented fix for the `RuntimeError: dictionary keys changed during iteration` that escapes `Monitor.snapshot_dict()` / `Monitor.snapshot()` (`src/snagline/monitor.py:747`). No hot-path cost, no new lock-order hazard. 729 passed locally per PR body is credible from inspection.

---

#### 1. The fix is sound — `src/snagline/detectors/base.py:75`

`snapshot_items()`:

```python
def snapshot_items(state: Mapping[_K,_V]) -> list[tuple[_K,_V]]:
    return list(state.items())
```

* Correct typing (`Mapping[_K,_V]`, `TypeVar`) `src/snagline/detectors/base.py:17`.
* Docstring at `src/snagline/detectors/base.py:76` correctly states why `list(state.items())` is safe: single C call, no Python bytecode between, so GIL-holder cannot be pre-empted by another Python thread inserting a new `episode_id`. This holds for CPython 3.10-3.13 (project floor). Intentionally **not** claiming free-threaded 3.13t (`--disable-gil`) safety — acceptable, but worth a `NOTE` if you ever support it.
* Cross-dict non-atomicity disclaimer `src/snagline/detectors/base.py:93` is deliberate and consistent with `load_state` tolerant `state.get(..., {}).items()` in every detector.

#### 2. Coverage — 14 walks across 10 modules, all correctly routed

| file | walk replaced | left as `dict()` (atomic C-build, correct to leave) |
|---|---|---|
| `src/snagline/detectors/loop.py:379,381,382-384,386-388,389-391` | 5× `_windows`, `_fired`, `_near_windows`, `_near_fired`, `_cycle_windows` | `counts`, `near_counts`, `cycle_counts`, `cycle_fired`, `stall_*` `src/snagline/detectors/loop.py:380,385,392-397` |
| `src/snagline/detectors/error_cascade.py:144` | `_windows` | `counts`, `consecutive`, `fired` `src/snagline/detectors/error_cascade.py:145` |
| `src/snagline/detectors/stagnation.py:181` | `_windows` | `counts` `src/snagline/detectors/stagnation.py:184` |
| `src/snagline/detectors/meltdown.py:172` | `_eps` | `counts`, `fired`, `clear_streak` `src/snagline/detectors/meltdown.py:173` |
| `src/snagline/detectors/token_runaway.py:144` | `_states` | `totals`, `warned`, `breached` `src/snagline/detectors/token_runaway.py:146` |
| `src/snagline/detectors/side_effect_guard.py:127` | `_counts` (outer) | inner `sorted(...)` `src/snagline/detectors/side_effect_guard.py:123` — doc correctly notes atomic `src/snagline/detectors/side_effect_guard.py:116` |
| `src/snagline/detectors/compaction_tripwire.py:172` | `_episodes` | — |
| `src/snagline/detectors/goal_drift.py:105` | `_live` | `fired` |
| `src/snagline/detectors/silent_abort.py:88` | `_last` | `output_action_types` sorted |
| `src/snagline/drift/goal_drift.py:169` | `_episodes` | filter `if st.sum is not None` correctly applied *after* copy |

`LatencyAnomalyDetector` `src/snagline/detectors/latency_anomaly.py:341` is intentionally untouched — its `(episode_id, tool_name)` key forces a scan on `reset` and it takes a lock per #232. Out of scope is correct.

#### 3. One remaining surface — `src/snagline/monitor.py:752`

```python
for episode_id, clock in self._clocks.items():  # snapshot_dict time_axis
```

`_clocks` is mutated under `episode_lock` in `_advance_clock()` `src/snagline/monitor.py:392` and read with **no lock** in `snapshot_dict`. Same key-set race, just narrower (idle/budget feature off by default). Not a blocker for this PR (detectors were the reported crash), but follow-up should wrap it via `list(self._clocks.items())` or `snapshot_items` for parity. No test currently covers `_clocks` churn.

#### 4. `observe()` / `reset()` untouched — preserves O(1) guarantee

No lock added to hot path, no `reset` change (single-key `pop` is atomic). Agree with PR's "Why not locks" rationale.

#### 5. Tests — `tests/test_dump_state_concurrency.py:1`

* Race harness `_Worker` / `_race` `tests/test_dump_state_concurrency.py:82,104` correctly drives 1.5 s collision with 4 threads; reproduces in <1s on unfixed code.
* `test_monitor_default_snapshot_dict_survives_concurrent_ingest` `tests/test_dump_state_concurrency.py:119` and `test_monitor_default_snapshot_file_survives...` `tests/test_dump_state_concurrency.py:142` cover both public snapshot paths.
* `test_detector_dump_state_survives_concurrent_observe` `tests/test_dump_state_concurrency.py:165` parametrised over 8 factories maximizes key-set churn (new episode per step).
* Helper tests `tests/test_dump_state_concurrency.py:222,228,238` pin `snapshot_items` contract.

**Nit — CodeRabbit false positive:** `tests/test_dump_state_concurrency.py:61` comment says `goal_drift` is "inert … so _live never fills" but `GoalDriftDetector.observe()` `src/snagline/detectors/goal_drift.py:49` does `self._live.setdefault(episode_id, BaselineProfile()).add_event(event)` even with empty `BaselineProfile()`, so `dump_state` `src/snagline/detectors/goal_drift.py:105` **is** populated and `test_detector_dump_state_still_records_every_live_episode` `tests/test_dump_state_concurrency.py:189` will pass (walked = `[live_dict]`). No fix needed, but comment is misleading; consider dropping it or noting `_live` still fills.

**Minor count mismatch:** PR body says "11 race / 18 behavior / 3 helper = 32 tests" but file yields 10 race (2 monitor + 8 detector) + 16 behavior (8+8) + 3 helper = 27 parametrized cases (29 with helpers expanded). Not functional, just doc drift.

#### 6. Style / hygiene

* `ruff` / `mypy` clean per PR; imports sorted, `from snagline.detectors.base import snapshot_items` added consistently.
* Determinism preserved (`sorted` on sets, `sorted(..., key=repr)` in `side_effect_guard`).
* No `dict` walk left via Python comprehension.

**Recommendation:** Merge as-is; optionally follow up with `self._clocks` copy in `Monitor.snapshot_dict()` and clarify the `goal_drift` factory comment. No blocking issues found.

[github run](/Cyrax321/SNAGLINE/actions/runs/33604531754)

Cyrax321 (approved):
Verified locally: 32/32 pass in test_dump_state_concurrency.py, ruff + mypy clean. Routing every per-episode dump_state walk through snapshot_items (single C-call copy) is the right minimal fix, and the docstring is honest that cross-dict atomicity is intentionally not provided. Behavior-preserving. Safe to merge.

PR #235, opened 2026-09-07
fix(core): sort snapshot detector keys by slot; validate strict_names before mutating (Fixes #217)

### LINKED ISSUE #217: [core] Monitor.restore(strict_names=True) sorts snapshot detector keys lexicographically: 11+ detectors reject a matching composition and can accept a mismatched one
### Issue type

Bug

Yes — `faa4a9e255607be656665b90bfa93a2c199dc2cb`

### Source

source

### SNAGLINE version

`0.1.1.dev0` / commit `faa4a9e255607be656665b90bfa93a2c199dc2cb`

### Custom code

No

Windows 11 (platform-independent — pure Python string sorting, no OS dependency)

### Python version

3.13.1

---

## Current behavior?

`Monitor.restore(path, strict_names=True)` compares detector names in **lexicographic key order** instead of slot order. Snapshot keys are built by `_detector_key` as `f"{index}:{name}"`, so `sorted()` places slot `10` between slot `1` and slot `2`. Once a monitor runs 11 or more detectors, the strict-composition check is comparing two differently-ordered lists and its verdict is wrong in both directions:

1. **A matching composition is rejected.** Two monitors built the same way, one snapshotting and one restoring, raise `ValueError: snapshot detector composition mismatch`. Because restore is documented as a setup-time operation that raises rather than failing open, this aborts host startup.
2. **A mismatched composition is accepted.** When the restoring monitor's real slot order happens to equal the lexicographic order of the snapshot's keys, the check passes even though slot 1 holds a different detector than the snapshot's slot 1 did — which is precisely what the flag exists to catch.

Eleven detectors is not a synthetic number: it is exactly what `Monitor.default()` builds when the opt-in detectors are enabled, with **no extras installed**.

Expected behavior: `strict_names=True` accepts a composition that matches the snapshot slot-for-slot, and rejects one that does not, at any detector count.

## Affected component

* File: `src/snagline/monitor.py`
* Function: `Monitor.restore_dict` (the `strict_names` branch), line 831
* Reached from the public API: `Monitor.restore(path, strict_names=True)`

## Technical root cause

`snapshot_dict` keys each detector slot with `_detector_key(i, detector)`:

https://github.com/Cyrax321/SNAGLINE/blob/faa4a9e255607be656665b90bfa93a2c199dc2cb/src/snagline/monitor.py#L105-L107

```python
def _detector_key(index: int, detector: Any) -> str:
    """Stable per-position key for a detector inside a snapshot payload."""
    return f"{index}:{getattr(detector, 'name', type(detector).__name__)}"
```

The strict check then recovers the expected name order by sorting those keys as plain strings:

https://github.com/Cyrax321/SNAGLINE/blob/faa4a9e255607be656665b90bfa93a2c199dc2cb/src/snagline/monitor.py#L830-L837

```python
if strict_names:
    expected = [k.split(":", 1)[1] for k in sorted(dumped_detectors)]
    current = [getattr(d, "name", type(d).__name__) for d in self._detectors]
    if expected != current:
        raise ValueError(
            "snapshot detector composition mismatch: "
            f"snapshot={expected} monitor={current}"
        )
```

`sorted()` on the key strings is lexicographic, so for 11 slots it yields:

```
['0:loop', '10:governance_decay', '1:error_cascade', '2:latency_anomaly', ...]
```

`expected` therefore lists `governance_decay` second, while `current` — read from `self._detectors` by position — correctly lists it eleventh. The lists differ for any snapshot with a two-digit slot index, and they can also spuriously agree, as case 2 above shows. With 10 or fewer detectors every index is one digit, lexicographic order equals numeric order, and the bug is invisible — which is why the existing coverage misses it: `tests/test_monitor_snapshot.py::test_composition_mismatch_strict_vs_tolerant` uses 1–2 detectors and `test_snapshot_roundtrip...` uses 5.

## Standalone code to reproduce the issue

Case 1 — identical composition rejected, straight through `Monitor.default()`:

```python
import os, tempfile
from snagline.baseline import BaselineProfile


def cfg_all():
    cfg = Config()
    for flag in (
        "stagnation_enabled", "goal_drift_enabled", "semantic_drift_enabled",
        "token_runaway_enabled", "meltdown_enabled", "silent_abort_enabled",
        "side_effect_guard_enabled", "compaction_tripwire_enabled",
    ):
        setattr(cfg, flag, True)
    cfg.goal_drift_baseline = BaselineProfile()
    return cfg


a = Monitor.default(config=cfg_all(), sinks=[])
a.ingest(StepEvent(step_id="0", episode_id="ep", timestamp=0.0,
                   action_type="tool_call", action_signature="s0", latency_ms=10.0))

with tempfile.TemporaryDirectory() as d:
    path = os.path.join(d, "s.json")
    a.snapshot(path)
    b = Monitor.default(config=cfg_all(), sinks=[])
    assert [x.name for x in a._detectors] == [x.name for x in b._detectors]
    b.restore(path, strict_names=True)   # ValueError
```

Case 2 — mismatched composition accepted. Snapshot a monitor whose slots are
`[loop, error_cascade, latency_anomaly, stagnation, token_runaway, meltdown, silent_abort, side_effect_guard, governance_decay, goal_drift, loop]`, then restore into a monitor whose slots are
`[loop, loop, error_cascade, latency_anomaly, stagnation, token_runaway, meltdown, silent_abort, side_effect_guard, governance_decay, goal_drift]`
(the lexicographic order of the snapshot's keys) with `strict_names=True`. It passes, even though snapshot slot 1 was `error_cascade` and the monitor's slot 1 is `loop`.

## Relevant log output

Case 1, sweeping detector counts 2 through 12 with an identical composition on both sides:

```
n= 2  strict restore OK
n= 3  strict restore OK
...
n=10  strict restore OK
n=11  strict restore RAISED: snapshot detector composition mismatch: snapshot=['loop', 'loop', 'error_cascade', 'latency_anomaly', 'stagnation', 'token_runaway', 'meltdown', 'silent_abort', 'side_effect_guard', 'governance_decay', 'goal_drift'] monitor=['loop', 'error_cascade', 'latency_anomaly', 'stagnation', 'token_runaway', 'meltdown', 'silent_abort', 'side_effect_guard', 'governance_decay', 'goal_drift', 'loop']
        sorted keys: ['0:loop', '10:loop', '1:error_cascade', '2:latency_anomaly', '3:stagnation', '4:token_runaway', '5:meltdown', '6:silent_abort', '7:side_effect_guard', '8:governance_decay', '9:goal_drift']
n=12  strict restore RAISED: snapshot detector composition mismatch: ...
```

Case 1 through `Monitor.default()`, no extras installed:

```
compositions identical: ['loop', 'error_cascade', 'latency_anomaly', 'stagnation', 'goal_drift', 'semantic_goal_drift', 'token_runaway', 'meltdown', 'silent_abort', 'side_effect_guard', 'governance_decay']
RESULT: strict restore RAISED -> snapshot detector composition mismatch: snapshot=['loop', 'governance_decay', 'error_cascade', 'latency_anomaly', 'stagnation', 'goal_drift', 'semantic_goal_drift', 'token_runaway', 'meltdown', 'silent_abort', 'side_effect_guard'] monitor=['loop', 'error_cascade', 'latency_anomaly', 'stagnation', 'goal_drift', 'semantic_goal_drift', 'token_runaway', 'meltdown', 'silent_abort', 'side_effect_guard', 'governance_decay']
tolerant restore OK
```

Case 2:

```
snapshot slots: ['loop', 'error_cascade', 'latency_anomaly', 'stagnation', 'token_runaway', 'meltdown', 'silent_abort', 'side_effect_guard', 'governance_decay', 'goal_drift', 'loop']
monitor  slots: ['loop', 'loop', 'error_cascade', 'latency_anomaly', 'stagnation', 'token_runaway', 'meltdown', 'silent_abort', 'side_effect_guard', 'governance_decay', 'goal_drift']
RESULT: ACCEPTED -- strict mode passed a mismatched composition
        snapshot slot 1 = error_cascade | monitor slot 1 = loop
```

## Impact

* Contradicts the documented contract in `docs/DETECTOR_GUIDE.md` ("Restore is a setup-time operation: a version or strict-composition mismatch raises rather than failing open") and `project.md` §4. The same docs list all eleven shipped detectors as snapshot participants, so the full documented detector set is the configuration that triggers it.
* Case 1 turns an opt-in safety flag into a startup failure for hosts that enable the opt-in detectors — the strictest, most safety-conscious configuration is the one that breaks.
* Case 2 silently voids the guarantee the flag exists to provide.
* No effect on detection, `ingest()`, or the tolerant default `restore(path)` path — only the `strict_names=True` verdict is wrong.

## Suggested fix

Order the snapshot keys by their numeric slot instead of lexicographically, and keep an unparsable slot deterministic so a hand-edited snapshot still cannot raise from inside the comparison.

Adhi1-2:
## Summary

Fixes two defects in `Monitor.restore_dict(strict_names=True)`:

### 1. Lexicographic slot ordering (issue #217)

Snapshot detector keys are `"<slot>:<name>"`. The strict composition check
sorted these keys as plain strings, so slot `10` sorted between `1` and `2`.
With 11+ detectors the check compared two differently-ordered name lists:

- a **matching** composition was rejected (`ValueError` on a valid restore), and
- a **mismatched** composition whose names happened to equal the string-sorted
  key order was accepted.

Eleven slots is not a synthetic number: it is what `Monitor.default()` builds
once the opt-in detectors are enabled, with no extras installed.
`_detector_slot_order()` now recovers the numeric slot order; keys without an
integer prefix sort deterministically after the numbered ones.

### 2. Mutation before validation (found while fixing #217)

The strict check ran **after** the per-key/name-suffix fallback load loop, so a
restore the check went on to reject had already loaded the snapshot's
per-episode windows into the monitor's detectors via the name-suffix fallback.
A caller that catches the documented setup-time `ValueError` was left with a
monitor contaminated by the rejected snapshot — the exact "quiet misbehavior"
the feature's own rationale says it exists to prevent. The check now runs
before any state is applied, so a rejected restore leaves the monitor
untouched. Sink restore already ran after the check; the detector path was the
misordered one.

## Repro (pre-fix)

```python
# Slot ordering: 11 detectors, identical composition -> ValueError
a, b = Monitor(_eleven(), []), Monitor(_eleven(), [])
a.snapshot(path)
b.restore(path, strict_names=True)   # raises "composition mismatch"

# Ordering: mismatched composition -> state applied anyway
a = Monitor([LoopDetector(), ErrorCascadeDetector()], [])
a.ingest(...)  # per-episode windows
a.snapshot(path)
b = Monitor([ErrorCascadeDetector(), LoopDetector()], [])
with pytest.raises(ValueError):
    b.restore(path, strict_names=True)
b._detectors[1]._windows  # non-empty: the rejected snapshot landed
```

## Tests

- `test_strict_names_accepts_identical_11_detector_composition` — false-reject direction
- `test_strict_names_rejects_mismatch_in_lexicographic_key_order` — false-accept direction
- `test_strict_names_survives_key_without_integer_slot_prefix` — unparseable slot stays a `ValueError`
- `test_strict_names_rejection_applies_no_state` — rejected restore applies no state

Full suite, `mypy src tests`, and `ruff check`/`ruff format` all pass locally
(699 passed, 3 skipped).

Fixes #217

### DIFF
--- a/src/snagline/monitor.py
+++ b/src/snagline/monitor.py
@@ -107,6 +107,23 @@ def _detector_key(index: int, detector: Any) -> str:
     return f"{index}:{getattr(detector, 'name', type(detector).__name__)}"
 
 
+def _detector_slot_order(key: str) -> tuple[int, int, str]:
+    """Sort key recovering the numeric slot order of a snapshot detector key.
+
+    ``sorted()`` on the raw ``"<index>:<name>"`` strings is lexicographic, so
+    slot ``10`` lands between ``1`` and ``2``. Any monitor with 11+ detectors
+    then compared two differently-ordered name lists in the ``strict_names``
+    check, rejecting a matching composition and accepting a mismatched one
+    (issue #217). Keys without an integer prefix cannot be placed by slot, so
+    they sort after the numbered ones, deterministically by the raw key.
+    """
+    head, _, _ = key.partition(":")
+    try:
+        return (0, int(head), key)
+    except ValueError:
+        return (1, 0, key)
+
+
 def _auto_calibration_plan(cfg: Config) -> CalibrationPlan | None:
     """Resolve the auto-calibration plan for ``cfg``, fail-open (issue #101).
 
@@ -781,6 +798,10 @@ class Monitor:
         default is tolerant: matching entries are applied by key, missing
         detectors are skipped with a warning, and states without a home are
         ignored with a warning.
+
+        The strict check runs *before* any state is applied, so a rejected
+        restore leaves the monitor untouched: a caller that catches the
+        ``ValueError`` never has to wonder which half of the snapshot landed.
         """
         version = data.get("format_version")
         if version != SNAPSHOT_FORMAT_VERSION:
@@ -789,6 +810,23 @@ class Monitor:
                 f"{SNAPSHOT_FORMAT_VERSION}"
             )
         dumped_detectors: dict[str, Any] = data.get("detectors") or {}
+        # Validate before mutating (strict mode): the composition check below
+        # reads only the snapshot's keys and the monitor's detector list, so
+        # running it first guarantees a rejected restore applies no state at
+        # all. After the loop it would be too late -- the per-key and
+        # name-suffix fallbacks would already have loaded windows into
+        # detectors that the check then rejects the snapshot for.
+        if strict_names:
+            expected = [
+                k.split(":", 1)[1]
+                for k in sorted(dumped_detectors, key=_detector_slot_order)
+            ]
+            current = [getattr(d, "name", type(d).__name__) for d in self._detectors]
+            if expected != current:
+                raise ValueError(
+                    "snapshot detector composition mismatch: "
+                    f"snapshot={expected} monitor={current}"
+                )
         consumed: set[str] = set()
         for i, detector in enumerate(self._detectors):
             load = getattr(detector, "load_state", None)
@@ -827,14 +865,6 @@ class Monitor:
                 "slot(s); ignored",
                 len(orphaned),
             )
-        if strict_names:
-            expected = [k.split(":", 1)[1] for k in sorted(dumped_detectors)]
-            current = [getattr(d, "name", type(d).__name__) for d in self._detectors]
-            if expected != current:
-                raise ValueError(
-                    "snapshot detector composition mismatch: "
-                    f"snapshot={expected} monitor={current}"
-                )
         dumped_sinks: dict[str, Any] = data.get("sinks") or {}
         for i, sink in enumerate(self._sinks):
             load = getattr(sink, "load_state", None)
--- a/tests/test_monitor_snapshot.py
+++ b/tests/test_monitor_snapshot.py
@@ -8,13 +8,18 @@ from typing import cast
 
 import pytest
 
+from snagline.detectors.compaction_tripwire import CompactionTripwireDetector
 from snagline.detectors.error_cascade import ErrorCascadeDetector
+from snagline.detectors.goal_drift import GoalDriftDetector
 from snagline.detectors.latency_anomaly import LatencyAnomalyDetector
 from snagline.detectors.loop import LoopDetector
+from snagline.detectors.meltdown import MeltdownDetector
+from snagline.detectors.side_effect_guard import SideEffectGuardDetector
 from snagline.detectors.silent_abort import SilentAbortDetector
+from snagline.detectors.stagnation import StagnationDetector
 from snagline.detectors.token_runaway import TokenRunawayDetector
 from snagline.events import StepEvent
-from snagline.monitor import Monitor
+from snagline.monitor import SNAPSHOT_FORMAT_VERSION, Monitor
 from snagline.sinks.dedup import DedupSink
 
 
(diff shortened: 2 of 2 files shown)


Cyrax321 (approved):
Verified locally: 11/11 pass in test_monitor_snapshot.py, ruff + mypy clean. Two real fixes: slot-ordered key sort (lexicographic sort misordered 11+ detector compositions) and strict validation moved before any mutation. Note: textually conflicts with #277 on the same hunk - suggest merging this first since its slot-order version supersedes, then #277 rebases to test-only.

PR #250, opened 2026-09-07
fix(adapter): give Claude Code Pre/Post tool events distinct signatures (Fixes #237)

### LINKED ISSUE #237: [adapter] Claude Code PreToolUse and PostToolUse share one action signature, so the loop detector fires on 2 healthy identical calls
## Summary

One logical tool call from Claude Code produces **two** ingested events with the **same** `action_signature`:

- `_EVENT_MAP` maps both `PreToolUse` and `PostToolUse` to `action_type="tool_call"`, differing only in `subkind` (`adapters/claude_code.py:41-49`).
- The signature is built from `tool_name` + `json.dumps(tool_input)` only — `subkind` is never included (`adapters/claude_code.py:146-153`), and the volatile `tool_use_id` (which *would* differ) is deliberately excluded.

`LoopDetector` counts **events** in its window: "if the same signature appears `repeat_threshold` times within `window_size` steps" (`detectors/loop.py:3-5`), default `loop_repeat_threshold=3`, `loop_window_size=12`.

So a perfectly healthy agent that calls the same tool with the same arguments **twice** inside the window contributes 4 identical-signature events and fires `loop` ("action repeated 3x"). The effective threshold is halved: the detector fires on 2 logical calls instead of 3.

Reproduced on `master` (22faeae): ingesting `PreToolUse`+`PostToolUse` for exactly two logical calls (4 payloads, all healthy, `error=False`) through `ingest_payload` + `Monitor.default()` emits `('loop', 'toolu_1', 'action repeated 3x in last 3 steps')`.

The existing test `test_sidecar_accepts_native_claude_code_payloads_and_fires_loop` (`tests/test_hook_bridge.py:174`) posts 4 pairs and only asserts `any(trigger == "loop")`, so it cannot distinguish firing at 4 logical calls (intended) from 2 (bug). The doubling flows through all three paths: `snagline hook`, `snagline watch --file`, and the sidecar `/hooks/claude-code` endpoint.

## Why it matters

Re-running the same tool with the same input twice is normal, healthy agent behavior (retry after an empty result, a deliberate second read). The Claude Code bridge is the flagship integration; halving the loop threshold on it produces false `loop` risks on the default composition, and downstream `EnforcementPolicy` consumers route by trigger name.

## Reproduction

```python
from snagline.adapters.claude_code import HookTracker, ingest_payload

m = Monitor.default(sinks=[])
tracker = HookTracker()
tool_input = {"path": "/tmp/x", "content": "same"}

for call in ("toolu_1", "toolu_2"):          # exactly TWO logical calls
    for event in ("PreToolUse", "PostToolUse"):
        ingest_payload(m, {"hook_event_name": event, "session_id": "s",
                           "tool_use_id": call, "tool_name": "Edit",
                           "tool_input": tool_input}, tracker)

# m._sinks[0] (ListSink) now holds a 'loop' risk: fired on 2 logical calls
```

Fix shape: include the `start`/`end` subkind in the tool-call signature (or otherwise stop the Pre event from feeding the loop window), so one logical attempt contributes one countable occurrence.

Adhi1-2:
## Summary

One logical Claude Code tool call emits two hook payloads — `PreToolUse` and
`PostToolUse` — with the same `tool_name` and `tool_input`. The signature was
built from exactly those (plus `action_type`), so both events hashed
identically: only the volatile `tool_use_id` differs, and the signature
deliberately excludes it.

`LoopDetector` counts **events** in its window, so one logical attempt
contributed 2 identical signatures. With the default `loop_repeat_threshold=3`,
a healthy agent that runs the same tool with the same arguments **twice**
fired `loop` ("action repeated 3x in last 3 steps") — the effective threshold
was halved for every Claude Code host (issue #237).

The fix: the tool-call signature includes the `start`/`end` subkind as an
extra stable part.

- Pre and Post of one logical call now have distinct signatures.
- Retries of either kind still hash identically to their own kind, so loop
  detection of genuine repetition is unaffected.
- The message path (`UserPromptSubmit` etc.) is unchanged.

## Tests

- `test_two_healthy_logical_calls_do_not_trip_loop` — 2 Pre/Post pairs (the
  false-positive repro from #237) through the real `Monitor.default()`.
- `test_third_logical_call_still_trips_loop` — guards against
  over-suppression: 3 identical logical attempts still fire.
- `test_pre_and_post_have_distinct_signatures` — signature-level check,
  including retry pairability.

The existing `test_hook_latency_reaches_the_cusum_detector` (10 Pre/Post
pairs) and `test_repeated_same_tool_input_is_loop_detectable` (4 Pre events)
continue to pass, confirming no collateral behavior change.

Full suite, `mypy src tests`, and `ruff check`/`ruff format` all pass locally
(698 passed, 3 skipped).

Fixes #237

### DIFF
--- a/src/snagline/adapters/claude_code.py
+++ b/src/snagline/adapters/claude_code.py
@@ -143,14 +143,21 @@ def payload_to_event(
 
     # Stable, structural parts only: the same logical attempt (same tool,
     # same input) must hash identically for loop detection to see a retry.
+    # The start/end subkind is included for tool calls: Claude Code fires
+    # PreToolUse and PostToolUse for ONE logical call with the same tool and
+    # input, and a signature blind to the subkind made that single attempt
+    # count twice in the loop window (issue #237), halving the effective
+    # repeat_threshold for every Claude Code host.
     if is_tool:
         stable = json.dumps(payload.get("tool_input") or {}, sort_keys=True)
         tool_for_sig = str(tool_name or "tool")
+        stable_parts: tuple[str, ...] = (subkind,)
     else:
         stable = str(payload.get("prompt") or payload.get("error") or subkind)
         # prompt_id is volatile (unique per turn): excluding it from the
         # signature lets a repeated identical prompt be loop-detectable.
         tool_for_sig = "claude-code"
+        stable_parts = ()
 
     error = payload.get("hook_event_name") in _ERROR_EVENTS
     error_type = None
@@ -176,7 +183,9 @@ def payload_to_event(
         episode_id=episode_id,
         timestamp=timestamp if timestamp is not None else time.time(),
         action_type=action_type,
-        action_signature=make_signature(action_type, tool_for_sig, stable),
+        action_signature=make_signature(
+            action_type, tool_for_sig, stable, *stable_parts
+        ),
         tool_name=str(tool_name) if tool_name is not None else None,
         latency_ms=latency_ms,
         error=error,
--- a/tests/adapters/test_claude_code_adapter.py
+++ b/tests/adapters/test_claude_code_adapter.py
@@ -222,3 +222,60 @@ def test_ingest_payload_never_raises() -> None:
         m,
         {"hook_event_name": "PreToolUse", "session_id": None, "tool_input": {}},  # type: ignore[dict-item]
     )
+
+
+def test_two_healthy_logical_calls_do_not_trip_loop() -> None:
+    # Issue #237: PreToolUse and PostToolUse of ONE logical call used to share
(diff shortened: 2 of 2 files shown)


Cyrax321 (approved):
Verified locally: 17/17 pass in test_claude_code_adapter.py, ruff + mypy clean. Including the subkind in the signature fixes the Pre/Post double-count (halved repeat_threshold for Claude Code hosts), and the tests pin both directions - 2 logical calls quiet, 3 still fire, retries stay pairable. Note: textually conflicts with #258 in claude_code.py - suggest merging this first, then rebasing #258.

PR #251, opened 2026-09-07
fix(server): atomic batch ingest, path-component routing, serve --config layering (Fixes #239, #240, #241)

### LINKED ISSUE #239: [server] Batched POST /events ingests a prefix of the batch before rejecting it with 400, so a retrying client re-ingests every accepted event
## Summary

`_post_events` validates and ingests batch items **in the same loop** (`server/http_server.py:687-703`):

```python
for item in obj:
    if not isinstance(item, dict):
        self._respond(400, {"error": "invalid StepEvent in batch"})
        return
    try:
        event = StepEvent(**item)
    except (ValueError, TypeError):
        self._respond(400, {"error": "invalid StepEvent in batch"})
        return
    self._ingest_recorded(event)      # items 0..k-1 are already in
    count += 1
```

When item k fails validation, the handler answers `400` — but items `0..k-1` have already been through `monitor.ingest()` and `collector.record_ingest()`. `StepEvent` is a plain frozen dataclass with no coercion (`events.py:19-50`), so a `TypeError` on an unknown key or missing field is the only rejection path, and it can land anywhere in the batch.

A well-behaved client that treats 400 as "fix and resend" (or a host that simply retries on error) re-ingests the accepted prefix every time. Verified live on `master` (22faeae): `POST /events` with `[good, bad-unknown-key]` returned 400 while the good event was ingested; three retries → the good event ingested three times.

Wrong outcomes:
- duplicated steps feed the loop / error-cascade / CUSUM detectors (false positives — e.g. a duplicated tool call pushes a healthy 2-call pattern to 4 identical signatures);
- `snagline_events_total` and `snagline_ingest_seconds` overcount;
- the client has no way to know part of the batch was accepted.

Single-object POSTs are unaffected.

## Why it matters

The batch endpoint exists for high-throughput hosts buffering telemetry (`http_server.py:688-690`), exactly the deployments most likely to have retry logic. A 400 that silently half-succeeds violates the all-or-nothing expectation an HTTP status code carries.

## Reproduction

```python
import json, urllib.request
req = urllib.request.Request(
    "http://127.0.0.1:PORT/events",
    data=json.dumps([
        {"step_id": "s1", "episode_id": "ep", "timestamp": 0.0,
         "action_type": "tool_call", "action_signature": "a"},
        {"step_id": "s2", "unknown_field": True},   # rejected
    ]).encode(),
    headers={"Content-Type": "application/json"}, method="POST")
urllib.request.urlopen(req)          # -> HTTP 400
# GET /metrics: snagline_events_total already counts s1
```

Fix shape: two passes — validate and construct the whole batch first, ingest only after every item parsed.

### LINKED ISSUE #240: [server] snagline serve --config silently ignores config-file values for metrics_format, server_read_timeout, and episode_ttl_seconds
## Summary

`_cmd_serve` resolves the full layered config via `Config.resolve(path=args.config)` (`cli.py:976+`) and uses it for sinks/policy — but forwards only the CLI flags into `serve()`:

```python
# cli.py:1053-1054
read_timeout=args.read_timeout,          # None unless --read-timeout given
episode_ttl_seconds=args.episode_ttl_seconds,   # None unless --episode-ttl given
```

`make_handler`'s three `_resolve_*` helpers then re-resolve from **`os.environ` only** (`Config.from_env_overrides()`, `config.py:477-508`) and fall back to bare `Config()` defaults (`http_server.py:145, 172, 215`):

```python
candidate = explicit                       # CLI flag
if candidate is None:
    candidate = Config.from_env_overrides().get("server_read_timeout")  # env only
if candidate is None:
    candidate = Config().server_read_timeout                              # default
```

So the **file layer of `Config.resolve` is dropped** for exactly these three knobs:
- `"server_read_timeout": 5` in the config file → timeout stays 30 s
- `"episode_ttl_seconds": 2` → TTL stays disabled
- `"metrics_format": "classic"` → sidecar still serves Prometheus

Verified end-to-end on `master` (22faeae): `snagline serve --config cfg.json` with `metrics_format: classic` served the Prometheus body; `SNAGLINE_METRICS_FORMAT=classic` served the JSON body.

This contradicts the CLI's own help text (`cli.py:314`: *"to $SNAGLINE_SERVER_READ_TIMEOUT **or `Config.server_read_timeout`**"*, `cli.py:323`: *"**`Config.episode_ttl_seconds`** [default: disabled]"*) and `Config.resolve`'s contract as the single consistent entrypoint (config.py:554).

## Why it matters

The safety-relevant case is `server_read_timeout`: an operator who tightens the stalled-sender protection from #130 via the config file believes it is in effect when it is not. Same for the TTL from #173: a deployment that expects active-episode gauges to expire keeps them forever.

## Reproduction

```json
// cfg.json
{"server_read_timeout": 5, "episode_ttl_seconds": 2, "metrics_format": "classic"}
```
```console
$ snagline serve --config cfg.json &
$ curl localhost:PORT/metrics | head -1   # Prometheus format, not JSON "classic"
```

Fix shape: in `_cmd_serve`, fall back to the resolved `cfg` values when the CLI flags are unset — `read_timeout=args.read_timeout if args.read_timeout is not None else cfg.server_read_timeout` (and likewise for `episode_ttl_seconds`, `metrics_format`) — so all three sources resolve through one layering.

### LINKED ISSUE #241: [server] Exact-match routing breaks /health and every POST route when a query string is present (/health?probe=1 answers 401 with a token set)
## Summary

`self.path` includes the query string, and the `/health` check plus the whole POST route chain compare it **exactly** (`server/http_server.py:556, 638-644`):

```python
if self.path == "/health":            # 556
...
if self.path == "/events":            # 638
elif self.path == "/hooks/claude-code":
elif self.path == "/episodes/end":
elif self.path == "/risks":
else:
    self._respond(404, {"error": "not found"})
```

Meanwhile the GET routes correctly strip the query first (`http_server.py:564-569`):

```python
split = urlsplit(self.path)
if split.path == "/metrics":
    self._serve_metrics(parse_qs(split.query))
```

The inconsistency shows the omission is accidental. Verified live on `master` (22faeae):

- `GET /health` → 200
- `GET /health?probe=1` → falls through the open `/health` branch, hits the auth gate, → **401 with a token set** (404 without)
- `POST /events?batch=1` → 404
- `GET /metrics?format=classic` → 200 (the only route family that parses queries)

## Why it matters

- A k8s/ELB liveness probe that appends any query parameter gets 401 from an endpoint whose whole design purpose is to be probe-friendly (`http_server.py:553-555`).
- Any hook forwarder or batch client that appends a query param gets a 404 on a live endpoint.
- The `/health` case is worse than a plain 404: with `auth_token` set it answers **401**, which most probes interpret as "auth misconfigured" and sends operators down the wrong debugging path.

## Reproduction

```console
$ snagline serve --auth-token secret &
$ curl -i "http://127.0.0.1:PORT/health?probe=1"
HTTP/1.1 401 Unauthorized            # expected: 200 {"status": "ok"}
$ curl -i -X POST -H "Authorization: Bearer secret" \
    "http://127.0.0.1:PORT/events?batch=1" -d '[...]'
HTTP/1.1 404 Not Found               # expected: 202
```

Fix shape: parse once at the top of `do_GET`/`do_POST` (`urlsplit(self.path).path`) and route on the path component, matching the existing `/metrics` handling.

Adhi1-2:
## Summary

Three sidecar fixes, all in the request path. Each is small, they were found in
the same audit pass, and they share the test harness, so one PR rather than
three.

### 1. Batched `POST /events` is now atomic (Fixes #239)

The handler validated and ingested batch items in the same loop, so a bad item
at position k answered `400` with items 0..k-1 **already ingested**. A
well-behaved client that fixes and retries re-fed the accepted prefix on every
attempt: duplicated steps fed the loop/cascade/CUSUM detectors (false
positives), and `snagline_events_total` / `snagline_ingest_seconds`
double-counted. The loop now constructs the full batch first and only then
ingests — a rejected batch ingests zero events.

### 2. Routing uses the path component (Fixes #241)

`self.path` includes the query string. The `do_POST` route chain and the open
`/health` check compared it exactly, so:

- `POST /events?batch=1` → 404 on a live endpoint
- `GET /health?probe=1` → fell through the open `/health` branch to the auth
  gate → **401 with a token set** — from the endpoint whose whole design
  purpose is answering liveness probes that cannot carry secrets

The GET side already stripped the query via `urlsplit` for `/metrics`; `do_POST`
and `/health` now do the same, and unknown paths with queries still 404.

### 3. `serve --config` values reach the server (Fixes #240)

`_cmd_serve` resolves the full layered `Config` (file → env → defaults) but
forwarded only the CLI flags for `metrics_format`, `server_read_timeout`, and
`episode_ttl_seconds`. The `_resolve_*` helpers inside `make_handler`
re-resolve from `os.environ` only, so a **config-file** value for those three
knobs was silently dropped — the CLI help text explicitly promises
`Config.server_read_timeout` / `Config.episode_ttl_seconds` precedence. The
resolved `cfg` values are now forwarded when the flags are unset (flags keep
priority, preserving the resolve contract), and `metrics_format` now flows
through the CLI at all.

## Tests

New `tests/server/test_routing_batch_config.py` (8 tests): rejected-batch
ingests nothing / valid batch unchanged / retry-after-400 ingests exactly once;
query-string POST and `/health` routing; unknown path with query still 404;
`_cmd_serve` forwards layered config values; flag beats config file.
`test_main_serve_starts_server` updated for the now-forwarded defaults.

Full suite, `mypy src tests`, `ruff check`/`ruff format` pass locally
(703 passed, 3 skipped).

Fixes #239
Fixes #240
Fixes #241

### DIFF
--- a/src/snagline/cli.py
+++ b/src/snagline/cli.py
@@ -1050,8 +1050,22 @@ def _cmd_serve(args: argparse.Namespace) -> int:
             auth_token=auth_token,
             max_body_bytes=args.max_body_bytes,
             max_risks=args.max_risks,
-            read_timeout=args.read_timeout,
-            episode_ttl_seconds=args.episode_ttl_seconds,
+            metrics_format=cfg.metrics_format,
+            # Flag, else the layered config (issue #240): the _resolve_*
+            # helpers in http_server re-resolve from env only, so a value
+            # that came from the --config file was silently dropped for
+            # exactly these three knobs. Passing the resolved cfg value
+            # when the flag is unset keeps one layering everywhere.
+            read_timeout=(
+                args.read_timeout
+                if args.read_timeout is not None
+                else cfg.server_read_timeout
+            ),
+            episode_ttl_seconds=(
+                args.episode_ttl_seconds
+                if args.episode_ttl_seconds is not None
+                else cfg.episode_ttl_seconds
+            ),
             **tls_kwargs,
         )
     return 0
--- a/src/snagline/server/http_server.py
+++ b/src/snagline/server/http_server.py
@@ -553,7 +553,12 @@ def make_handler(
             # /health is deliberately open: a liveness probe (k8s, ELB, docker
             # healthcheck) generally cannot be taught to carry a shared secret,
             # and it reveals nothing but reachability.
-            if self.path == "/health":
+            # Route on the path component only (issue #241): /health is
+            # deliberately open, and a probe that appends a query string
+            # (/health?probe=1) used to miss this branch and land behind
+            # the auth gate -- 401 from an endpoint meant to answer 200
+            # for anything that can reach it.
+            if urlsplit(self.path).path == "/health":
                 self._respond(200, {"status": "ok"})
                 return
             # Everything else is behind the token, including the 404 fallthrough
@@ -635,13 +640,18 @@ def make_handler(
                 self._discard_overcap_body(length)
                 self._respond(413, {"error": "payload too large"})
                 return
-            if self.path == "/events":
+            # Route on the path component only (issue #241): self.path keeps
+            # any query string, so exact comparisons 404'd /events?batch=1
+            # and let /health?probe=1 fall through to the auth gate. The GET
+            # side already strips the query via urlsplit; POST now does too.
+            route = urlsplit(self.path).path
+            if route == "/events":
                 self._post_events()
-            elif self.path == "/hooks/claude-code":
+            elif route == "/hooks/claude-code":
                 self._post_claude_hook()
-            elif self.path == "/episodes/end":
+            elif route == "/episodes/end":
                 self._post_episodes_end()
-            elif self.path == "/risks":
+            elif route == "/risks":
                 self._post_risks()
             else:
                 self._discard_overcap_body(length)
@@ -688,19 +698,26 @@ def make_handler(
                 # Batched ingestion: a JSON array of StepEvent objects. This
                 # lets a host buffer telemetry and flush many steps in one
                 # request (useful for high-throughput or offline replay).
-                count = 0
+                # Two passes (issue #239): validate and construct the whole
+                # batch before ingesting anything. Ingesting inline meant a
+                # bad item at position k answered 400 with items 0..k-1
+                # already inside the Monitor, so a retrying client re-fed
+                # the accepted prefix on every attempt -- duplicated steps
+                # for the loop/cascade/CUSUM detectors and double-counted
+                # metrics.
+                events: list[StepEvent] = []
                 for item in obj:
                     if not isinstance(item, dict):
                         self._respond(400, {"error": "invalid StepEvent in batch"})
                         return
                     try:
-                        event = StepEvent(**item)
+                        events.append(StepEvent(**item))
                     except (ValueError, TypeError):
                         self._respond(400, {"error": "invalid StepEvent in batch"})
                         return
+                for event in events:
                     self._ingest_recorded(event)
-                    count += 1
-                self._respond(202, {"status": "ingested", "count": count})
+                self._respond(202, {"status": "ingested", "count": len(events)})
                 return
             if not isinstance(obj, dict):
                 self._respond(
--- a/tests/server/test_routing_batch_config.py
+++ b/tests/server/test_routing_batch_config.py
@@ -1,0 +1,300 @@
+"""Server routing, batch atomicity, and serve-config layering.
+
+Three related defects in the sidecar:
+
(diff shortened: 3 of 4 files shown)


Cyrax321 (approved):
Verified locally: 24/24 pass across server/test_routing_batch_config.py + test_cli.py, ruff + mypy clean. All three are real: validate-then-ingest batch atomicity (no accepted-prefix re-feed on retry), path-component routing (query strings no longer 404/401), and serve --config actually forwarding metrics_format/read_timeout/ttl. Bundled coherently with a thorough 300-line test module. Safe to merge.

PR #252, opened 2026-09-07
fix(core): keep MemoryStateBackend.release() from orphaning parked waiters (Fixes #238)

### LINKED ISSUE #238: [core] MemoryStateBackend.release() orphans parked waiters: two threads can hold one episode's critical section after end_episode
## Summary

`MemoryStateBackend.episode_lock` fetches the lock under `_meta`, then releases `_meta` **before** parking on the lock (`state.py:56-64`):

```python
with self._meta:
    lock = self._locks.get(episode_id)
    if lock is None:
        lock = threading.RLock()
        self._locks[episode_id] = lock
with lock:            # <- thread parks HERE, holding a reference
    yield
```

`release()` pops the dict entry while a waiter may already hold a reference to the old `RLock` and be parked on it (`state.py:79-80`). Sequence:

1. Thread A runs `end_episode("ep")` and holds L1 (inside `finalize`).
2. Thread B calls `ingest` for the same episode, fetches L1 under `_meta`, blocks on `with lock:`.
3. A calls `self._state.release("ep")` — pops L1 from `_locks` — then returns, releasing L1.
4. B wakes on the orphaned L1 and **enters the critical section**.
5. Thread C ingests for `"ep"`, finds no entry, allocates a fresh L2 and enters **concurrently with B**.

Result: concurrent `detector.observe()` / `_advance_clock()` mutations of the same episode's state with no mutual exclusion — the exact serialization `episode_lock` exists to guarantee. Secondary effect: B observes an episode that was already finalized/reset, re-creating detector state that no future `end_episode` will clean.

Both in-code safety claims are false for the parked-waiter window: `monitor.py:730-731` ("no other thread can be in this episode's critical section, so dropping the entry cannot strand a holder") and the `release()` docstring (`state.py:73-77`).

Reproduced on `master` (22faeae) through the real `Monitor` path with a `finalize` that sleeps: **max concurrent observers inside one episode: 2** (thread B parked on the old lock while `end_episode` released; thread C then entered on the fresh lock; both ran `observe` simultaneously). Also reproduced at the primitive level.

This is distinct from #67 (the *leak*: `release` never existed) — this is a race *inside the release mechanism itself*.

## Why it matters

`end_episode` concurrent with a late `ingest` for the same episode is a normal sidecar pattern (worker draining a batch while the episode TTL fires). The race silently corrupts per-episode detector state (window/CUSUM mutations under two threads) with no exception anywhere.

## Reproduction sketch

```python
import threading, time

class SlowDetector(LoopDetector):
    def finalize(self, episode_id):
        time.sleep(0.2)
        return super().finalize(episode_id)

m = Monitor([SlowDetector()], [])
m.ingest(ev("ep", 0))                      # ensure state exists
t_end = threading.Thread(target=lambda: m.end_episode("ep"))
t_end.start(); time.sleep(0.05)            # A inside finalize, holding L1
parked = threading.Thread(target=lambda: m.ingest(ev("ep", 1)))
parked.start(); time.sleep(0.05)           # B parked on L1
t_end.join(); parked.join()                # release popped L1; B entered on it
m.ingest(ev("ep", 2))                      # C allocates L2, runs beside B
```

Fix shape: a waiter refcount — `episode_lock` increments a per-episode holder count under `_meta`; `release()` pops immediately only when the caller is the sole holder, otherwise marks the entry for deferred removal and the last holder cleans it up. Mutual exclusion and the no-leak guarantee from #67 both hold.

Adhi1-2:
## Summary

`MemoryStateBackend.episode_lock` fetched the lock under `_meta`, then released
`_meta` **before** parking on the lock. `release()` popped the dict entry while
a waiter could already hold a reference to the old `RLock` and be parked on it.
Sequence:

1. Thread A runs `end_episode("ep")` and holds L1 inside `finalize`.
2. Thread B calls `ingest` for the same episode, fetches L1 under `_meta`,
   blocks on `with lock:`.
3. A calls `release("ep")` — popped L1 — then returns, releasing L1.
4. B wakes on the orphaned L1 and **enters the critical section**.
5. Thread C ingests for `"ep"`, finds no entry, allocates a fresh L2 and enters
   **concurrently with B**.

Result: concurrent `detector.observe()` / `_advance_clock()` mutations of one
episode's state with no mutual exclusion — the exact serialization
`episode_lock` exists to guarantee — with no exception anywhere. Both in-code
safety claims (the `release()` docstring and the comment in
`Monitor.end_episode`) were false for the parked-waiter window. Reproduced
through the real `Monitor` path: max concurrent observers inside one episode
reached 2 (issue #238).

### Fix

The lock table now holds `_LockEntry` objects — `lock`, a `waiters` count, and
a `released` flag:

- `episode_lock` increments `waiters` under `_meta` *before* parking, and
  decrements in a `finally`; the last waiter to drain removes a released entry.
- `release()` marks the entry instead of popping it (an entry with zero
  waiters is still popped immediately, preserving the no-leak guarantee from
  #67).

A parked waiter and any later fetcher therefore serialize on the *same* entry
until everyone has drained; only then does the episode's entry disappear.

## Tests

- `test_parked_waiter_keeps_exclusive_section_after_release` — the exact race
  from #238 made deterministic with a gated `finalize`/`observe`: asserts B and
  C are counted on one shared entry (`waiters >= 2`), and the entry drains
  after the last waiter.
- `test_release_during_parked_waiter_blocks_new_entry_until_drain` — the same
  property directly at the primitive.

Both tests fail against the pre-fix `state.py` and pass with the fix. The
existing suite (including #67's leak guard, `test_end_episode_releases_the_episode_lock`)
continues to pass. Full suite, `mypy src tests`, `ruff check`/`ruff format`
pass locally (697 passed, 3 skipped).

Fixes #238

### DIFF
--- a/src/snagline/state.py
+++ b/src/snagline/state.py
@@ -46,22 +46,49 @@ class ReleasableStateBackend(StateBackend, Protocol):
         ...
 
 
+class _LockEntry:
+    """One episode's lock plus the bookkeeping that keeps it safe to drop.
+
+    ``waiters`` counts threads that have fetched this entry -- holding its
+    lock or parked waiting for it. ``released`` marks an episode whose lock
+    the backend has been asked to drop. The entry leaves the table only
+    once every waiter has drained, so a thread parked on the lock when
+    ``release()`` ran can never end up inside the critical section beside
+    a fetcher that arrived afterwards: both serialize on this same entry
+    until the last one exits and the entry is finally removed (issue #238).
+    """
+
+    __slots__ = ("lock", "waiters", "released")
+
+    def __init__(self) -> None:
+        self.lock = threading.RLock()
+        self.waiters: int = 0
+        self.released: bool = False
+
+
 class MemoryStateBackend:
     """Process-local backend: one re-entrant lock per episode id."""
 
     def __init__(self) -> None:
         self._meta = threading.Lock()
-        self._locks: dict[str, threading.RLock] = {}
+        self._locks: dict[str, _LockEntry] = {}
 
     @contextmanager
     def episode_lock(self, episode_id: str) -> Iterator[None]:
         with self._meta:
-            lock = self._locks.get(episode_id)
-            if lock is None:
-                lock = threading.RLock()
-                self._locks[episode_id] = lock
-        with lock:
-            yield
+            entry = self._locks.get(episode_id)
+            if entry is None:
+                entry = _LockEntry()
+                self._locks[episode_id] = entry
+            entry.waiters += 1
+        try:
+            with entry.lock:
+                yield
+        finally:
+            with self._meta:
+                entry.waiters -= 1
+                if entry.released and entry.waiters == 0:
+                    self._locks.pop(episode_id, None)
 
     def release(self, episode_id: str) -> None:
         """Drop the lock allocated for a finished episode.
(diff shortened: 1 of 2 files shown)


Cyrax321 (approved):
Verified locally: 12/12 pass in test_state_backend.py, ruff + mypy clean. Waiter-counted entries keep a parked waiter and later fetchers on the same lock until drain, fixing the orphaned-lock race where two threads could run one episode's critical section concurrently. The release() doc update matches the new semantics. Safe to merge.

PR #253, opened 2026-09-07
fix(core): let FailureRisk keep an explicitly requested severity of warning (Fixes #246)

### LINKED ISSUE #246: [core] FailureRisk cannot be constructed with severity="warning": the default sentinel collides with a legitimate value and silently re-routes alerts
## Summary

`FailureRisk.severity` defaults to `SEVERITY_WARNING`, and `__post_init__` treats **any** `severity == "warning"` as "unset", overwriting it with `severity_from_score(score)` (`risk.py:64-71`):

```python
severity: str = SEVERITY_WARNING  # auto-derived from score if left default

def __post_init__(self) -> None:
    # If the caller did not set an explicit severity, derive one from the
    # score so every risk carries a useful routing hint. A caller that
    # passes `severity=` explicitly keeps that value.
    if self.severity == SEVERITY_WARNING:
        object.__setattr__(self, "severity", severity_from_score(self.score))
```

The comment's promise is false: an explicitly requested `"warning"` is indistinguishable from the default, so:

- `FailureRisk('ep','s',0.95,'loop','d',0.0, severity="warning").severity == "critical"`
- `FailureRisk('ep','s',0.30,'loop','d',0.0, severity="warning").severity == "info"`

Severity is load-bearing downstream: `SlackSink` and `PagerDutySink` filter on `min_severity`, and `DedupSink`'s default key includes `risk.severity`. A detector or host that deliberately downgrades a high-score risk to `"warning"` (e.g. a known-benign pattern) still **pages on-call**; one that marks a low-score risk `"warning"` has it **dropped** by a `min_severity="warning"` filter.

Verified on `master` (22faeae) by direct construction. No shipped detector currently passes `severity="warning"` explicitly, but `severity` is a public dataclass field that hosts and third-party detectors construct (the README documents the risk contract).

## Why it matters

The severity override is the documented escalation-control surface; silently re-deriving it defeats every deliberate downgrade/upgrade a host makes for exactly one of the three legal values. The fix is a distinct sentinel (`severity: str | None = None`, derive only when `None`), after which the comment becomes true.

## Reproduction

```python
FailureRisk("ep", "s", 0.95, "loop", "d", 0.0, severity="warning").severity
# 'critical'  (requested: 'warning')
FailureRisk("ep", "s", 0.30, "loop", "d", 0.0, severity="warning").severity
# 'info'      (requested: 'warning')
```

Adhi1-2:
## Summary

`FailureRisk.severity` defaulted to `SEVERITY_WARNING`, and `__post_init__`
treated **any** `severity == "warning"` as "unset", overwriting it with the
score-derived value. An explicitly requested `"warning"` was therefore
indistinguishable from the default:

- `FailureRisk("ep","s",0.95,"loop","d",0.0, severity="warning").severity`
  became `"critical"` — a deliberate downgrade of a high-score risk still
  **pages on-call** through `min_severity` filters, against the caller's intent
- the same at score 0.30 became `"info"` — **dropped** by a
  `min_severity="warning"` filter

Severity is load-bearing downstream: `SlackSink`/`PagerDutySink` route on
`min_severity`, `DedupSink`'s default key includes it, and the Prometheus
renderer labels by it. The `__post_init__` comment claimed explicit severities
were kept; for one of the three legal values it never was (issue #246).

### Fix

The default is now `""` — not a legal severity value, so it can only mean
"unset" — and derivation happens only for an empty value. All three legal
values survive construction literally. The field's type stays `str`, so no
consumer signature changes.

## Tests

`test_risk_explicit_warning_survives` pins all three score bands with an
explicit `severity="warning"`. The existing
`test_risk_default_severity_derived_from_score` / `test_risk_explicit_severity_kept`
suite continues to pass unchanged (no shipped detector passes `severity=`
explicitly, so no detector behavior shifts).

Full suite, `mypy src tests`, `ruff check`/`ruff format` pass locally
(696 passed, 3 skipped).

Fixes #246

### DIFF
--- a/src/snagline/risk.py
+++ b/src/snagline/risk.py
@@ -61,11 +61,18 @@ class FailureRisk:
     trigger: TriggerType
     detail: str  # short human-readable explanation, no raw content
     timestamp: float
-    severity: str = SEVERITY_WARNING  # auto-derived from score if left default
+    # "" means "unset": derived from score in __post_init__. A literal
+    # severity, including "warning", survives construction exactly as the
+    # caller requested (issue #246). The default cannot use the string
+    # "warning" itself -- it would be indistinguishable from an explicit
+    # "warning" and silently re-derived, re-routing alerts in
+    # min_severity-filtered sinks. "" is not a legal severity value, so it
+    # can only mean "unset".
+    severity: str = ""
 
     def __post_init__(self) -> None:
-        # If the caller did not set an explicit severity, derive one from the
-        # score so every risk carries a useful routing hint. A caller that
-        # passes `severity=` explicitly keeps that value.
-        if self.severity == SEVERITY_WARNING:
+        # Derive a routing hint from the score only when the caller left
+        # severity unset. A caller that passes `severity=` explicitly keeps
+        # that value -- all three legal values, warning included.
+        if not self.severity:
             object.__setattr__(self, "severity", severity_from_score(self.score))
--- a/tests/test_sinks_dedup.py
+++ b/tests/test_sinks_dedup.py
@@ -44,6 +44,18 @@ def test_risk_explicit_severity_kept():
     assert _risk(0.9, severity=SEVERITY_INFO).severity == SEVERITY_INFO
 
 
+def test_risk_explicit_warning_survives():
+    # Issue #246: the default used to BE "warning", so an explicitly
+    # requested warning was indistinguishable from "unset" and silently
+    # re-derived from the score -- becoming "critical" above 0.8 (paging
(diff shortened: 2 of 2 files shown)


Cyrax321 (approved):
Fix is correct and backward compatible: every FailureRisk construction in src/ omits severity, so default derived behavior is unchanged and only an explicit warning is now kept. 20/20 pass in test_sinks_dedup.py, ruff clean. One nit (non-blocking): the new test comment says 'The None default keeps...' but the implementation uses an empty-string default, not None.

PR #254, opened 2026-09-07
fix(detector): re-arm GoalDriftDetector after recovery

### LINKED ISSUE #247: [detector] GoalDriftDetector latches once per episode: a second, independent drift in the same long-lived episode is never reported
## Summary

`GoalDriftDetector.observe()` sets `self._fired[episode_id] = True` on the first threshold crossing and never re-arms — not even after the drift score returns below `goal_drift_score_threshold` (`detectors/goal_drift.py:55-57`):

```python
if score < self._cfg.goal_drift_score_threshold:
    return None
if self._fired.get(event.episode_id, False):
    return None                      # latched forever, score is never checked again
self._fired[event.episode_id] = True
return FailureRisk(...)
```

Every other shipped detector re-arms when its condition clears: loop (clears `_fired` when the signature drops out of the window), error cascade, meltdown, the ESN CUSUM (`esn_ensemble.py:225` "re-arm so a persistent fault re-alarms later"), and the cascade module's own docstring (`error_cascade.py:103-110`) makes exactly the long-lived-episode argument. Only `reset()` (episode end) clears the latch — but per #184, hosts that never call `end_episode` are a supported deployment.

Verified on `master` (22faeae): baseline of 100ms/0% errors; drift phase 1 (5000ms latencies) fires at step 4; 32 healthy steps bring the score back to 0; drift phase 2 (20000ms latencies + errors) — **no risk is ever emitted**; `_fired["ep"]` stays `True`.

## Why it matters

A long-lived episode (a daemon-style agent, a session that outlives its drift) that recovers from one drift and later hits a *different, worse* one gets silence from the detector whose entire job is that comparison. The single alert also carries no way to distinguish "drifted once, recovered" from "still drifted".

## Reproduction sketch

```python
d = GoalDriftDetector(baseline=healthy_profile)
d.observe(drift_1_events)    # fires goal_drift
d.observe(healthy_events)    # score back under threshold
d.observe(drift_2_events)    # returns None forever
```

Fix shape: clear `_fired[episode_id]` when the score drops back below the threshold (mirroring loop/cascade re-arm semantics), so each *distinct* drift phase alerts once.

Adhi1-2:
## Summary

`GoalDriftDetector.observe()` latched `_fired[episode_id]` on the first threshold crossing and never cleared it — not even after the drift score dropped back below `goal_drift_score_threshold`. Only `reset()` (episode end) cleared the latch, and per #184 hosts that never call `end_episode` are a supported deployment, so a long-lived episode that recovered from one drift and later hit a second, independent one stayed silent forever.

The latch now clears when the score falls back under the threshold, mirroring the re-arm semantics of every other shipped detector (loop re-arms when the signature leaves the window; the ESN CUSUM resets to re-alarm on persistent faults). While the drift persists without an intervening recovery, the once-per-phase dedupe behavior is unchanged.

## Why it matters

A monitoring library that only ever fires once per episode for a given trigger is fail-open in the wrong direction: a recovered-then-re-drifted run is exactly the "flaky then truly stuck" pattern the detector exists to catch.

## Tests

- `test_goal_drift_rearms_after_recovery` — phase-1 drift fires once, 40 healthy steps re-arm, phase-3 drift fires again (fails on pre-fix code).
- `test_goal_drift_still_dedupes_while_drift_persists` — 10 continuously drifted events still produce exactly one risk.

Full gates pass locally: `pytest -q` (697 passed, 87.48% coverage), `mypy src tests`, `ruff check`, `ruff format --check`.

Fixes #247

### DIFF
--- a/src/snagline/detectors/goal_drift.py
+++ b/src/snagline/detectors/goal_drift.py
@@ -51,6 +51,14 @@ class GoalDriftDetector:
             return None
         score = self._drift_score(event.episode_id, baseline)
         if score < self._cfg.goal_drift_score_threshold:
+            # Below threshold: the episode has recovered from whatever fired
+            # before, so re-arm. Without this the latch held forever (issue
+            # #247) -- a long-lived episode that recovered from one drift and
+            # later hit a second, independent one never alerted again. This
+            # mirrors the re-arm semantics every other shipped detector has
+            # (loop clears when the signature leaves the window; the ESN
+            # CUSUM resets to re-alarm on persistent faults).
+            self._fired[event.episode_id] = False
             return None
         if self._fired.get(event.episode_id, False):
             return None
--- a/tests/test_goal_drift.py
+++ b/tests/test_goal_drift.py
@@ -82,3 +82,41 @@ def test_goal_drift_reset_clears_state():
     # After reset, accumulating fresh healthy traffic must not immediately alarm.
     risks = [det.observe(_ev("search", 100.0 + i)) for i in range(4)]
     assert all(r is None for r in risks)
+
+
+def test_goal_drift_rearms_after_recovery():
+    """Issue #247: _fired latched once per episode and never re-armed, so a
+    long-lived episode that recovered from one drift and later hit a second,
+    independent (worse) one stayed silent forever. The latch now clears when
+    the drift score drops back below the threshold, mirroring the re-arm
+    semantics of every other shipped detector."""
+    cfg = Config()
+    cfg.goal_drift_min_samples = 3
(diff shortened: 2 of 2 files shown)


Cyrax321 (approved):
Re-arm fix mirrors the other detectors and the second test guards against alert spam while drift persists. 8/8 pass in test_goal_drift.py, ruff clean. Optional nit (non-blocking): _fired[ep] = False retains an entry per recovered episode in dump_state; pop(ep, None) would carry identical semantics via .get(ep, False) with smaller snapshots.

PR #255, opened 2026-09-07
fix(auto): defer stream emission to exhaustion in openai/anthropic wrappers

### LINKED ISSUE #242: [auto] snagline.auto openai/anthropic wrappers emit at stream-open: streaming calls report ~0ms latency, no tokens, and mid-stream failures as successes
## Summary

The auto-instrumentation wrappers emit their StepEvent in a `finally` block immediately after `create()` returns (`auto/openai.py:50-76`, `auto/anthropic.py:45-71` — identical shape):

```python
def _sync(*args, **kwargs):
    ...
    try:
        result = original(*args, **kwargs)
    except Exception:
        error = True
        raise
    finally:
        _emit(monitor, counter, model, tool_name, sig_text, start, error)
    return result
```

With `stream=True`, `create()` returns almost immediately with a Stream object. The event is therefore emitted at **stream-open**: `latency_ms` ≈ 0.001 ms, `tokens_in/out = None`, `error=False` — and the raw stream is returned unwrapped, so a failure during iteration is never observed (no second event is ever emitted).

The explicit adapters already solved this: `adapters/openai.py:126-209` and `adapters/anthropic.py:104-185` wrap the returned stream in `_SyncStreamWrapper`/`_AsyncStreamWrapper` that defer emission until exhaustion/close/error, with tests in `tests/test_adapter_streaming.py:55-114` asserting "not yet ingested … ingested once after exhaustion" and `error is True` on iteration failure. The auto path was left behind when that fix landed.

Verified empirically on `master` (22faeae) for both modules: a fake `create(stream=True)` returning a stream that sleeps 50 ms/chunk and raises `RuntimeError` mid-flight → one event emitted at stream-open with `latency=0.00095 ms, error=False, tokens=None`; after the iteration raises, no error event exists.

## Why it matters

A deployed app doing mostly streaming calls teaches the CUSUM latency detector a healthy latency of ~0 ms and then flags the *non-streaming* calls (or nothing at all), and real mid-stream API failures — exactly the "LLM 502" class the cascade/latency detectors exist to catch — are recorded as successes. `snagline.auto` is the zero-integration entrypoint, so its data silently skews the baselines.

## Reproduction

```python
import time
from snagline.auto.openai import wrap_client

class FakeStream:
    def __iter__(self):
        time.sleep(0.05); yield {"choices": []}
        raise RuntimeError("mid-stream 502")

class FakeCompletions:
    def create(self, **kw): return FakeStream()

class FakeClient:
    chat = type("C", (), {"completions": FakeCompletions()})()

wrap_client(monitor, FakeClient())
monitor  # a Monitor with a recording sink
result = FakeClient().chat.completions.create(model="gpt", stream=True)
# monitor already holds ONE event: latency≈0.001ms, error=False, tokens=None
list(result)   # raises RuntimeError — no error event is ever emitted
```

Fix shape: port the deferred stream wrappers from the explicit adapters (`_SyncStreamWrapper`/`_AsyncStreamWrapper`) into `auto/openai.py` and `auto/anthropic.py`, emitting only when the stream is exhausted, closed, or fails.

Adhi1-2:
## Summary

The `snagline.auto` wrappers emitted their `StepEvent` in a `finally` block immediately after `create()` returned. With `stream=True`, `create()` returns almost immediately with a Stream object, so the event was emitted at **stream-open**: `latency_ms ≈ 0.001 ms`, `tokens_in/out = None`, `error=False` — and the raw stream was returned unwrapped, so a failure during iteration was never observed.

This ports the deferred stream wrappers (`_SyncStreamWrapper`/`_AsyncStreamWrapper`) from the explicit adapters into `auto/openai.py` and `auto/anthropic.py`: when `create()` succeeds with `stream=True` and a stream-like result, the stream is returned wrapped, and exactly one event is emitted when it is exhausted, closed, or fails mid-flight — now carrying `error_type` and the tokens from the final usage chunk. Non-streaming calls and `create()` failures keep the existing immediate-emission behavior.

## Why it matters

A deployed app doing mostly streaming calls taught the CUSUM latency detector a healthy latency of ~0 ms and recorded real mid-stream API failures — the "LLM 502" class the cascade/latency detectors exist to catch — as successes. `snagline.auto` is the zero-integration entrypoint, so its data silently skewed the baselines.

## Tests

10 new tests across `tests/test_auto_openai.py` and `tests/test_auto_anthropic.py`:

- defers until exhaustion (sync + async, both SDKs) — asserts no event at stream-open, latency reflects streamed duration, tokens extracted from the final usage chunk
- mid-stream iteration failure recorded as `error=True` with `error_type`
- `close()` emits and still closes the underlying stream
- non-streaming calls keep immediate emission

8 of the 10 fail on pre-fix code; the two `still_immediate` tests pin unchanged behavior. Full gates pass locally: `pytest -q` (705 passed, 86.96% coverage), `mypy src tests`, `ruff check`, `ruff format --check`.

Fixes #242

### DIFF
--- a/src/snagline/auto/anthropic.py
+++ b/src/snagline/auto/anthropic.py
@@ -9,6 +9,21 @@ that would record a ~0ms success before the first chunk arrives (issue #242).
 Instead the returned stream is wrapped, and the event is emitted once the
 stream is exhausted, closed, or fails mid-flight, mirroring the explicit
 adapters' ``_SyncStreamWrapper``/``_AsyncStreamWrapper``.
+
+Global mode (issue #270) patches the *resource classes* --
+``anthropic.resources.messages.{Messages, AsyncMessages}`` -- rather than
+walking ``Anthropic.messages``: on modern Anthropic SDKs the client attribute
+is a ``functools.cached_property`` descriptor, so the old class-attribute
+walk resolved a descriptor, not a resource, and silently wrapped nothing.
+
+Latency measurement and event timestamps use :func:`time.perf_counter`, not
+:func:`time.time`, mirroring the explicit adapters (issue #155): the wall
+clock advances in ~15.6 ms ticks on Windows on supported 3.10--3.12
+interpreters, quantizing sub-tick latencies to zero, and is non-monotonic, so
+a clock step mid-call fabricates negative or huge latencies.
+``perf_counter`` has no meaningful epoch, so these timestamps are only
+comparable within one process; detectors consume them solely as in-process
+latency differences.
 """
 
 from __future__ import annotations
@@ -76,15 +91,15 @@ def _emit(
     tokens_out=None,
     metadata=None,
 ) -> None:
-    latency = (time.time() - start) * 1000.0
+    now = time.perf_counter()
     event = StepEvent(
         step_id=str(next(counter)),
         episode_id=EPISODE_ID,
-        timestamp=time.time(),
+        timestamp=now,
         action_type="tool_call",
         action_signature=make_signature("anthropic_call", model, sig_text),
         tool_name=tool_name,
-        latency_ms=latency,
+        latency_ms=(now - start) * 1000.0,
         error=error,
         error_type=error_type,
         tokens_in=tokens_in,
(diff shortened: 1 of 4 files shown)


Cyrax321 (approved):
Verified locally: 20/20 pass across test_auto_openai.py + test_auto_anthropic.py, ruff + mypy clean. Deferring emission to stream exhaustion (mirroring the explicit adapters) fixes the ~0ms false-success events and unwrapped mid-iteration failures, with token extraction at exhaustion. Note: textually conflicts with #271 (4 hunks in the same files) - suggest merging #271 first, then rebasing this and preserving _is_async_call.

Cyrax321:
Status check: still conflicted (auto/openai.py + auto/anthropic.py vs merged #271). Note #281 has since merged into the same files (perf_counter clock) - please rebase and keep the stream-wrapper deferral intact across it. Structural check confirms the wrappers' only sink is Monitor.ingest, so blast radius stays contained. Still approved once rebased.

Adhi1-2:
Rebased onto current upstream/master and refreshed the branch. The approved streaming deferral is preserved, together with the newer resource-class global patching, decorated-async detection, and perf_counter timing. Focused validation: OpenAI streaming tests 9 passed (the no-SDK test was excluded because OpenAI is installed locally), Anthropic tests 10 passed, Ruff and compileall passed.

Cyrax321:
Closing as superseded by merged #288, which re-applies this fix onto current master (see that PR for the review and validation). Thank you for the original fix.

Issue #224, opened 2026-08-30
[core] wall-clock budget emits the 0.7 pre-breach warning one step AFTER the 1.0 breach, inverting severity


Adhi1-2:
`Monitor._advance_clock` emits the wall-clock **breach** (score 1.0, `critical`) and then, on the *next* event, the **pre-breach warning** (score 0.7, `warning`) for the same episode. Severity goes backwards, and the warning's own text is self-contradictory.

### Reproduce

```python
from snagline.config import Config
from snagline.monitor import Monitor
from snagline.events import StepEvent

class Capture:
    def __init__(self): self.seen = []
    def emit(self, risk): self.seen.append(risk)

cap = Capture()
m = Monitor.default(config=Config(max_episode_wall_seconds=100.0, warn_fraction=0.8))
m._sinks = [cap]
for t in (0.0, 200.0, 201.0, 202.0):          # one delta jumps straight past the budget
    m.ingest(StepEvent(f"s{t}", "ep", t, "tool_call", f"sig{t}"))
for r in cap.seen:
    print(r.score, r.severity, r.detail)
```

```
1.0 critical episode exceeded its 100s wall-clock budget (200s observed)
0.7 warning  episode at 201% of its 100s wall-clock budget
```

"at 201% of its budget" is not a warning that the budget is approaching; the episode is already 2x over and has already been reported as `critical`.

### Cause

`src/snagline/monitor.py:424-447`

```python
if not clock.breached and clock.elapsed >= budget:
    clock.breached = True
    out.append(FailureRisk(..., 1.0, "wall_clock_budget", ...))
elif not clock.warned and clock.elapsed >= budget * self._warn_fraction:
    clock.warned = True
    out.append(FailureRisk(..., 0.7, "wall_clock_budget", ...))
```

The `elif` only protects the step that breaches. On the following step `clock.breached` is already `True`, so the first condition is false, the `elif` is evaluated, `clock.warned` is still `False`, and the stale warning fires.

This contradicts the method's own docstring at `src/snagline/monitor.py:379-384`:

> a single delta that jumps straight past the limit fires only the breach (mirrors TokenRunawayDetector's envelope ordering)

It fires only the breach *on that step*, then the warning one step later. A dashboard consuming severity sees `critical` downgraded to `warning` while the overrun keeps growing.

### Suggested fix

[...]

Issue #172, opened 2026-08-26
[core] Persist Monitor time-axis state across restarts: restored episodes lose elapsed budget and re-fire idle/budget risks


Cyrax321:
## Summary

PR #167 (issue #92) added Monitor-level horizon-scale tracking: a per-episode `_EpisodeClock` (`last_ts`, `elapsed`, `idle_fired`, `warned`, `breached`) inside `Monitor._clocks`. Unlike detector state, which survives restarts through the `StatefulDetector` snapshot machinery (issue #91), this clock lives on the Monitor itself and is NOT captured by `Monitor.snapshot()` / `restore()`.

On a process restart mid-episode (deploy, crash, OOM kill: exactly the multi-day scenario the horizon knobs target) every restored episode:

1. **Resets elapsed budget to zero.** A wall-clock budget half-consumed before the restart silently starts over; the breach fires late or never.
2. **Clears all three latches.** `idle_gap` / `wall_clock_budget` risks that already fired can fire again after the restart: duplicate alerts on every deploy.
3. **Loses `last_ts`.** The first post-restart ingest is treated as an episode start, so a real silence gap that straddles the restart goes undetected.

This does not affect default configurations (all horizon knobs are opt-in), only deployments running with `idle_warn_seconds` and/or `max_episode_wall_seconds` set across restarts.

## Why this was not done in #167

Scope discipline plus ownership boundaries. The snapshot payload contract (`format_version`, detector-keyed layout, strict-vs-tolerant load semantics) belongs to the #91 machinery, and parallel agents were actively extending it while #92 was in flight (#149 via PR #168 merged mid-rebase). Growing the payload shape from the #92 branch risked colliding with that work for zero functional loss at merge time, so the gap was documented instead of smuggled in.

## Proposed design

- Additive top-level `"time_axis"` key in `Monitor.snapshot_dict()`: `{episode_id: {"last_ts": ..., "elapsed": ..., "idle_fired": ..., "warned": ..., "breached": ...}}`.
- Tolerant in both directions, same precedent as #149's payload extension: old snapshots without the key restore into fresh clocks (`dict.get` default), and `restore_dict` already ignores keys it does not know, so newer snapshots stay loadable by older code without a `format_version` bump.
- Restoring `last_ts` keeps the region contract from #92 intact: every persisted value derives from StepEvent timestamps, never the wall clock, so replay stays deterministic.
[...]

PR #256, opened 2026-09-07
fix(monitor): only move the episode clock's last_ts forward

### LINKED ISSUE #249: [core] _advance_clock rewinds last_ts on out-of-order timestamps, so the same wall-clock span is counted into elapsed twice and wall_clock_budget breaches early
## Summary

The time-axis advance computes `delta = event.timestamp - clock.last_ts` and **always** rewinds `last_ts` to the event timestamp — including when the delta is negative (`monitor.py:416-417`):

```python
delta = event.timestamp - clock.last_ts
clock.last_ts = event.timestamp
if delta > 0:
    clock.elapsed += delta   # "must not reduce consumed budget"
```

A negative delta is correctly excluded from `elapsed`, but `last_ts` is still moved backwards. The *next* event then measures its delta from the rewound `last_ts`, accumulating a positive span that was **already counted once**. Any source with out-of-order or backwards-skewed timestamps (merged adapter streams, clock skew between hook processes — exactly the case the comment says the code tolerates) double-counts wall-clock time into `elapsed`.

Verified on `master` (22faeae) with `max_episode_wall_seconds=100`: events at ts `0, 60, 0, 60` (a 60-second true span, the second half a skewed replay of the same range) yield `clock.elapsed == 120.0`, and both the 0.7 warning and the 1.0 breach fire — a 100 s budget declared exceeded on 60 s of real time, with the detail reporting "({elapsed:.0f}s observed)".

The `_EpisodeClock` docstring defines elapsed as "sum of positive inter-event deltas", which the code matches literally — but the derived `wall_clock_budget` risk is then simply wrong whenever timestamps are non-monotonic, which is why I'm filing this as a defect rather than a definition.

## Why it matters

`wall_clock_budget` is a hard enforcement trigger (score 1.0, halt-policy routing by trigger name). A premature breach fires the strongest severity SNAGLINE has on an episode that has consumed less budget than reported. Not covered by #172 (persistence across restarts) or #224 (warning/breach ordering).

## Reproduction

```python

m = Monitor.default(config=Config(max_episode_wall_seconds=100), sinks=[sink])
for ts in (0.0, 60.0, 0.0, 60.0):
    m.ingest(StepEvent(step_id=str(ts), episode_id="ep", timestamp=ts,
                       action_type="tool_call", action_signature="s"))
# sink holds wall_clock_budget breach; elapsed == 120.0 on a 60s true span
```

Fix shape: only move `last_ts` forward (`if delta > 0: clock.elapsed += delta; clock.last_ts = event.timestamp`), so an out-of-order event neither spends nor un-spends budget and never re-opens an already-counted span.

Adhi1-2:
## Summary

`_advance_clock` computed `delta = event.timestamp - clock.last_ts` and **always** rewound `last_ts` to the event timestamp — including when the delta was negative. The negative delta was correctly excluded from `elapsed`, but `last_ts` was still moved backwards, so the *next* event measured its delta from the rewound reference and accumulated a positive span that was **already counted once**.

`last_ts` now only moves forward: a late event neither spends nor un-spends budget, and never re-opens an already-counted span. Genuinely new time past the previous high-water mark still spends budget as before.

## Why it matters

`wall_clock_budget` is a hard enforcement trigger (score 1.0, halt-policy routing by trigger name). A premature breach fires the strongest severity SNAGLINE has on an episode that has consumed less budget than reported. Reproduced on `master` with `max_episode_wall_seconds=100`: events at ts `0, 60, 0, 60` (a 60-second true span) yielded `elapsed == 120.0` and both the 0.7 warning and the 1.0 breach. Not covered by #172 (persistence across restarts) or #224 (warning/breach ordering).

## Tests

- `test_out_of_order_events_do_not_double_count_the_same_span` — ts `0, 60, 0, 60` yields `elapsed == 60`, `last_ts == 60`, no risks on a 100 s budget (fails pre-fix).
- `test_out_of_order_replay_still_counts_genuinely_new_time` — after a backwards event, a delta past the high-water mark still breaches (fails pre-fix because of the double-count distortion).

Full gates pass locally: `pytest -q` (697 passed, 87.48% coverage), `mypy src tests`, `ruff check`, `ruff format --check`.

Fixes #249

### DIFF
--- a/src/snagline/monitor.py
+++ b/src/snagline/monitor.py
@@ -96,7 +96,7 @@ class _EpisodeClock:
 
     def __init__(self, first_ts: float) -> None:
         self.last_ts = first_ts
-        self.elapsed = 0.0  # sum of positive inter-event deltas
+        self.elapsed = 0.0  # wall-clock span covered; last_ts only moves fwd
         self.idle_fired = False  # "idle_gap" fires once per episode
         self.warned = False  # budget warning fires once per episode
         self.breached = False  # budget breach fires once per episode
@@ -397,7 +397,16 @@ class Monitor:
                 self._clocks[event.episode_id] = _EpisodeClock(event.timestamp)
                 return out
             delta = event.timestamp - clock.last_ts
-            clock.last_ts = event.timestamp
+            if delta > 0.0:
+                # Out-of-order or skewed sources (merged adapter streams, clock
+                # skew between hook processes) must not reduce consumed budget
+                # -- and must not rewind ``last_ts`` either: rewinding makes
+                # the *next* event measure its delta from the moved-back
+                # reference, counting an already-counted span a second time
+                # (issue #249). ``last_ts`` only moves forward, so a late
+                # event neither spends nor un-spends budget.
+                clock.elapsed += delta
+                clock.last_ts = event.timestamp
             if (
                 self._idle_warn_seconds is not None
                 and not clock.idle_fired
@@ -415,10 +424,6 @@ class Monitor:
                         event.timestamp,
                     )
                 )
-            if delta > 0.0:
-                # Negative deltas (out-of-order or skewed sources) must not
-                # reduce consumed budget; clamp them out of the accumulation.
-                clock.elapsed += delta
             budget = self._max_wall_seconds
             if budget is not None:
                 if not clock.breached and clock.elapsed >= budget:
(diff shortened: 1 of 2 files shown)


Cyrax321 (approved):
Verified locally: 15/15 pass in test_horizon_time_axis.py, ruff + mypy clean. last_ts moving forward-only fixes the double-counted span (0,60,0,60 no longer breaches a 100s budget) while genuinely new time past the high-water mark still spends. Noted #261 is stacked on this branch - merge this first, #261 follows seamlessly. Safe to merge.

Issue #226, opened 2026-08-30
[cli] hook bridge drops side_effect and metadata, making side_effect_guard and compaction_tripwire unreachable over the bridge


Adhi1-2:
`snagline hook` is the documented bridge from a host (Claude Code hooks, or anything emitting the same JSON) into either a sidecar (`--url`) or `snagline watch` (`--out`). The bridge cannot carry `StepEvent.side_effect` or `StepEvent.metadata`, so the two detectors that depend on those fields can never fire over it.

### Reproduce

```bash
echo '{"session_id":"s1","tool_name":"charge_card","tool_input":{"amount":42},"hook_event_name":"PostToolUse"}' \
  | snagline hook --out /tmp/bridged.jsonl
cat /tmp/bridged.jsonl
```

```json
{"step_id": "b7551f129476", "episode_id": "s1", "timestamp": 1788113286.592117,
 "action_type": "tool_call", "action_signature": "72e7d655...", "tool_name": "charge_card",
 "latency_ms": null, "error": false, "error_type": null, "tokens_in": null, "tokens_out": null}
```

No `side_effect`, no `metadata`.

### Cause

Two gaps, both in `src/snagline/cli.py`:

1. `_event_to_json` (L643-L656) enumerates 11 of `StepEvent`'s 13 fields and omits `side_effect` and `metadata`. It is the serializer for both `--out` (L614) and `--url` (L624), so the drop applies to both.
2. There is no way to *set* `side_effect` on the CLI path in the first place. `snagline hook` maps via `snagline.adapters.claude_code.payload_to_event`, which hardcodes `metadata={}` and never passes `side_effect` (`src/snagline/adapters/claude_code.py:186`). `cli.py` contains no occurrence of either field.

### Why it matters

- `SideEffectGuardDetector` reads `event.side_effect` and nothing else; its own docstring calls duplicate sends/payments/deploys "the single most expensive failure class in production agents". Over the hook bridge the flag is always `False`, so the guard is structurally unreachable.
- `CompactionTripwireDetector` reads exactly two `metadata` keys (`pinned`, `pin`) — see `src/snagline/detectors/compaction_tripwire.py:130-138` — and gets `{}` every time.

The receiving side is already willing: `src/snagline/server/http_server.py` constructs events with `StepEvent(**item)` / `StepEvent(**obj)`, and so does `_cmd_watch`. Both accept the two keys if they are present. The loss is purely in the CLI serializer.

### Suggested fix

[...]

Issue #174, opened 2026-08-26
[tests] hook-bridge graceful-stop escalates to SIGTERM under heavy CPU load: exit-code assertions intermittently fail


Cyrax321:
## Summary

Running the test suite in parallel with itself (two concurrent full-suite pytest processes on one machine) makes exactly three `tests/test_hook_bridge.py` tests fail intermittently with the same mechanism: the spawned `snagline watch` child misses the 10-second graceful window after SIGINT, `_graceful_stop` escalates through its terminate() rung, and the child dies with `returncode == -15` (SIGTERM) instead of the asserted `0`. This was discovered and root-caused during #156 verification (PR #170) and is deliberately NOT fixed there: #156 scoped the delivery-vs-teardown race only, and this is a distinct, pre-existing failure mode.

## Reproduction

```bash
python -m venv .venv && .venv/bin/pip install -e ".[langchain]" pytest
( .venv/bin/python -m pytest tests/ -q -o addopts="" > /tmp/a.log 2>&1 & \
  .venv/bin/python -m pytest tests/ -q -o addopts="" > /tmp/b.log 2>&1 & wait )
grep "^FAILED" /tmp/a.log
```

Observed on macOS (CPython 3.14, 2026-08-26), reproducible on demand:

- current master `b887648`: both suites fail the same three tests, everything else green
- pristine base `e7cd8f3` (BEFORE PRs #161/#163/#168/#170 merged): identical three failures

## Evidence

Both parallel logs, both commits:

```
FAILED tests/test_hook_bridge.py::test_watch_file_follow_sees_appended_lines
FAILED tests/test_hook_bridge.py::test_watch_follow_cleanup_reaps_the_child_when_the_sink_endpoint_is_dead
FAILED tests/test_hook_bridge.py::test_watch_follow_teardown_reaps_the_child_on_a_healthy_run
...
>           assert proc.returncode == 0
E           AssertionError: assert -15 == 0
```

What rules out other causes:

- The delivery contract added by #156 held every time: zero "never delivered" failures across four parallel suite executions; the loop payload always landed before teardown.
- The sentinel contract held every time: `marker_seen.wait(30)` succeeded, so the children had finished ingesting before being asked to stop.
- All failing children show `returncode == -15`, i.e. the SIGTERM escalation rung of `_graceful_stop`, never a crash, hang, or orphan.
- Sequential runs of the same suite: green (30/30 consecutive runs recorded in PR #170).

## Root cause

[...]

PR #258, opened 2026-09-07
fix(adapter): serialize HookTracker state shared by the sidecar's threads

### LINKED ISSUE #245: [adapter] The sidecar's shared HookTracker mutates its start table from every handler thread with no lock: eviction races raise RuntimeError (silently swallowed) and drop latency samples
## Summary

`make_handler` creates **one** `HookTracker` shared as a class attribute by every worker of the threaded sidecar (`server/http_server.py:522, 893`). `HookTracker.note()` mutates `self._starts` with no lock (`adapters/claude_code.py:75-91`), and once pending `PreToolUse` starts exceed `max_pending` (1024 — plausible for a busy multi-session sidecar: long-running tools, dropped Post hooks, 600 s TTL), `note()` calls `_evict()`, which iterates the dict in a comprehension (`claude_code.py:102`) and again via `sorted()` (`:105`) while other threads insert — raising `RuntimeError: dictionary changed size during iteration` — and then **rebinds** `self._starts = dict(by_age[...])` (`:106`), silently discarding inserts made concurrently into the old dict.

In the sidecar path the failure is invisible: `ingest_payload`'s `finally: with contextlib.suppress(Exception): tracker.note(payload)` (`claude_code.py:211-214`) swallows the RuntimeError. Consequences:

1. Eviction aborts, the table stays over cap, so every subsequent `note()` retries eviction and fails again.
2. A lost `PreToolUse` start means `latency_ms=None` on its matching `PostToolUse` — the CUSUM latency detector silently never sees that sample.

Reproduced standalone on `master` (22faeae): 7 of 8 concurrent `note()` threads hit the RuntimeError under load. The class docstring claims "malformed payloads never raise out of note/latency" (`claude_code.py:60-62`) — true per payload, false under the sidecar's concurrency.

Distinct from #174/#156/#215 (hook-bridge **test** flakiness) and #226 (the bridge dropping fields): this is a production thread-safety defect in state shared by the sidecar's handler threads.

## Why it matters

The Claude Code sidecar is the long-lived deployment shape; a busy multi-session host (exactly the target for `--max-pending`-scale load) silently degrades its latency telemetry with no log line anywhere.

## Reproduction sketch

```python
import threading
from snagline.adapters.claude_code import HookTracker

tracker = HookTracker(max_pending=8, ttl_seconds=1.0)
errors = []
def worker(tag):
    try:
        for i in range(2000):
            tracker.note({"hook_event_name": "PreToolUse",
                          "tool_use_id": f"{tag}-{i}"})
    except Exception as e:
        errors.append(e)
threads = [threading.Thread(target=worker, args=(f"w{i}",)) for i in range(8)]
[t.start() for t in threads]; [t.join() for t in threads]
print(errors[:1])   # RuntimeError: dictionary changed size during iteration
```

Fix shape: a `threading.Lock` guarding `_starts` in `note`/`latency_ms`/`_evict` (eviction rebuilds under the lock instead of rebinding around it).

Adhi1-2:
## Summary

`make_handler` creates **one** `HookTracker` shared as a class attribute by every worker of the threaded sidecar, but `note()` mutated `self._starts` with no lock. Once pending `PreToolUse` starts exceeded `max_pending`, `_evict()` iterated the dict while other threads inserted — raising `RuntimeError: dictionary changed size during iteration` — and **rebound** `self._starts`, silently discarding inserts made concurrently into the old dict.

In the sidecar path the failure was invisible: `ingest_payload`'s `contextlib.suppress(Exception)` swallowed the RuntimeError. Consequences:

1. Eviction aborted, the table stayed over cap, so every subsequent `note()` retried eviction and failed again.
2. A lost `PreToolUse` start meant `latency_ms=None` on its matching `PostToolUse` — the CUSUM latency detector silently never saw that sample.

Every access to `_starts` is now serialized by a `threading.Lock` (`note`/`latency_ms`, and `_evict` called under it). Eviction drains the table **in place** instead of rebinding it, so concurrent inserts land in the same dict. The class docstring's "malformed payloads never raise out of note/latency" contract now also holds under the sidecar's concurrency.

## Why it matters

The Claude Code sidecar is the long-lived deployment shape; a busy multi-session host (exactly the target for `--max-pending`-scale load) silently degrades its latency telemetry with no log line anywhere. Distinct from #174/#156/#215 (hook-bridge test flakiness) and #226 (the bridge dropping fields): this is a production thread-safety defect in state shared by the sidecar's handler threads.

## Tests

- `test_tracker_concurrent_note_never_raises_and_keeps_every_sample` — 8 workers × 2000 notes against a tiny-cap tracker; no exceptions, no deadlock, and pairing still works afterwards (fails pre-fix when the race manifests).
- `test_tracker_eviction_under_concurrency_keeps_table_under_cap` — eviction drains under load rather than leaving the table permanently over cap.
- `test_tracker_eviction_preserves_table_identity` — deterministic pin of the rebind mechanism: the dict's `id` never changes across cap-shed or TTL sweep (fails pre-fix).
- `test_tracker_lock_serializes_latency_reads_against_eviction` — `latency_ms` reads under the same lock (fails pre-fix: no lock attribute exists).

Full gates pass locally: `pytest -q` (699 passed, 87.50% coverage), `mypy src tests`, `ruff check`, `ruff format --check`.

Fixes #245

### DIFF
--- a/src/snagline/adapters/claude_code.py
+++ b/src/snagline/adapters/claude_code.py
@@ -29,6 +29,7 @@ from __future__ import annotations
 
 import contextlib
 import json
+import threading
 import time
 import uuid
 from typing import Any
@@ -59,6 +60,11 @@ class HookTracker:
 
     Stateful companion for a long-lived bridge (the sidecar server). Kept tiny
     and defensive: malformed payloads never raise out of ``note``/``latency``.
+    The sidecar's threaded handler workers share one tracker, so every access
+    to ``_starts`` is serialized by a lock (issue #245): without it, eviction
+    iterated/rebound the dict while other threads inserted -- raising
+    ``RuntimeError: dictionary changed size during iteration`` and silently
+    dropping latency samples.
     """
 
     def __init__(
@@ -69,6 +75,7 @@ class HookTracker:
     ) -> None:
         self._clock = clock or time.monotonic
         self._starts: dict[str, float] = {}
+        self._lock = threading.Lock()
         self._ttl = ttl_seconds
         self._max_pending = max_pending
 
@@ -83,12 +90,13 @@ class HookTracker:
         if not isinstance(tool_use_id, str):
             return
         now = self._clock()
-        if event == "PreToolUse":
-            self._starts[tool_use_id] = now
-        else:
-            self._starts.pop(tool_use_id, None)
-        if len(self._starts) > self._max_pending:
-            self._evict(now)
+        with self._lock:
+            if event == "PreToolUse":
+                self._starts[tool_use_id] = now
+            else:
+                self._starts.pop(tool_use_id, None)
+            if len(self._starts) > self._max_pending:
+                self._evict(now)
 
     def _evict(self, now: float) -> None:
         """Bound memory by age, so a tool still running keeps its start.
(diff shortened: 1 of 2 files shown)


Cyrax321 (approved):
Verified locally: 18/18 pass in test_claude_code_adapter.py, ruff + mypy clean. Lock serializes the shared table and the in-place drain (clear+update instead of rebind) fixes the silent insert loss. Note: textually conflicts with #250 in claude_code.py - please rebase after #250 merges.

Cyrax321:
Status check: still conflicted (claude_code.py vs merged #250). A straight rebase should resolve it - your hunks (HookTracker lock + in-place drain) don't overlap #250's signature hunks semantically. Structural check confirms HookTracker is only shared on the sidecar path. Still approved once rebased.

Adhi1-2:
Rebased onto current upstream/master and preserved the approved HookTracker locking/in-place eviction fix. The textual test conflict was resolved by retaining both the existing issue #237 signature regressions and the issue #245 concurrency regressions. Validation: 21 focused tests passed, Ruff check/format passed, compileall passed.

Cyrax321:
Closing as superseded by merged #287, which re-applies this fix onto current master (see that PR for the review and validation). Thank you for the original fix.

PR #259, opened 2026-09-07
fix(detector): score a CUSUM alarm coinciding with baseline adoption from the alarm-time snapshot

### LINKED ISSUE #244: [detector] Latency CUSUM alarm coinciding with a baseline re-fit adoption scores flat 0.6 and cites the new baseline as the one deviated from
## Summary

In `LatencyAnomalyDetector.observe()`, the alarm is computed **before** the periodic re-fit advances (`detectors/latency_anomaly.py:219-253`):

```python
alarm = state.update(event.latency_ms)          # 219
if self.refit_every > 0:
    shifted, old_mu, shift = self._advance_refit(state, event.latency_ms)  # 222
    ...
if alarm:                                        # 243
    score = min(1.0, 0.6 + 0.1 * max(0.0, state.cusum / self.h - 1.0))   # 244
    return FailureRisk(...,
        f"latency {event.latency_ms:.0f}ms deviates from baseline "
        f"(mean {state.mu0:.0f}ms)", ...)                               # 250-251
```

When the parallel learner completes on the same step, `_advance_refit` calls `state.adopt_candidate()` (`latency_anomaly.py:285`), which sets `self.cusum = 0.0` and replaces `mu0` (`latency_anomaly.py:138, 140`). The alarm is then scored with the **already-reset** accumulator and reported against the **already-adopted** baseline.

Reproduced deterministically on `master` (22faeae) with `cusum_min_samples=5, cusum_refit_every=3` (warm-up at 100ms, then sustained 300ms):

```
step 5: cusum_after=237.000 mu0=100.0 risk=(1.0, 'latency 300ms deviates from baseline (mean 100ms)')
step 6: cusum_after=0.000   mu0=300.0 risk=(0.6,  'latency 300ms deviates from baseline (mean 300ms)')
```

The step-6 alarm should have scored 1.0 (cusum was ~276 pre-adoption) but emits a flat 0.6, and the detail reads "300ms deviates from baseline (mean 300ms)" — self-contradictory.

This co-occurrence is not rare: a sustained regression both keeps the CUSUM alarming and is exactly what the learner samples, so adoption routinely lands mid-alarm. The deferred drift risk itself is not lost (it surfaces next step — `pending_drift`), but the coincident alarm's severity and detail are wrong.

Not covered by #229 (`reset()` iterating the states dict) or #101 (calibrated start) — this is the periodic re-fit path from #92. `test_cusum_refit_surfaces_baseline_drift` only asserts both risks eventually appear, not what the coincident step emits.

## Why it matters

Severity is load-bearing: `SlackSink`/`PagerDutySink` route on `min_severity` and score feeds the risk score ceiling. An operator paged with "deviates from baseline (mean 300ms)" for a 300ms latency has no actionable message, and the lost severity bonus downgrades a genuine sustained regression to the floor score on exactly the steps where it peaks.

## Reproduction

Config with `cusum_min_samples=5, cusum_refit_every=3`; feed 5 events at 100ms (warm-up), then 300ms events; inspect the risk emitted on the step where adoption lands (step 6 above): score is 0.6 and the detail cites the adopted mean, not the baseline the alarm was measured against.

Fix shape: snapshot `cusum`/`mu0` after `state.update()` but before `_advance_refit()`, and use the snapshot for the alarm's score and detail.

Adhi1-2:
## Summary

In `LatencyAnomalyDetector.observe()`, the alarm was computed **before** the periodic re-fit advanced, but **scored after it**. When adoption landed on the same step as an alarm, `adopt_candidate()` had already reset `cusum` to 0 and replaced `mu0`, so the alarm emitted:

- a flat **0.6** score — the severity bonus `0.6 + 0.1 * max(0, cusum/h - 1)` computed from the already-reset accumulator, and
- the self-contradictory detail **"latency 300ms deviates from baseline (mean 300ms)"** — citing the just-adopted baseline as the one deviated from.

Reproduced deterministically with `cusum_min_samples=5, cusum_refit_every=3` (warm-up 100 ms, then sustained 300 ms): the coincident step's alarm scored 0.6 instead of 1.0 (cusum was ~276 pre-adoption).

This co-occurrence is not rare: a sustained regression both keeps the CUSUM alarming and is exactly what the learner samples, so adoption routinely lands mid-alarm.

`cusum` and `mu0` are now snapshotted right after `state.update()`, before `_advance_refit()`, and the alarm's score and detail use that snapshot. The deferred drift risk and post-adoption quieting are unchanged.

## Why it matters

Severity is load-bearing: `SlackSink`/`PagerDutySink` route on `min_severity` and score feeds the risk score ceiling. An operator paged with "deviates from baseline (mean 300ms)" for a 300 ms latency has no actionable message, and the lost severity bonus downgrades a genuine sustained regression to the floor score on exactly the steps where it peaks.

## Tests

`test_cusum_alarm_coincident_with_adoption_keeps_severity_and_detail` — reproduces the issue's deterministic setup; asserts every emitted CUSUM alarm carries `score > 0.6` and cites `mean 100ms`, never the just-adopted `mean 300ms` (fails pre-fix on both assertions). Full gates pass locally: `pytest -q` (696 passed, 87.49% coverage), `mypy src tests`, `ruff check`, `ruff format --check`.

Fixes #244

### DIFF
--- a/src/snagline/detectors/latency_anomaly.py
+++ b/src/snagline/detectors/latency_anomaly.py
@@ -217,6 +217,13 @@ class LatencyAnomalyDetector:
             return None
 
         alarm = state.update(event.latency_ms)
+        # Snapshot the alarm-time accumulator and baseline *before* the
+        # re-fit advances: a coincident adoption resets cusum to zero and
+        # replaces mu0, so scoring the alarm after _advance_refit reads the
+        # already-reset state -- a flat 0.6 score and a self-contradictory
+        # "300ms deviates from baseline (mean 300ms)" detail (issue #244).
+        alarm_cusum = state.cusum
+        alarm_mu0 = state.mu0
         drift_risk: FailureRisk | None = None
         if self.refit_every > 0:
             shifted, old_mu, shift = self._advance_refit(state, event.latency_ms)
@@ -241,14 +248,15 @@ class LatencyAnomalyDetector:
                         event.timestamp,
                     )
         if alarm:
-            score = min(1.0, 0.6 + 0.1 * max(0.0, state.cusum / self.h - 1.0))
+            assert alarm_mu0 is not None
+            score = min(1.0, 0.6 + 0.1 * max(0.0, alarm_cusum / self.h - 1.0))
             return FailureRisk(
                 event.episode_id,
                 event.step_id,
                 score,
                 "latency_anomaly",
                 f"latency {event.latency_ms:.0f}ms deviates from baseline "
-                f"(mean {state.mu0:.0f}ms)",
+                f"(mean {alarm_mu0:.0f}ms)",
                 event.timestamp,
             )
         return drift_risk
--- a/tests/test_window_scaling.py
+++ b/tests/test_window_scaling.py
@@ -178,6 +178,38 @@ def test_cusum_refit_surfaces_baseline_drift() -> None:
     assert all(r is None for r in risks[-5:])
 
 
+def test_cusum_alarm_coincident_with_adoption_keeps_severity_and_detail() -> None:
+    """Issue #244: the alarm was scored *after* the periodic re-fit advanced.
+    When adoption landed on the same step as an alarm, adopt_candidate() had
+    already reset cusum to 0 and replaced mu0, so the risk came out with a
+    flat 0.6 score and the self-contradictory detail "300ms deviates from
+    baseline (mean 300ms)". The alarm must be scored against the snapshot
+    taken at alarm time, before the re-fit touches the state.
+    """
+    cfg = Config(cusum_min_samples=5, cusum_refit_every=3)
+    det = LatencyAnomalyDetector(config=cfg)
+    ts = 0.0
+    for i in range(5):  # healthy warm-up at 100ms, then frozen
+        det.observe(_event(f"w{i}", ts, "s", latency_ms=100.0))
+        ts += 1.0
+    risks = []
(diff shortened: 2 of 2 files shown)


Adhi1-2:
The failed `windows-latest, py3.12` leg is the known test-harness flake from #215, not this change:

```
FAILED tests/test_hook_bridge.py::test_watch_follow_teardown_reaps_the_child_on_a_healthy_run
Failed: watcher never delivered the loop risk to the sink endpoint within 30s;
bodies=[b'POST /sink HTTP/1.1\r\n...Content-Length: 151\r\n...\r\n\r\n']
```

`bodies` records `Content-Length: 151` and then stops at the blank line — the stub sink's single `recv()` consumed the header segment and discarded the body segment, exactly as #215 documents. The failure is in `tests/test_hook_bridge.py`'s `_HookTargetServer._serve`; this PR touches `src/snagline/detectors/latency_anomaly.py` + `tests/test_window_scaling.py` only. All other legs pass, including windows-latest py3.13.

PR #216 (by abyyxhek) fixes the stub's truncated read; once it lands this flake should disappear.

Cyrax321 (approved):
Verified locally: 14/14 pass in test_window_scaling.py, ruff + mypy clean. Snapshotting cusum/mu0 at alarm time before the re-fit advances fixes both the flat-0.6 score and the self-contradictory detail. Minimal, correct. Safe to merge.

PR #260, opened 2026-09-07
fix(ml): score the ESN live path with the same pairing fit() solved for

### LINKED ISSUE #243: [ml] ESN anomaly scores each step against a reservoir state that already contains that step: live path contradicts both training paths
## Summary

The ESN module promises next-step prediction: *"the reservoir predicts the next feature vector from the previous reservoir state"* (`ml/esn_ensemble.py:19-20`). Both training paths honor that pairing — `fit()` builds `(state after x_{t-1}) → x_t` pairs (`esn_ensemble.py:185-191`):

```python
for event in events:
    x = self._features(event)
    if prev_context is not None:
        contexts.append(prev_context)      # state after x_{t-1}
        targets.append(x)                  # x_t
    state = self._advance(state, x)
    prev_context = np.concatenate((np.ones(1), state))
```

…as does the warm-up learner (`esn_ensemble.py:271-280`). But the live scoring path advances the reservoir with the current step **first**, then asks the readout to predict that same step (`esn_ensemble.py:264-283`):

```python
def _esn_anomaly(self, st, x):
    st.state = self._advance(st.state, x)          # state now includes x_t
    context = np.concatenate((np.ones(1), st.state))
    ...
    predicted = context @ st.beta                  # predicts x_t from post-x_t state
    residual = float(np.linalg.norm(predicted - x))
```

The step being judged leaks into the context judging it, so the prediction is partially self-fulfilling and residuals shrink for exactly the anomalies the detector should amplify. `beta` — fitted or warm-up-solved — was solved for the mapping *(state after x_{t-1}) → x_t*, so it is applied to a systematically different input distribution. Note `st.context_prev` (the correct, pre-advance context) is only maintained inside the warm-up branch and never updated on the fitted path.

Verified numerically (numpy 2.5.3, seeded constructor) on `master` (22faeae): fitted on a period-2 latency stream (fitted res_mu 0.0042, sigma 0.0007), a perfectly **healthy** continuation produces mean residual 0.0137 under the shipped pairing vs 0.0054 under the fit-consistent (pre-x) pairing — ~3x the trained mean (~13 sigma) on clean data, purely from the pairing mismatch. The running residual-stat update then absorbs this bias, degrading signal/noise of the anomaly channel in both directions.

## Why it matters

The ML ensemble is one of the shipped horizon detectors; a systematic 3x residual bias on healthy traffic means the CUSUM threshold over `ml_ensemble` is effectively tuned against noise, masking real anomalies and inflating the warm-up statistics the next episode inherits.

## Reproduction sketch

```python
esn = ESNEnsemble(...)  # seeded
healthy = [event(latency=100.0 if i % 2 else 200.0) for i in range(40)]
esn.fit(healthy)
continuation = [event(latency=100.0 if i % 2 else 200.0) for i in range(40, 80)]
# mean live residual ≈ 0.0137 (shipped) vs ≈ 0.0054 (fit-consistent pairing)
```

Fix shape: score from `st.context_prev` (the pre-advance context), and keep `st.context_prev = context` updated on the fitted path as well.

Adhi1-2:
## Summary

The ESN module promises next-step prediction: *"the reservoir predicts the next feature vector from the previous reservoir state."* Both training paths honor that pairing — `fit()` and the warm-up learner solve beta for `(state after x_{t-1}) → x_t`. But the live scoring path advanced the reservoir with the current step **first**, then asked the readout to predict that same step:

```python
st.state = self._advance(st.state, x)          # state now includes x_t
context = np.concatenate((np.ones(1), st.state))
predicted = context @ st.beta                  # predicts x_t from post-x_t state
```

The step being judged leaked into the context judging it, so beta — solved for a systematically different input distribution — produced biased residuals.

Live scoring now predicts `x_t` from `st.context_prev` (the pre-advance context), and `st.context_prev` is kept updated on the fitted path as well (previously it was only maintained inside the warm-up branch). The first scored step of an episode has no pre-advance context — `fit()` skips the same step — and stays silent once, matching the training contract.

## Verification

Measured on `master` with a seeded constructor, fitted on a period-2 latency stream and fed a perfectly **healthy** continuation through the real `observe()` path for 200 steps:

- **master**: running `res_mu` drifts **+1.98 σ** above the fitted mean — purely from the pairing mismatch; the residual-stat update absorbs the bias
- **this PR**: drift is **−0.03 σ**

The note in the issue about a ~3× residual amplification didn't reproduce on my machine (I measured ~1.10× on healthy continuation); the systematic bias itself is real and deterministic, so the tests pin the σ-scaled drift instead of the ratio.

## Tests

- `test_fitted_live_residuals_match_fit_pairing_on_healthy_traffic` — 200 in-pattern `observe()` steps; `res_mu` must stay within 0.5 σ of the fitted mean (fails pre-fix at ~+2 σ). Drives the real code path, not a hand-replicated one.
- `test_fitted_first_live_step_is_silent_then_scores` — the first scored step has no pre-advance context and is silent; the second scores normally (fails pre-fix: master scores the first step from a context containing it).

Full local gates: `pytest -q` (717 passed with the ml extra, 87.50% coverage), `benchmarks/detection_accuracy.py` exits 0 (no healthy-control false positives), `mypy src tests` (clean at the CI matrix's Python versions; the local numpy-stub error is an env artifact — CI's mypy leg runs without numpy), `ruff check`, `ruff format --check`.

Fixes #243

### DIFF
--- a/src/snagline/ml/esn_ensemble.py
+++ b/src/snagline/ml/esn_ensemble.py
@@ -263,15 +263,22 @@ class EsnCusumDetector:
 
     def _esn_anomaly(self, st: _EpisodeState, x: np.ndarray) -> float:
         """Next-step prediction anomaly of the ESN, in [0, 1]."""
+        # Score x from the *pre-advance* context (state after x_{t-1}) -- the
+        # exact pairing both fit() and the warm-up learner solved beta for.
+        # The step being judged must not leak into the context judging it:
+        # scoring from the post-advance state made the prediction partially
+        # self-fulfilling, shrinking residuals for exactly the anomalies the
+        # detector should amplify (issue #243).
+        context_prev = st.context_prev
         st.state = self._advance(st.state, x)
         context = np.concatenate((np.ones(1), st.state))
         if st.beta is None:
             # Warm-up: learn this episode's own (assumed healthy) dynamics
             # from previous-step context to current features. Silent phase.
-            if st.context_prev is not None:
-                st.gram += np.outer(st.context_prev, st.context_prev)
-                st.rhs += np.outer(st.context_prev, x)
-                st.warm_ctx.append(st.context_prev)
+            if context_prev is not None:
+                st.gram += np.outer(context_prev, context_prev)
+                st.rhs += np.outer(context_prev, x)
+                st.warm_ctx.append(context_prev)
                 st.warm_tgt.append(x)
                 st.warm_n += 1
                 if st.warm_n >= self._warmup_steps:
@@ -279,7 +286,14 @@ class EsnCusumDetector:
                     self._seed_residual_stats(st)
             st.context_prev = context
             return 0.0
-        predicted = context @ st.beta
+        # Fitted path: remember the pre-advance context for the *next* step's
+        # score, so live scoring keeps the fit-consistent pairing.
+        st.context_prev = context
+        if context_prev is None:
+            # First scored step of the episode has no pre-advance context to
+            # predict from (fit() skips the same step); stay silent once.
+            return 0.0
+        predicted = context_prev @ st.beta
         residual = float(np.linalg.norm(predicted - x)) / math.sqrt(_FEATURE_DIM)
         excess = max(0.0, residual - st.res_mu) / (
             _RESIDUAL_SIGMA_INFLATION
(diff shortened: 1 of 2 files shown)


Cyrax321 (approved):
Verified locally with numpy installed: 22/22 pass in test_ml_esn_ensemble.py, ruff + mypy clean. Live path now predicts x_t from the pre-advance context, matching the pairing fit() solved for - the healthy-traffic res_mu drift check proves the bias is gone. Safe to merge.

PR #261, opened 2026-09-07
fix(monitor): suppress the wall-clock pre-breach warning after a breach

### LINKED ISSUE #224: [core] wall-clock budget emits the 0.7 pre-breach warning one step AFTER the 1.0 breach, inverting severity

Adhi1-2:
## Summary

When a single delta jumped straight past the budget, the breach fired (score 1.0, `critical`) but `clock.warned` stayed `False`. On the following step `clock.breached` made the first condition false, the `elif` was evaluated with `warned` still `False`, and the stale 0.7 warning fired:

```
1.0 critical episode exceeded its 100s wall-clock budget (200s observed)
0.7 warning  episode at 201% of its 100s wall-clock budget
```

Severity runs backwards — a dashboard sees `critical` downgraded to `warning` while the overrun keeps growing — and the warning's own text is self-contradictory.

The breach now marks the warning threshold as passed (`warned = True` alongside `breached`), so a jump straight past the limit emits only the breach, for good — matching what the method's docstring already promised. Issue #224 also suggested checking `TokenRunawayDetector` (cited as the model for this ordering): I verified its warning lives inside the `not-breached` block and cannot re-fire after its breach, so no change is needed there.

## Tests

`test_jump_past_budget_emits_no_stale_warning_afterward` — the issue's exact reproduction (jump to 200 s, then 201 s, 202 s); asserts exactly one `wall_clock_budget` risk: the breach, at `critical` (fails pre-fix, which emits the stale warning on the following step). Full gates pass locally: `pytest -q` (718 passed, 87.50% coverage), `mypy src tests`, `ruff check`, `ruff format --check`.

**Note:** this branch stacks on #256 (both touch `_advance_clock`); if #256 lands first this applies cleanly on top, otherwise it can be rebased.

Fixes #224

### DIFF
--- a/src/snagline/monitor.py
+++ b/src/snagline/monitor.py
@@ -96,7 +96,7 @@ class _EpisodeClock:
 
     def __init__(self, first_ts: float) -> None:
         self.last_ts = first_ts
-        self.elapsed = 0.0  # sum of positive inter-event deltas
+        self.elapsed = 0.0  # wall-clock span covered; last_ts only moves fwd
         self.idle_fired = False  # "idle_gap" fires once per episode
         self.warned = False  # budget warning fires once per episode
         self.breached = False  # budget breach fires once per episode
@@ -397,7 +397,16 @@ class Monitor:
                 self._clocks[event.episode_id] = _EpisodeClock(event.timestamp)
                 return out
             delta = event.timestamp - clock.last_ts
-            clock.last_ts = event.timestamp
+            if delta > 0.0:
+                # Out-of-order or skewed sources (merged adapter streams, clock
+                # skew between hook processes) must not reduce consumed budget
+                # -- and must not rewind ``last_ts`` either: rewinding makes
+                # the *next* event measure its delta from the moved-back
+                # reference, counting an already-counted span a second time
+                # (issue #249). ``last_ts`` only moves forward, so a late
+                # event neither spends nor un-spends budget.
+                clock.elapsed += delta
+                clock.last_ts = event.timestamp
             if (
                 self._idle_warn_seconds is not None
                 and not clock.idle_fired
@@ -415,10 +424,6 @@ class Monitor:
                         event.timestamp,
                     )
                 )
-            if delta > 0.0:
-                # Negative deltas (out-of-order or skewed sources) must not
-                # reduce consumed budget; clamp them out of the accumulation.
-                clock.elapsed += delta
             budget = self._max_wall_seconds
             if budget is not None:
                 if not clock.breached and clock.elapsed >= budget:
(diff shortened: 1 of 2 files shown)


Cyrax321 (approved):
Verified locally: 16/16 pass in test_horizon_time_axis.py, ruff + mypy clean. Marking the warning threshold passed on a jump-over-budget breach stops severity running backwards (critical then stale warning). This branch already contains #256's last_ts fix, so merging #256 first then this one is seamless. Safe to merge after #256.

Issue #90, opened 2026-08-24
[detector] CompactionTripwire: governance-decay detection across context compactions


Cyrax321:
## Summary

Add `CompactionTripwireDetector`: verify that pinned governance constraints survive the host agent's context compactions. Contract: whenever the host harness compacts context, the adapter emits a `compaction` StepEvent carrying hashes of the constraints that MUST remain in-context; the tripwire then requires a corresponding `constraint_present` marker event within N subsequent steps, else fires `governance_decay`.

Detection-only, zero dependencies, and the first runtime tripwire anywhere (open source or commercial) for the compaction-erases-constraints failure class.

## Why

- arXiv:2606.22528 (*Governance Decay*) quantifies it: across 7 models / 1,323 episodes, context compaction raises policy-violation rates from 0% to 30% (up to 59%); when the constraint is dropped by summarization, violations hit 38%. Adversaries can force or steer eviction (volume and summarizer-injection variants defeat every tested model). Compaction happens routinely at 5-20k tokens; every production agent that runs longer than a window is exposed.
- Our privacy stance (hashes only) fits perfectly: constraint canonical texts never reach SNAGLINE, only their SHA-256 pins.
- Nobody in the competitor set (agent-vitals, agentguard, pisama, atlas) or platform layer covers this. First-mover documentation value is real: the phenomenon has a name and a benchmark now.

## Design

New action types (adapter-defined per §4.1 of the spec; core treats them opaquely):

```python
step("compaction", metadata={"pinned": ["<sha256>", ...]})     # harness compacted
step("constraint_present", metadata={"pin": "<sha256>"})        # pin observed in live context
```

```python
class CompactionTripwireDetector:
    name = "governance_decay"
    def __init__(self, grace_steps=3): ...
    def observe(self, event):
        if event.action_type == "compaction":
            self._pending[ep] = {"pins": set(...), "deadline": step_no + grace}
        elif event.action_type == "constraint_present":
            self._pending[ep]["pins"].discard(...)
        # deadline check: any pins remaining -> FailureRisk(score=0.9)
```

Adapter responsibility (documented in ADAPTER_GUIDE + FRAMEWORK_BRIDGES): harnesses that expose compaction hooks (LangGraph pre-compaction callbacks, Claude Code auto-compact hooks) emit the `compaction` event; the presence markers come from whatever introspection the host allows. Where a host offers no visibility, the tripwire is inert-by-design and docs say so honestly.

## Acceptance criteria

[...]

PR #262, opened 2026-09-07
fix(cli): hook bridge dropped side_effect and metadata

### LINKED ISSUE #226: [cli] hook bridge drops side_effect and metadata, making side_effect_guard and compaction_tripwire unreachable over the bridge

Adhi1-2:
## Problem

`_event_to_json` in `src/snagline/cli.py` serialized only 11 of the 13 `StepEvent` fields. `side_effect` and `metadata` were dropped on the way to `--out` (the JSONL file) and `--url` (the forwarded POST body).

Consequence (issue #226): the two host-contract fields that opt into whole detectors never survived the bridge —

- `StepEvent.side_effect=True` (issue #88's contract) lost, so **SideEffectGuardDetector** could never fire over `snagline hook --out/--url`;
- `metadata={"pinned": [...]}` on compaction events (issue #90's contract) lost, so **CompactionTripwireDetector** was structurally unreachable: the written/forwarded event arrived with `metadata={}` and `_open_window` correctly returned `None` (no pins → nothing can decay).

A host doing everything right — marking its payments, emitting pin hashes — got a silent no-op through the bridge.

## Fix

Serialize all 13 fields in `_event_to_json`. `metadata` stays host-provided: the bridge never adds content, it just stops subtracting — whatever the host emits in its payload is what lands in the `--out` file or the POST body. The privacy posture is unchanged and documented in the code comment.

(The issue's second piece — a repeatable `--side-effect-tool NAME` allowlist flag for Claude-Code-native payloads — is deliberately deferred; this PR is the smallest change that makes the bridge honest for canonical-event hosts.)

## Tests

Five regression tests in `tests/test_hook_bridge.py`, all failing on `master` (vacuity-verified by reverting `cli.py` and re-running):

- `test_hook_cli_round_trips_metadata_to_the_out_file` — the `--out` line carries `metadata` and `side_effect` verbatim;
- `test_hook_cli_round_trips_side_effect_to_the_out_file` — the marked flag survives;
- `test_hook_out_file_feeds_compaction_tripwire` — **end-to-end**: a compaction with one never-re-confirmed pin, written through four `hook --out` invocations and replayed with `SNAGLINE_COMPACTION_TRIPWIRE_ENABLED=1`, fires `governance_decay` (silent on master);
- `test_hook_out_file_feeds_side_effect_guard` — **end-to-end**: the same marked non-idempotent action written twice, replayed with `SNAGLINE_SIDE_EFFECT_GUARD_ENABLED=1`, fires `side_effect_duplicate` (silent on master);
- `test_hook_cli_forwards_side_effect_and_metadata_to_sidecar` — the exact POST body through `--url` carries both fields.

Gates: 720 passed / 2 skipped, coverage 87.50%, `mypy src tests` clean, `ruff check` + `ruff format --check` clean.

Fixes #226

### DIFF
--- a/src/snagline/cli.py
+++ b/src/snagline/cli.py
@@ -641,6 +641,13 @@ def _cmd_hook(args: argparse.Namespace) -> int:
 
 
 def _event_to_json(event: StepEvent) -> dict:
+    # All 13 StepEvent fields, so the hook bridge round-trips losslessly.
+    # ``side_effect`` and ``metadata`` were previously dropped, which made
+    # SideEffectGuardDetector and CompactionTripwireDetector structurally
+    # unreachable over `snagline hook --out/--url` (issue #226). ``metadata``
+    # is host-provided: whatever the host emits in its payload is what lands
+    # in the --out file or the POST body (documented privacy posture -- the
+    # bridge never adds content, it just stops subtracting).
     return {
         "step_id": event.step_id,
         "episode_id": event.episode_id,
@@ -653,6 +660,8 @@ def _event_to_json(event: StepEvent) -> dict:
         "error_type": event.error_type,
         "tokens_in": event.tokens_in,
         "tokens_out": event.tokens_out,
+        "side_effect": event.side_effect,
+        "metadata": event.metadata,
     }
 
 
--- a/tests/test_hook_bridge.py
+++ b/tests/test_hook_bridge.py
@@ -255,6 +255,157 @@ def test_hook_cli_never_fails_the_host() -> None:
     assert r3.returncode == 0  # unmapped events are silently dropped
 
 
+# --- issue #226: the bridge must not drop side_effect / metadata --------------
+
+_PIN_A = "448a0c330ba83452956a4b294c1af26dc0def64cffbec53e8e6c99118655dbd1"
+
+
+def _canonical_compaction(step_id: str) -> dict:
+    """A canonical (already-StepEvent-shaped) compaction payload.
+
+    A host emitting canonical events through the command hook carries the
+    tripwire contract fields (action_type "compaction" plus pinned hashes in
+    ``metadata``). Pre-#226 the bridge dropped ``metadata`` on the way to
+    ``--out``/``--url``, so the written/forwarded event lost its pins and
+    CompactionTripwireDetector was structurally unreachable downstream.
+    """
+    from snagline.events import make_signature
+
+    return {
+        "step_id": step_id,
+        "episode_id": "ep-bridge",
+        "timestamp": 1718300100.0,
+        "action_type": "compaction",
+        "action_signature": make_signature("compaction", None),
+        "tool_name": None,
+        "metadata": {"pinned": [_PIN_A]},
+    }
+
+
+def _canonical_side_effect(step_id: str) -> dict:
+    """A canonical non-idempotent action payload (issue #88's host contract)."""
+    from snagline.events import make_signature
+
+    return {
+        "step_id": step_id,
+        "episode_id": "ep-bridge",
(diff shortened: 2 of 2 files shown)


Cyrax321 (approved):
Verified locally: 14/14 pass in test_hook_bridge.py, ruff + mypy clean. Two-line src fix restoring the two dropped StepEvent fields that gated SideEffectGuard and CompactionTripwire over the bridge, with end-to-end replay proofs for both detectors. Safe to merge.

PR #263, opened 2026-09-07
fix(cli): watch finalizes ingested episode ids, not the filename

### LINKED ISSUE #225: [cli] snagline watch calls end_episode() with the filename, so finalize-based detectors never fire (replay does it right)
`snagline watch` calls `monitor.end_episode()` with an id derived from the *filename*, not from the events it ingested. Every finalize-based detector is therefore dead in `watch` unless the operator happens to pass `--episode-id` matching the id inside the file.

### Reproduce

`/tmp/ep.jsonl`:

```json
{"step_id":"1","episode_id":"real-ep","timestamp":1.0,"action_type":"message","action_signature":"a"}
{"step_id":"2","episode_id":"real-ep","timestamp":2.0,"action_type":"tool_call","action_signature":"b","tool_name":"search"}
```

```bash
export SNAGLINE_SILENT_ABORT_ENABLED=1

snagline replay /tmp/ep.jsonl --summary
# {"episode_id": "real-ep", "step_id": "2", "score": 0.7, "trigger": "silent_abort",
#  "detail": "episode ended on 'tool_call', not an output step", "timestamp": 2.0}
# replayed 2 steps; 1 risk(s) emitted

snagline watch --file /tmp/ep.jsonl
# snagline watch: ingested 2 step(s)          <-- nothing emitted

snagline watch --file /tmp/ep.jsonl --episode-id real-ep
# {"episode_id": "real-ep", ... "trigger": "silent_abort" ...}
# snagline watch: ingested 2 step(s)
```

Same events, same config, same monitor composition: `replay` reports the silent abort, `watch` reports nothing.

### Cause

`src/snagline/cli.py:549` and `:575`

```python
episode = args.episode_id or (args.file or "stdin")
...
finally:
    monitor.end_episode(episode)
```

`end_episode("/tmp/ep.jsonl")` finalizes an episode that never existed, so `SilentAbortDetector.finalize` (and any future finalize-based detector) is asked about the wrong key while `real-ep` is never finalized.

`replay` already does this correctly (`src/snagline/cli.py`, `_cmd_replay`):

```python
episodes: set = set()
...
    episodes.add(event.episode_id)
    monitor.ingest(event)
for episode_id in episodes:
    with suppress(Exception):
        monitor.end_episode(episode_id)
```

Relatedly, the `--episode-id` help text at `src/snagline/cli.py:216` promises behaviour that does not exist:

```
help="Episode id override [default: from the events / file name].",
```

Nothing ever reads the id "from the events".

### Suggested fix

Mirror `replay`: collect `event.episode_id` into a set as events are ingested and finalize each one in the `finally` block (keep `suppress(Exception)` so teardown stays fail-open). Keep `args.episode_id` as an override for the case where it is genuinely wanted, and keep a fallback for the "zero events ingested" case. Then either the help text becomes accurate or it should be trimmed.

Note for whoever picks this up: `watch --follow` is long-lived, so "finalize on teardown" is the right hook, but a follow session that sees several episodes will now finalize all of them at exit rather than only one. That is the intended behaviour and worth a test.

### Files

- `src/snagline/cli.py` L216, L549, L575

Adhi1-2:
## Problem

`_cmd_watch` called `monitor.end_episode()` with an id derived from the **filename** (or `--episode-id` / `"stdin"`) — never with the episode ids actually carried by the ingested events:

```python
episode = args.episode_id or (args.file or "stdin")
...
finally:
    monitor.end_episode(episode)
```

`end_episode("/tmp/ep.jsonl")` finalizes an episode that never existed, so every finalize-based detector is dead in `watch`: `SilentAbortDetector.finalize` is asked about the wrong key while the real episode is never finalized. Same events, same config, same monitor composition: `replay` reports the silent abort, `watch` reports nothing — unless the operator happens to pass `--episode-id` matching the id inside the file.

Relatedly, the `--episode-id` help text promised `[default: from the events / file name]` — nothing ever read the id "from the events".

## Fix

Mirror what `_cmd_replay` already does correctly:

- collect `event.episode_id` into a set as events are ingested;
- in the `finally` teardown, finalize each of those ids (with `suppress(Exception)` so teardown stays fail-open);
- still finalize the override/filename fallback id, so the zero-events case and operator overrides keep today's behaviour.

A `--follow` session that sees several episodes now finalizes all of them at exit — that is the intended behaviour (per the issue's note) and is covered by a test.

The help text now says each event's own `episode_id` is the default attribution.

## Tests

Four regression tests in `tests/test_watch_cli.py`; the two core ones fail on `master` (vacuity-verified by reverting `cli.py` and re-running):

- `test_watch_finalizes_ingested_episode_ids_not_the_filename` — the issue's exact repro shape: two events ending on a `tool_call` with `SNAGLINE_SILENT_ABORT_ENABLED=1` fire `silent_abort` for `real-ep` in `watch --file` (silent on master);
- `test_watch_finalizes_every_episode_in_a_multi_episode_file` — two episodes in one file both fire, exactly once each;
- `test_watch_episode_id_override_still_fires_for_overridden_id` — `--episode-id` keeps working;
- `test_watch_zero_events_still_finalizes_the_fallback_id` — empty stdin case unchanged.

Gates: 719 passed / 2 skipped, coverage 87.51%, `mypy src tests` clean, `ruff check` + `ruff format --check` clean.

Fixes #225

### DIFF
--- a/src/snagline/cli.py
+++ b/src/snagline/cli.py
@@ -213,7 +213,8 @@ def _build_parser() -> argparse.ArgumentParser:
     p_watch.add_argument(
         "--episode-id",
         default=None,
-        help="Episode id override [default: from the events / file name].",
+        help="Episode id to attribute events to [default: each event's own "
+        "episode_id; a synthesized id otherwise].",
     )
     p_watch.add_argument(
         "--sink",
@@ -548,6 +549,12 @@ def _cmd_watch(args: argparse.Namespace) -> int:
         heartbeat.touch()  # evidence of life from startup
     episode = args.episode_id or (args.file or "stdin")
     steps = 0
+    # Finalize every episode actually ingested, not the filename (issue
+    # #225): end_episode must be asked about the ids the events carried or
+    # finalize-based detectors (silent_abort) can never fire in watch. A
+    # set, like replay, so a multi-episode file/follow session finalizes
+    # each one exactly once.
+    episodes: set = set()
     try:
         with suppress(KeyboardInterrupt):
             for line in _iter_lines(
@@ -568,11 +575,20 @@ def _cmd_watch(args: argparse.Namespace) -> int:
                     )
                     continue
                 monitor.ingest(event)
+                episodes.add(event.episode_id)
                 steps += 1
                 if heartbeat is not None:
                     heartbeat.touch()
     finally:
-        monitor.end_episode(episode)
+        # Fail-open teardown (issue #225): finalize every episode that was
+        # actually ingested; ``episode`` (the override, the filename or
+        # "stdin") is still finalized so the zero-events case -- and any
+        # operator override -- keeps today's behaviour. Finalize-based
+        # detectors see the right ids; an id already in the set is harmless
+        # (end_episode is idempotent for unknown ids).
+        for episode_id in episodes | {episode}:
+            with suppress(Exception):
+                monitor.end_episode(episode_id)
     print(f"snagline watch: ingested {steps} step(s)", file=sys.stderr)
     return 0
 
--- a/tests/test_watch_cli.py
+++ b/tests/test_watch_cli.py
@@ -94,3 +94,107 @@ def test_iter_lines_calls_on_wait_while_following(tmp_path):
         pass
     assert consumed == ['{"step_id": "s1"}']
     assert len(waits) == 3
+
+
+# --- issue #225: finalize-based detectors must see the real episode ids -------
+
+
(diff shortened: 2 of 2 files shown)


Cyrax321 (approved):
Verified locally: 10/10 pass in test_watch_cli.py, ruff + mypy clean. Real bug - watch finalized the filename instead of the ingested episode ids, so finalize-based detectors (silent_abort) were dead in watch. The replay-parity test plus multi-episode, override, and zero-events cases cover the contract. Safe to merge.

PR #264, opened 2026-09-07
fix(bench): derive the trigger column width instead of hardcoding 16

### LINKED ISSUE #223: [benchmark] detection_accuracy table hardcodes a 16-char trigger column, so side_effect_duplicate and wall_clock_budget rows misalign
The `--format table` renderer in `benchmarks/detection_accuracy.py` pads the trigger column to 16 characters, but two shipped trigger names are longer, so their rows lose column alignment.

### Reproduce

```bash
python benchmarks/detection_accuracy.py --fixtures benchmarks/fixtures --format table
```

```
stagnation          4    0    0      1.000    1.000   1.000
side_effect_duplicate    4    0    0      1.000    1.000   1.000
governance_decay    4    0    0      1.000    1.000   1.000
...
wall_clock_budget    4    0    0      1.000    1.000   1.000
```

`side_effect_duplicate` is 21 chars and `wall_clock_budget` is 17; both push the numeric columns right and break the alignment with the `-----` rules and the header.

### Cause

`benchmarks/detection_accuracy.py:439` / `:445` / `:453`

```python
header = f"{'trigger':<16} {'TP':>4} {'FP':>4} {'FN':>4} {'precision':>10} {'recall':>8} {'f1':>7}"
...
f"{trig:<16} {s.tp:>4} {s.fp:>4} {s.fn:>4}"
```

The width predates the newer triggers in `SHIPPED_TRIGGERS`.

### Suggested fix

Derive the width instead of hardcoding it, e.g.

```python
width = max(len("trigger"), *(len(t) for t in rows))
```

and use `{trig:<{width}}` in the header, both row templates, and the `"-" * n` rules so the table stays aligned as triggers are added. Good first task: one file, purely cosmetic, immediately visible before/after.

### Files

- `benchmarks/detection_accuracy.py` L439, L445, L453

Adhi1-2:
## Problem

The `--format table` renderer in `benchmarks/detection_accuracy.py` padded the trigger column to a hardcoded 16 characters, but two shipped trigger names are longer:

- `side_effect_duplicate` — 21 chars
- `wall_clock_budget` — 17 chars

Both overflowed the cell, pushing their numeric columns right and breaking alignment with the header and the `-----` rules (issue #223):

```
stagnation          4    0    0      1.000    1.000   1.000
side_effect_duplicate    4    0    0      1.000    1.000   1.000
```

The width predates the newer triggers in `SHIPPED_TRIGGERS`.

## Fix

Derive the width instead of hardcoding it — from `SHIPPED_TRIGGERS` **and** any out-of-vocabulary rows the report carries, so an extra-long unshipped trigger also gets an aligned row:

```python
width = max(
    len("trigger"),
    *(len(t) for t in SHIPPED_TRIGGERS),
    *(len(t) for t in report.per_trigger),
)
```

Used in the header and both row templates; the rules already follow `len(header)`. After (from this branch):

```
trigger                 TP   FP   FN  precision   recall      f1
----------------------------------------------------------------
side_effect_duplicate    4    0    0      1.000    1.000   1.000
wall_clock_budget        4    0    0      1.000    1.000   1.000
```

## Tests

Two regression tests in `tests/test_detection_accuracy.py`, both failing on `master` (vacuity-verified by reverting the benchmark file and re-running):

- `test_format_table_aligns_columns_for_long_trigger_names` — the numeric block begins at the same offset in the header and in **every** shipped trigger row, and the rules still span exactly the header width;
- `test_format_table_extra_trigger_widens_the_column` — a 32-char unshipped trigger name fits untruncated and its TP column aligns with the header's.

Gates: 717 passed / 2 skipped, coverage 87.50%, `mypy src tests` clean, `ruff check` + `ruff format --check` clean, accuracy gate (`detection_accuracy.py --fixtures benchmarks/fixtures`) exits 0 (healthy controls that fired: 0).

Fixes #223

### DIFF
--- a/benchmarks/detection_accuracy.py
+++ b/benchmarks/detection_accuracy.py
@@ -435,14 +435,27 @@ def score(outcomes: Sequence[EpisodeOutcome]) -> ScoreReport:
 
 
 def format_table(report: ScoreReport) -> str:
+    # Width is derived, not hardcoded (issue #223): side_effect_duplicate (21)
+    # and wall_clock_budget (17) both overflowed the old fixed 16-char column
+    # and pushed their numeric columns out of alignment with the header and
+    # rules. Deriving from every row's trigger keeps the table aligned as
+    # triggers are added.
+    width = max(
+        len("trigger"),
+        *(len(t) for t in SHIPPED_TRIGGERS),
+        *(len(t) for t in report.per_trigger),
+    )
     lines: list[str] = []
-    header = f"{'trigger':<16} {'TP':>4} {'FP':>4} {'FN':>4} {'precision':>10} {'recall':>8} {'f1':>7}"
+    header = (
+        f"{'trigger':<{width}} {'TP':>4} {'FP':>4} {'FN':>4}"
+        f" {'precision':>10} {'recall':>8} {'f1':>7}"
+    )
     lines.append(header)
     lines.append("-" * len(header))
     for trig in SHIPPED_TRIGGERS:
         s = report.per_trigger.get(trig, TriggerStats())
         lines.append(
-            f"{trig:<16} {s.tp:>4} {s.fp:>4} {s.fn:>4}"
+            f"{trig:<{width}} {s.tp:>4} {s.fp:>4} {s.fn:>4}"
             f" {s.precision:>10.3f} {s.recall:>8.3f} {s.f1:>7.3f}"
         )
     # Triggers outside the shipped vocabulary still get honest rows.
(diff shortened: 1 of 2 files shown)


Cyrax321 (approved):
Bug confirmed by running the bench on base: side_effect_duplicate overflows the hardcoded 16-char column. Fix is correct and handles edges (empty report falls back to header width; unshipped long names widen the column). 16/16 pass in test_detection_accuracy.py, ruff clean. Suggest merging before #265 so the README table can be re-pasted with the new renderer.

PR #269, opened 2026-09-07
fix: restore detector windows at effective size, not base (error_cascade / meltdown / loop)

### LINKED ISSUE #268: load_state truncates scaled windows to base size in error_cascade, meltdown, and loop detectors
## Bug

When window auto-scaling is active (`cfg.window_scale_steps > 0`, issue #92), the live per-episode windows grow with the episode (`effective_window_size`), but `load_state` rebuilds them at the **base** `self.window_size` — silently discarding the tail of every restored window.

Affected `load_state` implementations:

| Detector | Restore site | Effect |
|---|---|---|
| `ErrorCascadeDetector` | `src/snagline/detectors/error_cascade.py:148-151` — `deque(flags, maxlen=self.window_size)` | windowed density alarm loses errors that fell out of the base tail; restored detector under-counts and can miss a cascade live state would catch |
| `MeltdownDetector` | `src/snagline/detectors/meltdown.py:178-181` — `w.push(key, self.window_size)` | `observe` gates on `len(w.window) < target` → restored detector returns `None` for `(effective − base)` steps while entropy is actively collapsed |
| `LoopDetector` | `src/snagline/detectors/loop.py:389-400` — `deque(sigs, maxlen=self.window_size)` for both `_windows` and `_near_windows` | earlier occurrences of a repeating signature are truncated out; restored detector needs extra occurrences before the windowed rule fires |

`StagnationDetector` has the same defect and is tracked separately as issue #218 (PR #219 fixes only `stagnation.py`).

## Reproduction (meltdown)

`meltdown_window_size=20, window_scale_steps=10, max_window=40`, 220 steps in one episode (window grown to 40), then a snapshot → restore while entropy is collapsed and the alarm is re-armed:

* **LIVE** detector: fires immediately (entropy 0.00 bits, 1 distinct in last 40 steps).
* **RESTORED** detector: `risk = None` — and stays silent for **+19 more events** until the rebuilt 20-slot window refills to the effective target of 40.

## Reproduction (loop)

Same scaling config; two occurrences of a repeating signature 13–40 steps back — inside the live 40-window, outside the restored 12-tail:

* **LIVE** fires on the 3rd occurrence.
* **RESTORED** stays silent (the first two occurrences were truncated out) and only fires at what would be a 5th occurrence.

## Impact

A snapshot/restore (monitor restart, sidecar handover, replay from snapshot) mid-episode with scaling on leaves three detectors partially blind exactly when their conditions are active — the window of history that scaling exists to preserve is the part that gets thrown away. Restores with scaling off (the default, `window_scale_steps=0`) are unaffected.

Existing `test_state_round_trip` tests don't catch this because they use the base window size without scaling.

Found with a line-level audit while investigating #218's family of restore bugs.

Adhi1-2:
## Summary

With window auto-scaling on (`window_scale_steps > 0`, issue #92), live detector windows grow with the episode via `effective_window_size` — but `load_state` rebuilt every window at the **base** `self.window_size`, silently truncating the oldest items. A snapshot/restore mid-episode (monitor restart, sidecar handover, replay from snapshot) discarded exactly the history the scaler grew the window to preserve:

- **meltdown** (`meltdown.py`): `observe` gates on `len(w.window) < target`, so the restored detector returned `None` for `(effective − base)` refill steps while entropy was actively collapsed. Traced: live fires immediately, restored stays silent for +19 events.
- **loop** (`loop.py`): `_windows`, `_near_windows`, and `_cycle_windows` all truncated early occurrences out — a loop whose first repeats lived inside the scaled 40-window but outside the base 12-tail fell below `repeat_threshold` after restore. Traced: live fires on the 3rd occurrence, restored is silent.
- **error_cascade** (`error_cascade.py`): windowed density alarms lost errors living inside the scaled window.

## Fix

Each `load_state` now sizes the rebuilt window from the persisted per-episode counts — the same `effective_window_size(base, counts[ep], scale_steps, max_window)` the live path uses — falling back to `len(window)` for pre-#92 payloads that carry no counts. With scaling off (the default), the effective size is always the base, so default-config behavior is unchanged.

Same family as #218 / PR #219, which fixed `StagnationDetector` only; this covers the other three windowed detectors.

## Tests

New `tests/test_scaled_window_restore.py` (8 tests): JSON round-trip + live-vs-restored twin comparisons per detector with scaling on, restored-window length assertions, a scaling-off guard, and a pre-#92 payload fallback. **7 of 8 fail on pre-fix code** (verified by swapping in upstream files; the 8th asserts the unchanged scaling-off path and correctly passes both ways).

Full gates: `pytest -q` 723 passed / 87.51% coverage, `mypy src tests` clean, `ruff check` + `ruff format --check` clean, accuracy gate exit 0 (0 healthy-control firings).

Fixes #268

## Summary by CodeRabbit

* **Bug Fixes**
  * Snapshot restoration now preserves the effective auto-scaled window sizes for error-cascade, loop, and meltdown detection.
  * Restored detectors retain historical events and continue detecting issues consistently after restart.
  * Older snapshots without scaling metadata remain compatible.
  * Fixed restoration behavior when auto-scaling is disabled so windows continue using their configured base size.

* **Tests**
  * Added coverage for scaled-window restoration, restart consistency, and legacy snapshot compatibility.

### DIFF
--- a/src/snagline/detectors/error_cascade.py
+++ b/src/snagline/detectors/error_cascade.py
@@ -20,7 +20,7 @@ from typing import Any
 
 from snagline.config import Config
 from snagline.detectors.base import snapshot_items
-from snagline.detectors.windowing import next_window
+from snagline.detectors.windowing import effective_window_size, next_window
 from snagline.events import StepEvent
 from snagline.risk import FailureRisk
 
@@ -148,8 +148,22 @@ class ErrorCascadeDetector:
         }
 
     def load_state(self, state: dict[str, Any]) -> None:
+        # Restore each window at its effective size, not the base (issue
+        # #268): a snapshot taken mid-episode with auto-scaling on holds
+        # ``effective_window_size(...)`` flags, and a base-sized maxlen
+        # discards the oldest -- errors that lived inside the scaled window
+        # fall out of the density count exactly when a cascade is in
+        # progress. With scaling off the effective size is always the base.
         self._windows = {
-            ep: deque(flags, maxlen=self.window_size)
+            ep: deque(
+                flags,
+                maxlen=effective_window_size(
+                    self.window_size,
+                    int(state.get("counts", {}).get(ep, len(flags))),
+                    self._scale_steps,
+                    self._max_window,
+                ),
+            )
             for ep, flags in state.get("windows", {}).items()
         }
         # Tolerant .get(): pre-#92 snapshots carry no scaler positions.
--- a/src/snagline/detectors/loop.py
+++ b/src/snagline/detectors/loop.py
@@ -46,7 +46,7 @@ from typing import Any, cast
 
 from snagline.config import Config
 from snagline.detectors.base import snapshot_items
-from snagline.detectors.windowing import next_window
+from snagline.detectors.windowing import effective_window_size, next_window
 from snagline.events import StepEvent
 from snagline.risk import FailureRisk, TriggerType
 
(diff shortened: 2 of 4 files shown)


Cyrax321 (approved):
Verified locally: 8/8 pass in test_scaled_window_restore.py, ruff + mypy clean. Extends the #219 effective-window restore to error_cascade/loop/meltdown with tolerant fallbacks for pre-#92 snapshots. Pairs with #219 - merge in either order, no conflicts. Safe to merge.

Cyrax321:
Rebase needed after #233 merged: its snapshot_items() hunks in loop.py / error_cascade.py / meltdown.py now sit adjacent to this PR's load_state hunks. No code changes needed on your side - a straight rebase should resolve it. Still approved from my side.

Cyrax321:
Status check: still conflicted - see my earlier note on rebasing after #233. No change in verdict, still approved.

Adhi1-2:
Rebased onto current upstream master and resolved the snapshot_items conflicts while preserving the approved effective-window restore behavior. Focused validation passes: 8 tests, Ruff check/format, and compileall. Pushed rebased commit e580b7a; please re-check mergeability.

Cyrax321:
Closing as superseded by merged #286, which re-applies this fix onto current master (see that PR for the review and validation). Thank you for the original fix.

PR #271, opened 2026-09-07
fix: global-mode auto-instrumentation actually patches modern SDKs

### LINKED ISSUE #270: Global-mode instrument_openai / instrument_anthropic patch nothing on modern SDKs but still return True
## Bug

The documented P0 one-liner — `docs/INTEGRATION_MATRIX.md`:

```python
instrument_openai(monitor)    # patches openai.OpenAI / AsyncOpenAI globally
```

— produces a **completely unmonitored process that reports success**.

## Root cause

In global mode, `wrap_client` is called with the *class* (`openai.OpenAI`), not an instance, and walks class attributes:

`src/snagline/auto/openai.py:88-101` — for `"chat.completions.create"` it does `getattr(OpenAI, "chat")`, which on every OpenAI SDK since 1.0 returns the `functools.cached_property` *descriptor object*, not a resource instance. `getattr(descriptor, "completions", None)` is `None` → the loop `continue`s → no method is ever wrapped.

`src/snagline/auto/anthropic.py:76-85` — identical structure: `getattr(Anthropic, "messages")` is the `cached_property` descriptor, `getattr(descriptor, "create", None)` is `None`, nothing is wrapped.

Yet `instrument_openai` / `instrument_anthropic` still return `True` ("anything was patched") with no warning logged.

## Verified reproduction

`openai` 3.8.0:
```python
>>> type(openai.OpenAI.__dict__['chat'])
<class 'functools.cached_property'>
>>> instrument_openai(monitor)
True
>>> client = openai.OpenAI(api_key='sk-test')
>>> client.chat.completions.create.__qualname__
'Completions.create'          # NOT the snagline wrapper
```

`anthropic` 1.4.0: same — `type(anthropic.Anthropic.__dict__['messages'])` is `functools.cached_property`, `client.messages.create.__qualname__` stays `'Messages.create'`, return value `True`. Zero events reach the monitor in both cases.

## Why tests miss it

`tests/test_auto_openai.py` / `tests/test_auto_anthropic.py` use fake objects where `chat` / `messages` are **plain class attributes** — exactly the shape `wrap_client` can walk. The real-SDK shape (`cached_property` on the class) is never exercised.

## Impact

Silent total loss of detection on the primary advertised entry point. Per-instance use (`instrument_openai(monitor, client)`) still works — the failure is specific to global mode.

## Suggested fix direction

Either patch the resource classes themselves (`openai.resources.chat.completions.Completions.create` / `anthropic.resources.messages.Messages.create`), unwrap `cached_property` descriptors (`desc.func`) and patch inside, or per-instance-hook the client constructor — and at minimum return `False` / log a warning when the walk wrapped nothing.

Adhi1-2:
## Summary

Global-mode `instrument_openai` / `instrument_anthropic` walked the *client* classes' attributes — but since OpenAI SDK 1.0 and current Anthropic SDKs, `OpenAI.chat` / `Anthropic.messages` are `functools.cached_property` **descriptors**. The walk resolved a descriptor, wrapped nothing, and still returned `True`. The documented P0 one-liner (`docs/INTEGRATION_MATRIX.md`) produced a completely unmonitored process reporting success.

Verified against the real SDKs: `openai` 3.8.0 — `type(openai.OpenAI.__dict__['chat'])` is `functools.cached_property`, `instrument_openai(monitor)` returns `True`, yet `client.chat.completions.create.__qualname__` stays `'Completions.create'` and zero events reach the monitor. Same for `anthropic` 1.4.0.

## Fixes

1. **Patch the resource classes, not the client surface.** Global mode now wraps `Completions`/`AsyncCompletions` (chat + legacy) and `Messages`/`AsyncMessages` — the classes every instance dispatches `create` to. Covers all clients, present and future.
2. **Honest return values.** Re-instrumenting an already-wrapped SDK returns `True` (idempotent — wrappers are marked and skipped so calls never double-count); a genuinely unpatchable SDK returns `False` with a warning instead of a silent success. `wrap_client` also warns when it patches nothing.
3. **Async detection through decorators.** Both SDKs decorate `create` with `@required_args`, which hides the coroutine from `iscoroutinefunction` — so `_wrap_one` picked the sync wrapper, never awaited the call, recorded coroutine-creation latency, and reported `error=False` for every failure. `_wrap_one` now also checks `inspect.unwrap`.

## Tests

New `tests/test_auto_global_mode.py` (9 tests): fake SDK modules with the real-world shapes (`cached_property` client surface, decorated `create` with `__wrapped__`) — resource-class patching, sync + async event emission, async error propagation, idempotency, unpatchable-SDK `False`, and the exact #270 client-walk failure shape. **All 9 substantive tests fail on pre-fix code** (vacuity-verified by swapping in upstream files).

End-to-end verified on real `openai` 3.8.0 and `anthropic` 1.4.0 (sync + async legs, correct `error` flags, idempotent second call).

Full gates: `pytest -q` 732 passed / 87.99% coverage, `mypy src tests` clean, `ruff check` + `ruff format --check` clean, accuracy gate exit 0.

Fixes #270

### DIFF
--- a/src/snagline/auto/anthropic.py
+++ b/src/snagline/auto/anthropic.py
@@ -3,6 +3,12 @@
 Mirrors ``snagline.auto.openai``: wraps ``client.messages.create`` so each
 call emits a ``StepEvent``. Import-safe (no-op when the SDK is absent) and
 handles sync and async clients.
+
+Global mode (issue #270) patches the *resource classes* --
+``anthropic.resources.messages.{Messages, AsyncMessages}`` -- rather than
+walking ``Anthropic.messages``: on modern Anthropic SDKs the client attribute
+is a ``functools.cached_property`` descriptor, so the old class-attribute
+walk resolved a descriptor, not a resource, and silently wrapped nothing.
 """
 
 from __future__ import annotations
@@ -38,9 +44,24 @@ def _emit(monitor, counter, model, tool_name, sig_text, start, error) -> None:
     monitor.ingest(event)
 
 
+def _is_async_call(original) -> bool:
+    """True if calling ``original`` returns a coroutine that must be awaited.
+
+    ``inspect.iscoroutinefunction`` alone misses async methods that carry a
+    decorator (the SDK wraps ``create`` with ``@required_args``), which would
+    misclassify them as sync: the wrapper would never await the call, record
+    only the coroutine-creation latency, and report ``error=False`` for every
+    failure. ``inspect.unwrap`` sees through ``__wrapped__`` chains; on plain
+    (including undecorated async) functions it is the identity.
+    """
+    return inspect.iscoroutinefunction(original) or inspect.iscoroutinefunction(
+        inspect.unwrap(original)
+    )
+
+
 def _wrap_one(monitor, original, tool_name):
     counter = itertools.count()
-    is_async = inspect.iscoroutinefunction(original)
+    is_async = _is_async_call(original)
 
     def _sync(*args, **kwargs):
         sig_text = str(kwargs.get("messages") or args)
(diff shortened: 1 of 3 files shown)


Cyrax321 (approved):
Verified locally: 9/9 pass in test_auto_global_mode.py, ruff + mypy clean. Patching the resource classes instead of walking cached_property descriptors fixes global-mode being a silent no-op on modern SDKs, the double-wrap guard prevents double-counting on re-instrument, and _is_async_call sees through @required_args. P0 fix. Suggest merging before #255 (they conflict; this one's hunks are more localized).

PR #275, opened 2026-09-08
fix(monitor): enforce per-episode retention cap across snapshot restore (#273)

### LINKED ISSUE #273: Monitor.restore() bypasses the per-episode retention cap: restored episodes' detector state is never evicted and retained_episodes lies
## Bug

`Monitor.restore_dict()` loads every snapshot episode's detector state but never registers those episode ids in the `_live_episodes` LRU, so the #184 per-episode retention cap does not apply across a restore:

* the restored episodes' detector windows/CUSUM state are invisible to future eviction — they survive forever while `metrics()["retained_episodes"]` reports only the capped count
* the one eviction the restore *does* perform (the time-axis loop, `monitor.py:882-890`) only pops `self._clocks[evicted]` — it never runs `_evict_episode`, so the evicted id's **detector state** is never reset either

## Reproduction (executed)

Source monitor (`max_live_episodes=100, max_episode_wall_seconds=1000`) ingests 5 episodes A–E, snapshot, restore into `Monitor.default(Config(max_live_episodes=2, ...))`:

```
live episodes:          ['D', 'E']
retained_episodes:      2
loop detector windows:  ['A', 'B', 'C', 'D', 'E']   # <-- 3 over cap, invisible to the LRU
```

Feeding 10 new episodes evicts only the new ids — **A, B, C's detector state survives every eviction, permanently**. With horizon knobs off (the default config), the time-axis loop never runs at all and the same leak occurs with `retained_episodes == 0`-style drift between the gauge and live state.

## Impact

The #184 guarantee — bounded memory; detector state for evicted ids is gone; the LRU evicts the least-recently-seen id — silently fails across a restart. A long-running monitor that snapshots/restores accumulates unbounded detector state for every episode that ever lived in a snapshot, while its retention gauge reports the cap.

## Suggested fix direction

`restore_dict` should register each restored episode id in `_live_episodes` (as `ingest` does) and run the full `_evict_episode` teardown (`detector.reset()` + clock release, not just `_clocks.pop`) for over-cap ids, oldest-first — so the cap holds across restore and the gauge tells the truth.

Distinct from #184 (closed; introduced the cap and the ingest-time eviction — this is a bypass in its interaction with restore), and from #91/#172/#205 (snapshot/restore persistence), #228/#229/#231 (snapshot-time concurrency).

soorajsuru2004-sudo:
Closes #273

## Summary
When restoring a snapshot, `Monitor.restore_dict()` previously loaded detector state for all episodes without registering them into the `_live_episodes` LRU if the horizon time-axis was inactive. Furthermore, the eviction logic only cleared `self._clocks[evicted]` instead of executing `self._evict_episode(evicted)`. This caused over-cap restored episodes to linger indefinitely in detector memory while `retained_episodes` reported only the cap.

## Changes
- **`snapshot_dict()`**: Persists `"live_episodes": list(self._live_episodes.keys())` to preserve exact LRU ordering across restarts.
- **`restore_dict()`**:
  - Gathers all restored episode IDs in ascending LRU order across `live_episodes`, `time_axis` (ordered by `last_ts`), and loaded detectors.
  - Registers them in `_live_episodes` and pops over-cap episodes oldest-first.
  - Calls `self._evict_episode(evicted)` for each over-cap ID, executing `detector.reset(ep)` across all detectors and releasing clocks/locks.
- **Tests**: Added `test_cap_enforced_across_restore_and_detector_freed` and `test_cap_enforced_across_restore_without_horizon_knobs` in `tests/test_per_episode_cap.py` matching the reproduction.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restoring saved monitor state now correctly rebuilds active episode tracking.
  * Per-episode retention limits are enforced during restore, removing excess episode data and associated detector state.
  * Episode clocks and detector windows now remain consistent with the retained episodes, including when horizon limits are disabled.

### DIFF
--- a/src/snagline/monitor.py
+++ b/src/snagline/monitor.py
@@ -758,11 +758,14 @@ class Monitor:
                 "warned": clock.warned,
                 "breached": clock.breached,
             }
+        with self._live_lock:
+            live_episodes = list(self._live_episodes.keys())
         return {
             "format_version": SNAPSHOT_FORMAT_VERSION,
             "detectors": detectors,
             "sinks": sinks,
             "time_axis": time_axis,
+            "live_episodes": live_episodes,
         }
 
     def snapshot(self, path: str) -> None:
@@ -843,21 +846,7 @@ class Monitor:
                 load(dumped_sinks[key])
         time_axis_data = data.get("time_axis")
         if isinstance(time_axis_data, dict):
-            # Restore clocks in order of last_ts ascending so most recent
-            # survives if we must evict under cap. Fall back to sorted keys
-            # for deterministic behavior when last_ts is missing.
-            try:
-                ordered = sorted(
-                    time_axis_data.items(),
-                    key=lambda kv: (
-                        float(kv[1].get("last_ts", 0.0))  # type: ignore[arg-type]
-                        if isinstance(kv[1], dict)
-                        else 0.0
-                    ),
-                )
-            except Exception:
-                ordered = sorted(time_axis_data.items())
-            for episode_id, clock_data in ordered:
+            for episode_id, clock_data in time_axis_data.items():
                 if not isinstance(episode_id, str) or not isinstance(clock_data, dict):
                     continue
                 try:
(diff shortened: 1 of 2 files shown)


yuki-fuyutsuki:
**Deep Review — PR #275 `fix/restore-retention-cap-273`**

Confirmed fix for #273; root cause correctly identified. Manual execution verified.

### What is fixed correctly

* `master` `src/snagline/monitor.py:883-889` only did `self._clocks.pop(evicted, None)` — detector `reset()` + `state.release()` never ran. Leaked memory while `retained_episodes` under-reported. `src/snagline/monitor.py:946` now calls `self._evict_episode(evicted)` — correct. Matches `src/snagline/monitor.py:625` teardown (clocks + `detector.reset` + `release`, fail-open, episode_lock).
* `master` never registered episodes when horizon off (`_clocks` empty, `_live_episodes` empty). New `snapshot_dict()` `src/snagline/monitor.py:762` persists `live_episodes: list(self._live_episodes.keys())` preserving exact LRU order. `restore_dict()` `src/snagline/monitor.py:879-901` rebuilds ordered list from `live_episodes` → `time_axis sorted by last_ts` → detectors, so `Config(max_live_episodes=2)` no longer leaks when `max_episode_wall_seconds=None`. Verified `tests/test_per_episode_cap.py:201` passes.
* Eviction now `popitem(last=False)` oldest-first on `OrderedDict`, retaining newest (`D,E` in `tests/test_per_episode_cap.py:185`). Verified live snapshot `live_episodes:['A','B','C','D','E']` → target `['D','E']`, `self._clocks=={'D','E'}`.

### Remaining issues / fragility

1.  **Backward compat without version bump:** `SNAPSHOT_FORMAT_VERSION=1` `src/snagline/monitor.py:44` unchanged, adds optional `live_episodes`. `restore_dict()` `src/snagline/monitor.py:880` tolerates missing (`isinstance(...,list)`), so old → new and new → old both work. Intentional but should be documented; ambiguous if future incompatible change reuses version 1.

2.  **Brittle detector discovery** `src/snagline/monitor.py:905-930`: hardcoded attr tuple `("_windows","_counts","_fired","_near_windows","_cycle_windows","_tokens","_last","_live","_eps","_episodes","_states")` misses `Loop` extras (`_near_counts`,`_cycle_counts`,`_cycle_fired`,`_stall_*`,`_near_fired`), `Meltdown._clear_streak`, `TokenRunaway._totals/_warned/_breached`, `Stagnation._counts`. Works today only because every active episode has at least `_windows`/`_eps`/`_states`/`_live`/`_last`, so discovery still finds ID. Future detector adding isolated per-episode dict will be missed; cap enforcement would ignore those eps until `time_axis` fallback. Better: extract IDs directly from `dumped_detectors` payload JSON before `load_state()` or introspect all `dict[str,...]` attrs generically.

3.  **`sort_keys=True` destroys LRU for legacy snapshots** `src/snagline/monitor.py:773`: `json.dumps(...,sort_keys=True)` sorts `windows`/`time_axis` keys alphabetically. Without `live_episodes`, LRU lost. New field fixes new snapshots, but `sorted_clocks` fallback `src/snagline/monitor.py:888` (sort by `last_ts`) is correct; detector-only fallback `src/snagline/monitor.py:919` then yields alphabetical order, not true LRU, for interleaved workloads (e.g., ingestion order `A,C,B` → retains `B,C` alphabetically vs `C,B`). Acceptable for legacy migration but worth noting.

4.  **Semantics of `clear()`** `src/snagline/monitor.py:935`: `self._live_episodes.clear()` discards any pre-existing live episodes on the target before re-populating restored IDs. If `restore()` is called on warm monitor (not startup), in-flight episodes silently dropped. `master` merged instead. Assume `restore` is setup-time quiescent op; add docstring/assertion. Dead branch `src/snagline/monitor.py:937` (`if ep in _live_episodes`) can never fire after `clear()` + `seen` dedup — harmless but misleading.

5.  **Concurrency:** `restore_dict()` not documented as exclusive. `clear()` + evictions race with concurrent `ingest()` `src/snagline/monitor.py:335` on `_live_lock`. Should require quiescence.

### Nits

* `_tokens` in attr list never exists. Remove.
* `ordered_episodes` dedup via `seen_episodes` makes `move_to_end` path unreachable — keep only for defensiveness.
* Test `tests/test_per_episode_cap.py:218` only checks `len==2` after horizon-off restore, not eviction of `A,B,C` nor ordering; strengthen to assert exact survivors.

### Verification

Inline reproduction without `pytest` passes both new tests; legacy time-axis → `D,E` retained, horizon-off → 2 retained. Leak no longer reproducible.


[github run](/Cyrax321/SNAGLINE/actions/runs/34260223502)

Cyrax321 (changes requested):
Design is sound and green locally (10/10 per_episode_cap, ruff-format/mypy clean): persisting live_episodes LRU order, rebuilding retention from snapshot+clocks+detector state with backward-compat fallbacks, and evicting over-cap episodes via full _evict_episode instead of dropping just the clock. BLOCKING, trivial: 2x SIM118 in src/snagline/monitor.py - 'for ep in mapping.keys()' (line 920) should be 'for ep in mapping', and 'for key in states.keys()' (line 926) should be 'for key in states'. The ubuntu CI failures are lint-only.

Cyrax321:
Status check: the 2x SIM118 in monitor.py (lines 920/926) are still failing CI, and the branch has since gone conflicted against merged main - needs a rebase plus the two one-word fixes. Structural check confirms restore_dict -> Monitor -> _evict_episode is the single canonical teardown path, which is exactly what this PR routes over-cap restores through. Will merge once fixed + rebased.

Cyrax321:
Closing as superseded by merged #284, which re-applies this fix onto current master (see that PR for the review and validation). Thank you for the original fix.

PR #277, opened 2026-09-09
fix(monitor): reject incompatible restore before mutation

### LINKED ISSUE #236: [core] restore_dict(strict_names=True) applies snapshot state before the composition check rejects it
## Summary

`Monitor.restore_dict(strict_names=True)` validates the detector composition **after** it has already loaded state into the monitors detectors. The per-key / name-suffix fallback loop runs first (`monitor.py:810-835`), and only afterwards does the `strict_names` comparison raise (`monitor.py:847-857`):

```python
for i, detector in enumerate(self._detectors):
    ...
    load(entry)          # state lands in the detector
    consumed.add(matched_key)
...
if strict_names:
    ...
    if expected != current:
        raise ValueError("snapshot detector composition mismatch: ...")
```

So a caller that catches the documented setup-time `ValueError` is left with a monitor whose detectors already absorbed the rejected snapshot's per-episode state (windows, counters). Example: snapshot from `[LoopDetector, ErrorCascadeDetector]` restored into `[ErrorCascadeDetector, LoopDetector]` — the name-suffix fallback matches both detectors by name, both load state, and then the strict check raises. `ErrorCascadeDetector._windows['ep']` and `LoopDetector._windows['ep']` now hold the other monitor's data.

This inverts the feature's own rationale (`monitor.py:754-757`: *"a monitor that silently starts blank is precisely the quiet misbehavior this feature exists to prevent"*) — the monitor silently starts **contaminated** instead. The sink restore (line 858+) correctly runs after the check; only the detector path is misordered.

Found while fixing #217 (same code path); the fix is to run the strict check before the load loop — it reads only the snapshot keys and the detector list, so nothing needs to be applied first.

## Why it matters

`strict_names=True` exists to give operators a loud, safe failure at setup time. A rejected restore that half-applies defeats that contract: the operator retries with a corrected composition and the detectors carry stale foreign windows, skewing loop/cascade/latency detection with no error anywhere.

## Reproduction

```python
import pytest
from snagline.detectors.error_cascade import ErrorCascadeDetector

a = Monitor([LoopDetector(), ErrorCascadeDetector()], [])
for i in range(2):
    a.ingest(StepEvent(step_id=str(i), episode_id="ep", timestamp=float(i),
                       action_type="tool_call", action_signature="q-a",
                       tool_name="search"))
a.snapshot("/tmp/state.json")

b = Monitor([ErrorCascadeDetector(), LoopDetector()], [])
with pytest.raises(ValueError):
    b.restore("/tmp/state.json", strict_names=True)   # raises, as documented
b_loop = b._detectors[1]
assert b_loop._windows == {}      # FAILS: carries the rejected snapshot's window
```

Adhi1-2:
## Summary

Fixes #236.

`Monitor.restore_dict(strict_names=True)` previously loaded matching detector state before validating the snapshot composition. A rejected restore could therefore leave the monitor partially mutated.

## Changes

- Validate the detector composition before calling any detector `load_state()` method.
- Preserve tolerant restore behavior when `strict_names=False`.
- Add a regression test proving a rejected restore leaves the target detector state empty.

## Validation

- `pytest -q --no-cov tests/test_monitor_snapshot.py` — 8 passed
- `ruff check src/snagline/monitor.py tests/test_monitor_snapshot.py` — passed
- `mypy src tests` — passed

Related issue: #236

## Summary by CodeRabbit

* **Bug Fixes**
  * Strict monitor state restoration now rejects detector-composition mismatches before applying any detector state.
  * Failed strict restores no longer leave the monitor partially updated.

### DIFF
--- a/tests/test_monitor_snapshot.py
+++ b/tests/test_monitor_snapshot.py
@@ -172,6 +172,40 @@ def test_composition_mismatch_strict_vs_tolerant(tmp_path):
     )  # cascade has no ep state yet
 
 
+def test_strict_restore_rejection_does_not_apply_detector_state(tmp_path):
+    """A rejected strict restore leaves the target monitor untouched."""
+    path = str(tmp_path / "state.json")
+    source = Monitor([LoopDetector(), ErrorCascadeDetector()], [ListSink()])
+    source.ingest(
+        StepEvent(
+            step_id="0",
+            episode_id="ep",
+            timestamp=0.0,
+            action_type="tool_call",
+            action_signature="q-a",
+            tool_name="search",
+        )
+    )
+    source.ingest(
+        StepEvent(
+            step_id="1",
+            episode_id="ep",
+            timestamp=1.0,
+            action_type="tool_call",
+            action_signature="q-a",
+            tool_name="search",
+        )
+    )
+    source.snapshot(path)
+
+    target = Monitor([ErrorCascadeDetector(), LoopDetector()], [ListSink()])
+    with pytest.raises(ValueError, match="composition mismatch"):
+        target.restore(path, strict_names=True)
+
+    loop = cast(LoopDetector, target._detectors[1])
+    assert loop._windows == {}
+
+
 def test_dedup_sink_cooldown_survives_round_trip(tmp_path):
     from snagline.risk import FailureRisk
 


Cyrax321 (changes requested):
Minimal correct move (validate composition before any load_state) and green locally (8/8 snapshot + 687 full suite, mypy clean). Two things: (1) BLOCKING, trivial: ruff format wants the expected=[...] collapsed to one line in src/snagline/monitor.py (~line 793) - just run ruff format on the file. (2) After #235 merges, please rebase: it touches the same strict_names hunk and its slot-ordered version supersedes this one, so this PR should shrink to the move plus its valuable rejection-leaves-state-empty test.

Cyrax321:
Status check: still needs ruff format on monitor.py (~line 793, collapse the expected=[...] to one line) plus a rebase - the branch is now conflicted (the strict_names hunk moved under merged #235, whose slot-ordered version supersedes). After rebase this should shrink to the move + your rejection test. Will merge once done.

Cyrax321:
Closing as superseded by merged #285, which re-applies this fix onto current master (see that PR for the review and validation). Thank you for the original fix.

PR #278, opened 2026-09-09
fix(sink): preserve rate limiting across batch flushes

### LINKED ISSUE #276: [sink] BatchingSink rate limit resets at every batch boundary
### Issue type

Bug

Yes

### Source

source

### SNAGLINE version

0.1.1.dev0 / 22faeaea0eea8567416fbee78a35b4a6a5603bd8

### Custom code

No

Windows 11 Home 10.0.26200

### Python version

3.13.1

### Current behavior?

`BatchingSink(max_per_second=...)` only enforces the configured delivery gap inside one call to `_deliver()`. `_deliver()` resets its `last` timestamp to `0.0` for every batch, so the first alert in each interval, threshold, or manual flush bypasses pacing.

Two one-item batches can therefore reach the wrapped sink almost simultaneously even though `max_per_second` promises to pace actual deliveries. With `max_per_second=4`, the reproduction below measured a `0.000246s` gap instead of the required `0.25s` minimum.

Expected behavior: the rate limit applies continuously for the lifetime of the `BatchingSink`, including across separate batches and `flush_now()` calls.

Actual behavior: every batch resets the pacing clock, allowing bursts at batch boundaries.

This is distinct from #65 / #72, which enforce `max_batch` and drain queued alerts on close, and from #48, which introduced `BatchingSink`. Searches for `BatchingSink rate`, `batching rate limit`, `max_per_second`, `batch sink pacing`, and `rate limit resets` found no existing issue or open PR for cross-batch pacing.

```python
import time

from snagline.risk import FailureRisk, SEVERITY_INFO


class Recorder:
    def __init__(self):
        self.times = []

    def emit(self, risk):
        self.times.append(time.monotonic())


def risk(i):
    return FailureRisk(
        episode_id="ep",
        step_id=str(i),
        score=0.5,
        trigger="loop",
        detail="d",
        timestamp=time.time(),
        severity=SEVERITY_INFO,
    )


inner = Recorder()
sink = BatchingSink(
    inner,
    max_batch=1000,
    flush_interval=3600.0,
    max_per_second=4.0,
)
try:
    sink.emit(risk(1))
    sink.flush_now()
    sink.emit(risk(2))
    sink.flush_now()
finally:
    sink.close()

gap = inner.times[1] - inner.times[0]
print(f"delivery gap: {gap:.6f}s; expected >= 0.250000s")
assert gap >= 0.24
```

### Relevant log output

```text
delivery gap: 0.000246s; expected >= 0.250000s
  ...
AssertionError
```

Adhi1-2:
## Summary

Fixes #276.

`BatchingSink(max_per_second=...)` previously reset its pacing timestamp for every `_deliver()` call. Separate batches could therefore be delivered back-to-back, bypassing the configured rate limit.

## Changes

- Persist the last delivery timestamp across batches.
- Serialize delivery scheduling and wrapped-sink calls so concurrent flush paths share one pacing sequence.
- Keep pacing state separate from the queue lock so `emit()` remains non-blocking.
- Add deterministic cross-batch regression coverage using a fake monotonic clock.

## Validation

- `pytest -q --no-cov tests/test_batching_sink.py` — 7 passed
- `PYTHONPATH=. pytest -q --no-cov` — 696 passed, 3 skipped
- `ruff check` and `ruff format --check` — passed
- `mypy src tests` — passed

Related issue: #276

### DIFF
--- a/src/snagline/sinks/batching.py
+++ b/src/snagline/sinks/batching.py
@@ -44,6 +44,11 @@ class BatchingSink:
         self._min_gap = 1.0 / max_per_second if max_per_second else 0.0
         self._queue: collections.deque[FailureRisk] = collections.deque()
         self._lock = threading.Lock()
+        # Delivery pacing is shared across every batch and flush path. Keep it
+        # separate from the queue lock so sleeping or calling the wrapped sink
+        # never blocks ``emit``.
+        self._delivery_lock = threading.Lock()
+        self._last_delivery: float | None = None
         self._stop = threading.Event()
         # Set when the queue reaches ``max_batch`` (or on close) so the flusher
         # can wake early instead of sitting out the rest of the interval.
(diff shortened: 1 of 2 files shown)


Cyrax321 (approved):
Verified locally: 7/7 pass in test_batching_sink.py, ruff + mypy clean. Checked the locking by hand: ordering is always queue-lock then delivery-lock (never reversed), close() joins without holding the delivery lock, emit() stays non-blocking, and the queue is cleared under lock before delivery so no double-delivery on the close-vs-drain race. Safe to merge.

PR #279, opened 2026-09-09
fix(config): reject invalid boolean environment values

Adhi1-2:
Environment boolean coercion currently maps every unrecognized value to false. A typo such as SNAGLINE_FAIL_OPEN=flase therefore silently disables fail-open behavior instead of being ignored as documented for values that fail to coerce.

This change accepts explicit true/false aliases and raises ValueError for unknown boolean strings, allowing the existing environment loader to warn and ignore invalid overrides. It adds coverage for invalid values, false aliases, and preserving a config-file value when an invalid environment override is supplied.

Validation: 19 config-loader tests, 697 full-suite tests + 3 skipped, Ruff, mypy src tests, and git diff --check.

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid boolean configuration values are no longer silently interpreted as `False`.
  - Unrecognized boolean values are rejected, preserving the existing configured value instead of overriding it.
  - Additional accepted false-value aliases include `0`, `false`, `no`, `off`, and `f`.

### DIFF
--- a/src/snagline/config.py
+++ b/src/snagline/config.py
@@ -53,7 +53,12 @@ def _coercible_hint(hint: Any) -> Any:
 
 def _coerce(hint: type, value: str) -> Any:
     if hint is bool:
-        return value.strip().lower() in ("1", "true", "yes", "on", "t")
+        normalized = value.strip().lower()
+        if normalized in ("1", "true", "yes", "on", "t"):
+            return True
+        if normalized in ("0", "false", "no", "off", "f"):
+            return False
+        raise ValueError(f"invalid boolean value: {value!r}")
     if hint is int:
         return int(value)
     if hint is float:
--- a/tests/test_config_loader.py
+++ b/tests/test_config_loader.py
@@ -40,6 +40,27 @@ def test_from_env_skips_uncoercible_values():
     assert cfg.cusum_k == 0.5  # default unchanged, no raise
 
 
+def test_from_env_skips_uncoercible_boolean_values():
+    env = {"SNAGLINE_FAIL_OPEN": "flase"}
+    overrides = Config.from_env_overrides(environ=env)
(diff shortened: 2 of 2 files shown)


yuki-fuyutsuki:
**Code Review**

This PR fixes a silent misconfiguration bug where `src/snagline/config.py:54` mapped every unrecognized boolean env string to `False`, so `SNAGLINE_FAIL_OPEN=flase` incorrectly disabled fail-open. It now normalizes with explicit true/false alias sets (`1/true/yes/on/t` and `0/false/no/off/f`) and raises `ValueError` for unknowns, which `Config.from_env_overrides()` already handles via its `except ValueError: logger.warning` path to ignore and preserve the default/file value. The change is narrow, correctly reuses the existing `warn-and-ignore` contract for coercion failures (distinct from loud validation failures), and is well-covered by the three new tests in `tests/test_config_loader.py`. No correctness, security, or resource issues in the code path.

**low**
`README.md:228`
The code now explicitly recognizes `0/false/no/off/f` as false, but the documentation still states `Booleans accept 1/true/yes/on/t` only. This leaves the new alias set undiscoverable and perpetuates the pre-PR doc/code mismatch the PR description calls out. Update the prose to match `src/snagline/config.py:56-60`.
```suggestion
Booleans accept `1/true/yes/on/t` and `0/false/no/off/f`. Unknown keys and values that fail to
```


[github run](/Cyrax321/SNAGLINE/actions/runs/34433308393)

Cyrax321 (approved):
Verified locally: 19/19 pass in test_config_loader.py, ruff + mypy clean. Unknown booleans now raise so the loader warns and ignores them (consistent with existing uncoercible handling) instead of silently mapping typos like SNAGLINE_FAIL_OPEN=flase to False. Safe to merge.

Issue #155, opened 2026-08-26
[adapters] default clock is time.time: Windows tick quantizes sub-15ms latencies to zero, degrading CUSUM input


Cyrax321:
### Context

Discovered while root-causing the first Windows-only flake caught by the new windows-latest CI legs (#116, fixed tests-only in PR #139 commit b42d47c).

`SnaglineCallbackHandler` defaults its clock to wall time:

- src/snagline/adapters/langchain_adapter.py line 57: `self._clock = clock or time.time`

On Windows, `time.time()` advances in roughly 15.6 ms steps (system tick). Consequences for real users on Windows:

- Any tool/LLM call shorter than one tick records `latency_ms == 0.0`.
- Several fast calls record identical timestamps, so computed deltas are quantized to multiples of ~15.6 ms.

This feeds the CUSUM latency-anomaly detector with degenerate samples: during warm-up the running variance collapses toward zero (only saved by the sigma floors), fast operations are indistinguishable from instant failures, and spike detection loses sub-tick resolution entirely. macOS/Linux are largely unaffected since their clocks resolve microseconds.

The flake was exactly this mechanism in test form: a scripted 10 ms sleep measured as latency 0.0 about half the time on the Windows runner.

### Proposal

Default the handler clock to `time.perf_counter` instead: monotonic like the current intent, high resolution on every platform, unaffected by NTP steps (a wall-clock jump would otherwise fabricate negative or huge latencies). The constructor already accepts an injected `clock=`, so this is a one-line default change plus:

1. Audit other adapters for the same pattern (`grep -rn "clock or time.time" src/snagline/adapters/`); autogen/crewai/openai/anthropic adapters likely share it.
2. Check nothing persists or compares `timestamp=` fields across handlers mixing both clocks; StepEvent.timestamp semantics may deserve a docstring note if perf_counter becomes an event timestamp source (perf_counter has no meaningful epoch).
3. Add a deterministic regression test using the scripted-clock pattern from b42d47c asserting sub-millisecond latencies survive the default path.
4. Benchmarks in benchmarks/overhead_benchmark.py already use perf_counter-style measurement; confirm no shared helper needs the same treatment.

Risk assessment: low. Detectors only consume latency_ms differences within one process lifetime, which is exactly what perf_counter guarantees.

Refs #17, Refs #116, Board: #106

Cyrax321:
Claiming #155

Cyrax321:
[...]

PR #281, opened 2026-09-10
fix(auto): measure call latency on perf_counter, not the wall clock (Fixes #280)

### LINKED ISSUE #280: [auto] snagline.auto wrappers still measure latency on time.time: the issue #155 perf_counter fix never audited the auto-instrumentation path
### Issue type

Bug

Yes

### Source

source

### SNAGLINE version

0.1.1.dev0 / commit `540955dbdbe7809053379a01718557d9d953d561`

### Custom code

No

Windows 11 Home (platform-dependent timing; the fix is platform-independent)

### Python version

3.10–3.13 (any; 3.10–3.12 on Windows shows the quantization worst)

### Current behavior?

Issue #155 fixed this for `src/snagline/adapters/` (PR #161, commit `3dfb33a): every adapter clock moved from `time.time()` to `time.perf_counter`, because on Windows the wall clock advances in ~15.6 ms ticks on supported interpreters, so any tool/LLM call shorter than one tick recorded `latency_ms == 0.0` and fed the CUSUM latency detector degenerate samples. Issue #155's proposal step 1 asked to "audit other adapters for the same pattern" — the `snagline.auto` auto-instrumentation wrappers were never part of that audit and still hardcode the wall clock.

All three modules measure call latency (and event timestamps) with `time.time()`:

- `src/snagline/auto/openai.py` — `start = time.time()` in `_sync`/`_async` (lines 76, 90), `latency = (time.time() - start) * 1000.0` in `_emit` (line 39)
- `src/snagline/auto/anthropic.py` — same pattern (lines 69, 83, 33)
- `src/snagline/auto/langchain.py` — same pattern (lines 48, 65, 24)

Consequences for `instrument_openai` / `instrument_anthropic` / `instrument_langchain` users (the "attach to any system, zero call-site edits" path):

- On Windows (Python 3.10–3.12, all supported versions before CPython switched to `GetSystemTimePreciseAsFileTime`), every auto-instrumented call shorter than one system tick records `latency_ms == 0.0`; fast calls are quantized to multiples of ~15.6 ms. The latency/CUSUM detector is starved of usable samples exactly as in #155.
- On every platform and version, `time.time()` is non-monotonic: an NTP correction or manual date change during a call fabricates a negative or hugely inflated `latency_ms` (issue #155's second documented failure mode).

Expected: the auto wrappers use `time.perf_counter` for latency measurement and event timestamps, mirroring the explicit adapters (`src/snagline/adapters/openai.py`, `adapters/anthropic.py`, etc.), whose module docstrings already document exactly this rationale ("Latency measurement and event timestamps use `time.perf_counter`, not `time.time` ... issue #155").

```python
import time
from snagline.auto.openai import wrap_client

# Script the monotonic clock: a call lasting 2**-10 s (exactly 0.9765625 ms).
reads = iter([1024.0, 1024.0 + 2**-10])
time.perf_counter = lambda: next(reads)

class Completions:
    def create(self, **kw):
        return "ok"

class Chat:
    completions = Completions()

class Client:
    chat = Chat()

events = []
mon = type("M", (), {"ingest": staticmethod(events.append)})()
wrap_client(mon, Client()).chat.completions.create(model="gpt-4o", messages=[])

ev = events[0]
assert ev.latency_ms == 0.9765625  # AssertionError: wrapper read time.time, not perf_counter
```

Same failure for `snagline.auto.anthropic.wrap_client` and `snagline.auto.langchain.wrap_client`.

### Relevant log output

```
AssertionError: assert 0.0021457672119140625 == 0.9765625
 +  where 0.0021457672119140625 = <StepEvent latency_ms from the real time.time delta>
```

abyyxhek:
## Summary of Changes

All three `snagline.auto` auto-instrumentation wrappers (`openai`, `anthropic`, `langchain`) still measured call latency and event timestamps with `time.time()`. They now use `time.perf_counter`, taking a single reading in `_emit` shared by the latency delta and the event timestamp, exactly mirroring the explicit adapters.

## Justification

Issue #155 fixed this defect class for `src/snagline/adapters/` (PR #161) and its proposal explicitly asked to audit other modules for the same pattern — but the `snagline.auto` wrappers were never audited. Consequences for every `instrument_openai` / `instrument_anthropic` / `instrument_langchain` user (the zero-call-site-edits attach path):

- On Windows (Python 3.10–3.12, the supported interpreters before CPython switched to `GetSystemTimePreciseAsFileTime`), the wall clock advances in ~15.6 ms ticks, so every auto-instrumented call shorter than one tick records `latency_ms == 0.0` — the CUSUM latency detector is starved of usable samples exactly as documented in #155.
- On every platform, `time.time()` is non-monotonic: an NTP correction or manual date change during a call fabricates a negative or hugely inflated latency.

Closes #280.

## Kind of Contribution

Bug Fix

N/A — no detection-path change; the wrappers swap one clock call for another of equal cost. `snagline bench` measures `Monitor.ingest`, which this change does not touch.

## Unit Tests:

`tests/test_auto_perf_counter_clock.py` (new, 4 tests) reuses the deterministic scripted-clock pattern from `tests/adapters/test_perf_counter_clock.py`: `time.perf_counter` is scripted with binary-exact sub-millisecond readings and `time.time` is never touched, so any wrapper that falls back to the wall clock fails deterministically regardless of platform timing or Python-version clock precision:

- success-path sub-ms latency + timestamp assertions for each of the three wrappers,
- error-path assertion (a raising `create` must still emit a latency from the monotonic clock — the path where a backwards wall-clock step fabricates negative latencies).

All four tests fail on unmodified `master` (`540955d`) and pass with this change.

## Execution Tests

- `pytest tests/ -q --cov-fail-under=0` on Windows 11 / Python 3.13: **808 passed, 2 skipped** (full suite, includes the 23 pre-existing `test_auto_*` tests plus the 4 new ones).
- `ruff check` and `ruff format --check` clean on all touched files.
- `mypy src tests`: only the pre-existing Windows-only `signal.SIGKILL` error in `tests/test_hook_bridge.py:779` (POSIX-only attribute; CI runs mypy on Linux legs only).
- The standalone repro from issue #280 (scripted `perf_counter`, fake client) now asserts `latency_ms == 0.9765625` exactly.

Note: textually adjacent to open PR #255 (stream-emission deferral in the same two files) but orthogonal — #255 restructures emission timing and keeps `time.time`; this PR changes only the clock. Happy to rebase if #255 lands first.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved latency measurements and event timestamps for AI integrations, including more accurate sub-millisecond timing.
  * Prevented incorrect or negative latency values caused by system clock adjustments or coarse clock resolution.
  * Preserved accurate error-event reporting while allowing original errors to propagate normally.

* **Documentation**
  * Clarified timestamp behavior and limitations.

* **Tests**
  * Added coverage for timing accuracy across OpenAI, Anthropic, and LangChain integrations.

### DIFF
--- a/src/snagline/auto/anthropic.py
+++ b/src/snagline/auto/anthropic.py
@@ -9,6 +9,15 @@ Global mode (issue #270) patches the *resource classes* --
 walking ``Anthropic.messages``: on modern Anthropic SDKs the client attribute
 is a ``functools.cached_property`` descriptor, so the old class-attribute
 walk resolved a descriptor, not a resource, and silently wrapped nothing.
+
+Latency measurement and event timestamps use :func:`time.perf_counter`, not
+:func:`time.time`, mirroring the explicit adapters (issue #155): the wall
+clock advances in ~15.6 ms ticks on Windows on supported 3.10--3.12
+interpreters, quantizing sub-tick latencies to zero, and is non-monotonic, so
+a clock step mid-call fabricates negative or huge latencies.
+``perf_counter`` has no meaningful epoch, so these timestamps are only
+comparable within one process; detectors consume them solely as in-process
+latency differences.
 """
 
 from __future__ import annotations
@@ -30,15 +39,15 @@ logger = logging.getLogger("snagline")
 
 
 def _emit(monitor, counter, model, tool_name, sig_text, start, error) -> None:
-    latency = (time.time() - start) * 1000.0
+    now = time.perf_counter()
     event = StepEvent(
         step_id=str(next(counter)),
         episode_id="anthropic-auto",
-        timestamp=time.time(),
+        timestamp=now,
         action_type="tool_call",
         action_signature=make_signature("anthropic_call", model, sig_text),
         tool_name=tool_name,
-        latency_ms=latency,
+        latency_ms=(now - start) * 1000.0,
         error=error,
     )
     monitor.ingest(event)
@@ -66,7 +75,7 @@ def _wrap_one(monitor, original, tool_name):
     def _sync(*args, **kwargs):
         sig_text = str(kwargs.get("messages") or args)
         model = kwargs.get("model", "unknown")
-        start = time.time()
+        start = time.perf_counter()
         error = False
         try:
             result = original(*args, **kwargs)
@@ -80,7 +89,7 @@ def _wrap_one(monitor, original, tool_name):
     async def _async(*args, **kwargs):
         sig_text = str(kwargs.get("messages") or args)
         model = kwargs.get("model", "unknown")
-        start = time.time()
+        start = time.perf_counter()
         error = False
         try:
             result = await original(*args, **kwargs)
--- a/src/snagline/auto/langchain.py
+++ b/src/snagline/auto/langchain.py
@@ -3,6 +3,15 @@
 Wraps the common ``invoke`` / ``generate`` entrypoints on a LangChain model
 or chain so each call emits a ``StepEvent``. Import-safe: a no-op when
 LangChain is absent, and handles synchronous and asynchronous methods.
+
+Latency measurement and event timestamps use :func:`time.perf_counter`, not
+:func:`time.time`, mirroring the explicit adapters (issue #155): the wall
+clock advances in ~15.6 ms ticks on Windows on supported 3.10--3.12
+interpreters, quantizing sub-tick latencies to zero, and is non-monotonic, so
+a clock step mid-call fabricates negative or huge latencies.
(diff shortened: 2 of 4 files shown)


Cyrax321 (approved):
Verified locally: 23/23 pass (new perf_counter tests + all auto suites), ruff + format + mypy clean, CI fully green. Mirrors the #161 precedent for explicit adapters exactly (same docstring, single shared perf_counter reading for latency + timestamp), and the scripted-clock tests fail loudly if any wrapper ever reads time.time again. Safe to merge.

PR #282, opened 2026-09-10
fix(monitor): enforce retention cap across snapshot restore

### LINKED ISSUE #273: Monitor.restore() bypasses the per-episode retention cap: restored episodes' detector state is never evicted and retained_episodes lies

Adhi1-2:
Fixes #273.

Restored detector state must obey the target monitor's per-episode retention cap. This change:

- discovers restored episodes from clocks and detector state;
- rebuilds the live LRU and preserves clock recency;
- evicts over-cap episodes through `_evict_episode()` so detector and clock state are fully released;
- covers horizon-enabled and horizon-disabled restores, cleanup, leak prevention, and under-cap compatibility.

This is based on current `master` and addresses the implementation/rebase issues in the existing #275 attempt.

Validation: focused retention and snapshot tests (24 passed), Ruff check/format, and compileall. Mypy currently reports pre-existing errors in `src/snagline/auto/openai.py` unrelated to this change.

### DIFF
--- a/src/snagline/monitor.py
+++ b/src/snagline/monitor.py
@@ -911,6 +911,7 @@ class Monitor:
             if callable(load) and dumped_sinks.get(key) is not None:
                 load(dumped_sinks[key])
         time_axis_data = data.get("time_axis")
+        restored_clocks: list[tuple[str, _EpisodeClock]] = []
         if isinstance(time_axis_data, dict):
             # Restore clocks in order of last_ts ascending so most recent
             # survives if we must evict under cap. Fall back to sorted keys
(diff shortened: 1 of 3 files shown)


Cyrax321 (commented):
Reviewed against merged #284 (which closed #273): this is a solid rebase and CI is green, but the two implementations now overlap and #284 is the more complete fix - it persists the live-episode LRU order in the snapshot, sweeps all per-episode detector attrs (this version covers only _windows/_eps), and orders the fallback by clock recency. One edge to note here: when the just-added episode itself is the evicted one, its already-loaded detector state is left behind (the evicted != episode_id guard skips cleanup). Closing as superseded by #284 - thank you for the rebase effort and the extra horizon on/off coverage ideas.

Cyrax321:
Closing as superseded by merged #284 (see review above for the comparison). Thanks for rebasing and getting CI green - the effort is appreciated.

PR #294, opened 2026-09-18
fix(cli): make baseline --list-versions read-only on fit and retrain paths

### LINKED ISSUE #293: fix(cli): --list-versions is silently ignored and can write a file (fit) or store a version (retrain)
## Description

`snagline baseline --list-versions` is documented as a read-only listing flag:

> `--list-versions` — With --store-dir: list stored versions and exit (ignores fitting).

In practice the flag is only consulted on the `fit` path when `--store-dir` is set, and is never consulted on the `retrain` path at all. So an operator who asks for a listing can instead get a **write**:

- Without `--store-dir`, `--list-versions` is silently dropped and the command fits a baseline and writes `baseline.json` to the current directory.
- On `baseline retrain`, `--list-versions` is silently dropped and the command retrains and **stores a new version**.

A read-only listing flag having a write side effect is a footgun: it surprises the operator and, in the `retrain` case, mutates the very store the operator was trying to inspect.

## Current Behavior

`--list-versions` is only honoured inside the `if args.store_dir:` branch of `_cmd_baseline` (`src/snagline/cli.py`). Everywhere else it is ignored.

Path 1 — fit without `--store-dir` (silently writes a file):

```
$ snagline baseline w.jsonl --list-versions
  default: n=1 mean=0.0ms std=0.0ms errors=0
snagline baseline: wrote baseline.json
$ ls
baseline.json   <-- the operator asked to list, and got a written file
```

Path 2 — retrain (silently retrains and stores a new version):

```
$ snagline baseline retrain --store-dir store --jsonl w.jsonl --list-versions
snagline baseline retrain: stored version 1789717890.208175 for default/default from w.jsonl (1 tool(s), 1 step(s))
```

## Expected Behavior

`--list-versions` means "list and exit" wherever it is accepted:

- With `--store-dir`: list stored versions and exit 0 without fitting or writing anything (today's behavior — keep it).
- Without `--store-dir`: there is no store to list from, so the command should fail closed with a clear stderr message and a non-zero exit code, the same way other baseline usage errors already do (e.g. `--store-dir is required` on `retrain`). It must not write `baseline.json`.
- On `baseline retrain`: honour the flag the same way — list and exit 0 — instead of retrining.

## Reproduction

```shell
# Path 1: creates baseline.json even though no listing was possible
cd "$(mktemp -d)"
printf '{"step_id":"s1","episode_id":"e1","timestamp":1.0,"action_type":"tool_call","action_signature":"a1","tool_name":"t","latency_ms":10.0}\n' > w.jsonl
snagline baseline w.jsonl --list-versions
ls baseline.json   # exists; should not

# Path 2: stores a new version instead of listing
snagline baseline retrain --store-dir store --jsonl w.jsonl --list-versions
ls store/default/default/versions/
```

## Proposed Solution

Handle `--list-versions` before any fitting/writing in both `baseline` dispatch branches in `src/snagline/cli.py`, reusing the existing list-and-print block. When no `--store-dir` is given, fail closed with a message like `snagline baseline: --list-versions requires --store-dir` and exit code 2, mirroring the existing `--store-dir is required` usage error.

## Acceptance Criteria

- [ ] `snagline baseline TRAJ --store-dir DIR --list-versions` lists versions and exits 0 without writing anything (unchanged).
- [ ] `snagline baseline TRAJ --list-versions` (no `--store-dir`) exits non-zero with a clear message and writes no file.
- [ ] `snagline baseline retrain --store-dir DIR --jsonl F --list-versions` lists versions and exits 0 without storing a new version.
- [ ] Regression tests added for all three cases.
- [ ] Existing tests pass.
- [ ] Lint / type checks pass.

Adhi1-2:
## What

Fixes #293.

`snagline baseline --list-versions` is documented as read-only, but it was only consulted inside the fit path's `if args.store_dir:` branch and never on the `retrain` path. So an operator asking for a listing could instead get a **write**:

- `baseline TRAJ --list-versions` (no `--store-dir`) silently fit a baseline and wrote `baseline.json` to the cwd.
- `baseline retrain --store-dir DIR --jsonl F --list-versions` silently retrained and **stored a new version**, mutating the very store the operator was trying to inspect.

## Change

In `_cmd_baseline` (`src/snagline/cli.py`), handle `--list-versions` before any fitting, writing, or retraining — before the `retrain` dispatch — and extract the existing list-and-print block into a read-only `_list_stored_versions` helper:

- With `--store-dir`: list stored versions and exit 0, fitting/writing/storing nothing (unchanged behavior, now also reached from the `retrain` form).
- Without `--store-dir`: fail closed with `snagline baseline: --list-versions requires --store-dir` and exit 2, mirroring the existing `--store-dir is required` usage error on `retrain`. No file is written.

The `--list-versions` help text is updated to say it is read-only, requires `--store-dir`, and works on both forms.

## Acceptance criteria from #293

- [x] `baseline TRAJ --store-dir DIR --list-versions` lists versions and exits 0 without writing anything (unchanged).
- [x] `baseline TRAJ --list-versions` (no `--store-dir`) exits non-zero with a clear message and writes no file.
- [x] `baseline retrain --store-dir DIR --jsonl F --list-versions` lists versions and exits 0 without storing a new version.
- [x] Regression tests added for all three cases (plus the retrain-without-store case).
- [x] Existing tests pass — full suite: 815 passed, 3 skipped, 89.33% coverage.
- [x] `ruff check src tests`, `ruff format --check src tests`, and `mypy src tests` all clean.

## Verification

Reproduced the issue's exact commands against the patched CLI:

```
$ snagline baseline w.jsonl --list-versions            # was: wrote baseline.json
snagline baseline: --list-versions requires --store-dir
$ echo $?
2
$ ls baseline.json
ls: baseline.json: No such file or directory

$ snagline baseline retrain --store-dir store --jsonl w.jsonl --list-versions
snagline baseline: no stored versions for default/default   # fresh store; no version stored
$ echo $?
0
```

## Summary by CodeRabbit

- **Bug Fixes**
  - `snagline baseline --list-versions` now works consistently for fitting and retraining.
  - Listing versions is read-only: it exits successfully without fitting, retraining, or saving data.
  - The command now reports a clear usage error when `--list-versions` is used without `--store-dir`.
- **Documentation**
  - Updated the unreleased changelog with the corrected baseline version-listing behavior.


Cyrax321 (approved):

This PR makes --list-versions read-only on fit and retrain paths. All CI tests pass with 815 tests. The change fixes a real issue where listing could trigger writes. Well tested.

thank again

Cyrax321 (approved):
## Review — PR #294: fix(cli) make `baseline --list-versions` read-only

Checked out `pull/294/head` and reviewed against `origin/master`. 4 files, +114/-14.

### Verification performed
- ✅ **Full suite: 853 passed, 3 skipped** (100s) — no regressions
- ✅ `tests/test_baseline.py` + `tests/test_baseline_retrain.py`: 25 passed
- ✅ `tests/test_cli.py`: 18 passed
- ✅ All 4 new regression tests pass individually
- ✅ `ruff check src tests` clean, `ruff format --check` clean (139 files)
- ✅ `mypy src tests` — no issues in 139 files
- ✅ **Mutation-tested the fix**: reverted `cli.py` to `origin/master` and re-ran the new tests — **3 of 4 fail on unpatched code**, then pass again once the fix is restored. The tests genuinely detect the bug, not just the patch.
- ✅ **Manual CLI exercise**: reproduced both issue scenarios — `--list-versions` without `--store-dir` exits 2 with `--list-versions requires --store-dir` and writes no `baseline.json`; `retrain ... --list-versions` lists and exits 0 without storing a version.

### Assessment

The fix is correct and minimal. Two things make it good rather than just functional:

1. **The dispatch point is right.** `--list-versions` is now handled at the very top of `_cmd_baseline`, *before* the `trajectory == "retrain"` dispatch and before any fitting or writing. Because `fit` and `retrain` share one subparser (the literal `retrain` is a positional value, not a separate command), a single guard covers both paths — matching the "wherever it is accepted" contract from #293. Reordering rather than adding a second guard in the retrain function is the cleaner choice.
2. **The error handling follows house style exactly.** The new `--list-versions requires --store-dir` branch uses `file=sys.stderr` + `return 2`, mirroring the sibling `--store-dir is required` usage errors at `cli.py:740-743`. The extracted `_list_stored_versions` helper is pure read-only by construction.

### Findings

**1. Stray commit-hash fragment welded onto an unrelated changelog entry** — *minor, but it should be fixed*

`CHANGELOG.md`, last line of the `### Fixed` section:

```diff
-  knob (#317).
+  knob (#317). 57305ac (fix(cli): make --list-versions read-only on fit and retrain paths)
```

The new bullet for this PR was added correctly at the top of the section, but a second copy of the commit subject — prefixed with a raw short SHA, `57305ac` — was appended to the tail of the *pre-existing* `#317` token-budget entry. That entry is about `episode_token_budget` range-checking and has nothing to do with this change.

The appended line reads as a malformed sentence fragment in the released changelog, and the `57305ac` SHA doesn't even match this PR's actual head (`f5824c3`), so it's not a useful reference either. Recommend deleting ` 57305ac (fix(cli): make --list-versions read-only on fit and retrain paths)` from the end of the `#317` bullet, leaving it as `...naming the knob (#317).`

This is the only real defect I found; the code itself is sound.

**2. Observation, not a defect** — `_list_stored_versions` constructs a `BaselineStore` with `max_versions=args.max_versions or 10` purely to list. This is unchanged from the original code (the same construction was there before the refactor), and listing never prunes, so it's harmless. Flagging only so it's a conscious carry-forward rather than an accident — a `list_versions`-only accessor wouldn't need the retention cap at all.

### Verdict

✅ **Approve**, with one trivial cleanup requested: strip the stray `57305ac (...)` fragment from the `#317` changelog bullet (finding 1). It's cosmetic, so I'd merge after that one-line edit — or merge as-is and fix in a follow-up if the queue is moving. The functional change is correct, well-tested, and the tests provably catch the original bug.

Issue #299, opened 2026-09-18
perf: episode_lock generator machinery costs ~0.34 us/step, 38% of the default ingest budget


Adhi1-2:
## Summary

`MemoryStateBackend.episode_lock` is a `@contextmanager`. Every `ingest()` call pays for one generator frame — create, `send` to the `with`, `throw`/close on exit — on top of the two `_meta` lock acquisitions and the per-episode `RLock`. That generator machinery is the single largest fixed component of the default per-step cost.

This is on the **default** path: every step, every episode, scaling off. It is not a corner case.

## Measurement (upstream `master`, `PYTHONPATH=src`)

Isolated microbenchmark of the lock object itself, 500k iterations:

| shape | cost |
|---|---|
| raw `RLock` acquire/release (uncontended) | 0.053 µs |
| 2x plain `Lock` acquire/release | 0.058 µs |
| `@contextmanager` + 2 `_meta` + `RLock` (current) | **0.708 µs** |
| same logic as `__enter__`/`__exit__`, no generator | 0.372 µs |

So the generator alone accounts for roughly **0.34 µs** of every ingest. Against the published default of ~1.9 µs/step, the lock machinery is **~38% of the per-step budget**, and about half of it is pure generator overhead rather than actual synchronization.

## Why it matters

`ingest()` is the hot path — one call per agent step, by construction. The lock is correct and necessary (it shards by episode instead of serializing everything, which was the verified bottleneck in monitor.py). But the cost is being paid in a place where the default config gets no benefit from the generator form: the caller always does `with backend.episode_lock(...)` and never needs the generator to be a generator.

## Fix direction

Provide a non-generator entry point — a small class with `__enter__`/`__exit__` that does the same `_meta`-protected entry/waiter bookkeeping and `RLock` handoff — and have `episode_lock` return it. This keeps the `with backend.episode_lock(eid):` call sites in monitor.py untouched and keeps the `StateBackend` protocol working for backends that stay generator-based (Redis). The waiter/release safety logic from issue #238 is unchanged; only the calling convention loses a generator frame.

Measured: 0.708 -> 0.372 µs, a ~47% reduction in lock cost, or roughly **0.34 µs/step (~18%) off the default ingest path**.

## Environment

Python 3.10+, Apple silicon, upstream `master`. Microbenchmark repeated 500k iterations per shape; the ordering was stable across runs.

Adhi1-2:
[...]

PR #311, opened 2026-09-18
perf(state): episode_lock holds a plain object, not a generator

Adhi1-2:
## issue #299 — `MemoryStateBackend.episode_lock` pays a generator frame every ingest

`episode_lock` is a `@contextmanager`. Every agent step calls it, so every step pays for generator machinery — create the frame, resume into the `with`, throw on exit — on top of the `RLock` it wraps. The caller only ever enters and exits, so the generator buys nothing.

Measured in isolation, interleaved A/B (median of many runs, same machine):

| call | before | after |
|---|---|---|
| `episode_lock`, same episode | 0.708 µs | **0.396 µs** (−44%) |
| `episode_lock`, across 2000 episodes | 0.796 µs | **0.489 µs** |

The generator frame alone was ~0.34 µs of the ~1.9 µs **default** per-step budget — roughly a fifth of the headline number the README publishes.

## the change

`episode_lock` resolves the entry, counts the caller as a waiter, and returns a `_HeldLock` — a plain object with `__enter__`/`__exit__` that acquire and release the `RLock`. No frame is ever created.

The `StateBackend` protocol widens from `Iterator[None]` to `AbstractContextManager[None]`, which is what a caller actually relies on. A backend that prefers to `yield` still type-checks, because `contextlib._GeneratorContextManager` satisfies that shape — `RedisStateBackend` is untouched.

## correctness: the issue #238 invariant is unchanged

`release()` can run while another thread is parked on an episode's lock. The fix keeps the exact semantics that make that safe:

- the entry is resolved **and** the waiter counted under `_meta`, *before* the `RLock` is acquired — a parked waiter is already accounted for when `release()` looks at the count
- the entry leaves the table only once `waiters` drains to zero
- `__exit__` re-fetches the entry under `_meta` and tolerates a `None` lookup — `release()` may have already drained it

The old `@contextmanager` did all of this in its `finally`; the object mirrors it field-for-field.

## validation

- **Concurrency stress:** 16 threads × 4000 steps each across 40 episodes, with `end_episode` interleaved — zero errors, mutual exclusion on a shared episode never violated, lock table drained to the retention cap.
- **Tests:** 202 state / monitor / lock / episode / concurrency tests pass (`-k "state or lock or episode or monitor or concurrent or race"`).
- **Types:** `mypy src tests` — no new errors (only the 7 pre-existing in `auto/openai.py`).
- **End-to-end:** ingest median 1.88–1.92 → **1.53 µs/step** (~19% off the default path).

No behavior change; defaults unchanged.


Cyrax321 (approved):

This PR removes the generator overhead from episode_lock. CI tests pass. The change is performance focused with proper tests. Well done.

thank again

Issue #315, opened 2026-09-18
bug: snapshot/restore silently drops a deferred latency baseline-shift signal


Adhi1-2:
## the bug

`_WelfordCUSUM` carries `pending_drift` / `pending_old_mu` / `pending_shift` for a documented deferral: when a re-fitted baseline shifts beyond the alarm bar on the same step a CUSUM alarm already holds the single risk-return slot, the shift is held until the first quiet step and emitted then.

None of the three fields is written by `_state_to_dict` or read by `_state_from_dict` (`detectors/latency_anomaly.py:322`/`:354`). **A snapshot taken between detection and emission loses the signal permanently** — after restore the detector has no idea a shift was pending, and the "baseline shifted" risk is never emitted.

Repro, `cusum_k=0.5, cusum_h=2.0, min_samples=3, refit_every=4`, sequence `[12, 8, 8, 100, 100, 100, 100, 10, 100, 200, 100]` leaves `pending_drift=True, old_mu=9.33, mu0=70`. After `dump_state`/`load_state`, `pending_drift` is `False`. Feeding `[70.0]*6`:

- never-restarted detector → `latency_anomaly 0.8 "latency baseline shifted 9ms -> 70ms (61ms move, beyond the 2.0x-sigma alarm bar); baseline re-fitted"`
- restored detector → **nothing**

## scope

Gated on the opt-in `cusum_refit_every > 0`. Severity: **wrong detection** — a silently missed alert, in the feature whose entire purpose is making baseline drift visible instead of learned away.

## the fix

The three fields join the existing `refit_every > 0`-gated block in the serialization pair, so default-config snapshots stay byte-identical to pre-#92 payloads and old payloads load cleanly (tolerant `.get()`).

Also fixed in the same PR: `restore_dict` *merged* `_clocks` instead of replacing it, leaving orphan clocks for episodes whose detector state had just been wiped — a live monitor restored onto a snapshot that omits an episode left it resuming at a stale `elapsed` with the warning latch set while every detector treated it as brand new. Every other collection in `restore_dict` is replaced wholesale; `_clocks` now is too.

(PR to follow.)

Adhi1-2:
Fixed by #316 (merged): `pending_drift` / `pending_old_mu` / `pending_shift` are now serialized in the `refit_every > 0`-gated block (`detectors/latency_anomaly.py`) and `restore_dict` clears `_clocks` before loading instead of merging (`monitor.py`).

[...]

Issue #136, opened 2026-08-25
[sink] DedupSink cooldown snapshot over-suppresses after host reboot: restored monotonic timestamps sit in the clock's future


Cyrax321:
## Summary

`DedupSink.dump_state()` / `load_state()`, shipped with #91, persist raw `time.monotonic()` cooldown timestamps. That is correct for the primary restart-survivable scenario (process restart within the same boot, where the monotonic clock is continuous), but restoring a snapshot written before a **host reboot** leaves every restored key far in the future relative to the fresh clock.

## Failure mode

`emit()` suppresses when `last is not None and (now - last) < self._cooldown`. After a reboot:

- old snapshot timestamp `last` = previous uptime (say 30 days);
- new `now` starts near zero;
- `now - last` is large-negative, which is `< cooldown`, so the alert is suppressed;
- suppression lasts until `now` reaches `last + cooldown`: roughly the previous uptime, i.e. days to weeks of silent alert loss for every restored key.

This is exactly the fail-silent behavior an alert-suppression wrapper must not have (see the module docstring's own rationale against wall-clock steps).

## Why existing tests miss it

`test_dedup_sink_cooldown_survives_round_trip` restores within the same process, where monotonic values are still valid. Nothing simulates a clock reset.

## Suggested fixes (any one)

1. Store `time.time()` alongside monotonic in `dump_state`; in `load_state`, shift restored monotonic entries by the observed wall-clock skew.
2. On `load_state`, drop entries whose wall-clock age already exceeds `cooldown_seconds`.
3. Minimum viable: document same-boot-only semantics in both docstrings plus the ATTACH_ANY_SYSTEM hardening checklist, and skip restoration (log once, fail open) when the stored wall-clock age exceeds the cooldown.

Per the #91 contract, restore remains a setup-time operation; whichever option lands must not turn a stale snapshot into a raised exception inside ingest.

## Verification sketch

New test: write a snapshot, patch `time.monotonic` to return a much smaller value (simulating a reboot), restore, emit a duplicate risk, assert it is delivered rather than suppressed.

Refs #91. Board: #106

Cyrax321:
Claiming #136

Cyrax321:
Done: PR #163 merged. Wall-clock companion in dump_state, provenance-aware fail-open restore per the issue sketch.

PR #316, opened 2026-09-18
fix: anchor the restored time axis; persist deferred baseline shifts

### LINKED ISSUE #314: bug: a restored time axis fabricates (or freezes) the episode budget across a restart
## the bug

`snapshot_dict` persists each episode's `_EpisodeClock.last_ts`, and `restore_dict` reloads it verbatim. But `last_ts` is a raw `StepEvent.timestamp`, and the shipped auto-instrumentation stamps events with `time.perf_counter()` (`auto/openai.py:61`, `auto/anthropic.py:55`, `auto/langchain.py:37`) — a clock the code itself documents as having *"no meaningful epoch … only comparable within one process"*.

Snapshot/restore exists precisely for process restarts (issue #91). The restored `last_ts` is meaningless in the new process, and the first post-restart event measures its delta from it. It fails in **both** directions:

**False criticals** — source ingests one event at ts `0.4`, snapshot, restore into a fresh monitor, ingest one event at `40000.0` (new-process uptime). One healthy event emits two risks:

```
idle_gap          0.8  "no events for 39999.6s (threshold 30.0s)"
wall_clock_budget 1.0  "episode exceeded its 100s wall-clock budget (40000s observed)"
```

**Missed breach (mirror case)** — the source ran to `last_ts=3095.0` with 95s of a 100s budget consumed. After restore, every new-process timestamp is *below* the restored `last_ts`, so `delta <= 0`, `elapsed` never advances, and **200s of real new time produces no breach**. The budget is silently extended.

The codebase already handles exactly this clock-reset hazard for `DedupSink` (`sinks/dedup.py`, issue #136 — *"entries sitting in the future of the current monotonic clock … dropped"*). The omission in the time axis is an oversight, not a design choice.

## scope

Only reachable with opt-in horizon knobs (`max_episode_wall_seconds` / `idle_warn_seconds`) **and** restore-on-startup. Severity: **wrong detection** — both false criticals and missed breaches.

## the fix

A restored clock is now marked unanchored. The first post-restore event re-anchors `last_ts` and contributes no delta — exactly as an episode's first event always has — instead of measuring a span from a dead process's epoch. Accumulated `elapsed` is preserved across the restart, so budget already spent is not forgiven; no span is invented either.

(PR to follow.)

Adhi1-2:
Fixes #314 and #315.

## what was wrong

**1. A restored time axis invented (or froze) the episode budget** (#314). `restore_dict` reloaded `_EpisodeClock.last_ts` verbatim, but `last_ts` is a raw `StepEvent.timestamp` and the shipped auto-instrumentation stamps events with `time.perf_counter()` — a clock documented in-tree as having *"no meaningful epoch … only comparable within one process"*. Snapshot/restore exists for process restarts, so that first post-restore delta was measured against a dead epoch. It broke both ways:

- new process uptime 40000s → one healthy event emitted `idle_gap 0.8 "no events for 39999.6s"` **and** a critical `wall_clock_budget 1.0 "…(40000s observed)"`
- source ran long (e.g. `last_ts=3095.0`) → every new timestamp is *below* the restored value, `delta <= 0`, `elapsed` never advances, **200s of real new time breached nothing** — the budget was silently extended

The codebase already handles this exact clock-reset hazard for `DedupSink` (`sinks/dedup.py`, issue #136); the time axis was an oversight.

**2. `restore_dict` merged `_clocks` instead of replacing it** (#315). Every detector's state is rebuilt wholesale and the LRU is cleared, so a clock surviving for an episode the snapshot does not carry was orphaned — resuming at a stale `elapsed` with the warning latch set while every detector treated it as brand new.

**3. A deferred latency baseline-shift vanished across a restart** (#315). `pending_drift`/`pending_old_mu`/`pending_shift` — the deferral that holds a shift found on the same step a CUSUM alarm owned the risk slot — were never serialized. A snapshot between detection and emission permanently lost the "baseline shifted" risk.

## the fix

- A restored clock is built **unanchored**; the first post-restore event re-anchors `last_ts` and contributes no delta, exactly as an episode's first event always has. Accumulated `elapsed` survives the restart, so budget already spent is not forgiven and no span is invented.
- `restore_dict` now clears `_clocks` under the lock before loading, matching how every other collection is handled.
- The three deferred-shift fields join the existing `refit_every > 0`-gated serialization block, so default-config snapshots stay byte-identical to pre-#92 payloads and old payloads load cleanly via tolerant `.get()`.

## validation

- 6 new regression tests in `tests/test_horizon_time_axis.py` and `tests/test_window_scaling.py`, covering both failure directions, the budget-preserved property, the orphan-clock case, and the byte-identical default snapshot.
- Full suite: **833 passed** (up from 827), coverage **87.88%**, 1 skipped. The only 3 failures are pre-existing and unrelated (they fail identically on clean `upstream/master` — two are environmental: this dev venv has the SDKs the "absent SDK" test expects to be missing, and lacks the `snagline` console script).
- `mypy src tests`: no new errors (7 pre-existing in `auto/openai.py`).

Reachable only with the opt-in horizon knobs / `cusum_refit_every` plus restore-on-startup; default-config behavior is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved deferred baseline-shift alerts when detector state is saved and restored, ensuring they are delivered as expected after restart.
  * Improved monitoring after restoration by preventing inaccurate idle-gap or time-budget risks caused by process restarts.
  * Preserved elapsed budget usage across restores and prevented stale episode timing data from affecting monitoring.
  * Restored monitoring clocks using the first new event as a reliable timing reference.


Cyrax321 (approved):

This PR fixes issues with restored time axis and deferred baseline shifts. CI tests pass with 833 tests. The changes handle process restarts correctly and persist deferred state across snapshots. Well tested.

thank again

Issue #132, opened 2026-08-25
[config] StagnationDetector validation edge cases: min_novelty=0 silently disables firing, out-of-range env values crash Monitor.default()


Cyrax321:
Found during post-merge verification of #87 (landed in PR #127, merged as 4c76f9f). Filing separately because neither behavior blocks the acceptance criteria and the right resolution deserves its own discussion. No prior art found after searching open+closed issues (closest are #66 on env-key semantics and #101 on auto-calibration, both different topics).

## Behavior A: min_novelty=0.0 is accepted but makes the detector unable to ever fire

The constructor validates `0.0 <= min_novelty <= 1.0`, so `Config(stagnation_min_novelty=0.0)` passes. But the stale test in `observe()` is:

```python
w.novel_in_window / self.window_size < self.min_novelty
```

A novelty rate is always >= 0.0, so with `min_novelty=0.0` the condition is never true, `stale_windows` never increments, and the detector stays silent forever. A user reading the config comment ("stale when fewer than this share new") and setting 0 expecting "alert me when novelty hits exactly zero" gets a silently disabled safety net. Inconsistent with `patience`, where the degenerate value raises `ValueError`.

## Behavior B: out-of-range env values raise ValueError out of Monitor.default()

`Config.from_env_overrides()` only rejects values that fail coercion (logged and dropped, per issue #66 semantics). A value like `SNAGLINE_STAGNATION_MIN_NOVELTY=99` coerces cleanly into the Config, and then `StagnationDetector.__init__` raises ValueError when `Monitor.default(config=...)` constructs it:

```python
import os
os.environ["SNAGLINE_STAGNATION_MIN_NOVELTY"] = "99"
from snagline import Monitor
from snagline.config import Config
Monitor.default(config=Config.from_env())   # ValueError: min_novelty must be within [0.0, 1.0]
```

Before #127, nothing on the `Monitor.default()` path could raise, so this adds an environment-driven startup-crash path to the host agent. That sits awkwardly next to project.md principle 2 (monitoring must never crash the host) even though construction-time failure is arguably fail-fast rather than runtime fail-open. Same exposure exists for `stagnation_window_size=0` and `stagnation_patience=0` via env.

## Possible resolutions (to be argued)

1. Fail fast, but centrally: validate ranges once in `Config.resolve()` (or a `validate()` hook) with a clear error message, and document that invalid monitoring config aborts startup loudly rather than silently disabling detection.
[...]

Issue #119, opened 2026-08-25
[config] Make SNAGLINE_LOG_FORMAT / Config.log_format operational in Monitor.default() and CLI sink selection


Cyrax321:
## Context

#99 delivered the JSON-lines logging sink (#111, merged): `LoggingSink` + `JsonRiskFormatter`, plus a new `Config.log_format` knob (`"text"` | `"json"`, env `SNAGLINE_LOG_FORMAT`, config-file key). The implementing task had a strict file allowlist (`sinks/`, `config.py`, `tests/sinks/`, README row, project.md section 8), so the knob is currently **documented but not consumed anywhere in core**:

- `Monitor.default()` always installs `ConsoleSink()` unless callers pass `sinks=` explicitly (monitor.py, default()).
- `snagline watch` / `serve` CLI paths append `ConsoleSink()` directly (cli.py).
- Net effect: setting `SNAGLINE_LOG_FORMAT=json` today changes nothing unless a user constructs `LoggingSink` manually in code.

This issue tracks making the 12-factor key operational end to end.

## Tasks

- [ ] `Monitor.default(config)`: when `config.log_format == "json"`, install `LoggingSink` per the agreed composition semantics (issue #99 wording said emission happens "alongside console", so decide: replace console output, or add LoggingSink next to it)
- [ ] CLI (`watch`, `replay` if applicable, `serve`): honor the knob via `Config.resolve()` so config file -> env layering works identically everywhere
- [ ] Validate `log_format` against `{"text","json"}` at construction or resolve time: today any other string is silently accepted and undefined (documented in the field comment); prefer a loud failure or a logged fallback over silent no-op
- [ ] Docs: README has no consolidated environment-variable/config-key table; consider adding one listing `SNAGLINE_*` keys including `SNAGLINE_LOG_FORMAT` (the "config table" the original spec referenced; #111 documented the key in the sinks table row instead)

## Acceptance

- Setting `SNAGLINE_LOG_FORMAT=json` with zero code changes makes risk emission machine-readable JSON lines through the standard run paths.
- An invalid value fails loudly or falls back with a warning, never silently.
- No new dependencies; fail-open behavior of the sink itself unchanged.

Refs #99, #111

Board: #106

Cyrax321:
Duplication audit (against all open AND closed issues and PRs):

- #99 (closed): the parent feature, delivered by #111; this issue is its operational follow-up, cross-linked both ways. Not a duplicate.
[...]

PR #318, opened 2026-09-19
fix(detector): validate the token-runaway budget envelope knobs (#317)

### LINKED ISSUE #317: fix(detector): token-runaway budget envelope knobs are unvalidated and page at critical on the first step
### Issue type

Bug

Yes (commit `90b3519`)

### Source

source

### SNAGLINE version

0.1.1.dev0 (commit `90b35191c74bc4a14f08e7f66e23970be19050bb`)

### Custom code

No

Windows 11 Home Single Language 10.0.26200

### Python version

3.13.1

### Current behavior?

`TokenRunawayDetector`'s budget envelope has two unvalidated knobs, and both
fail in the false-positive direction — a mis-typed value pages at **critical**
severity instead of erroring at startup:

1. `episode_token_budget=0` (or any value `< 0`) fires a score-1.0
   `budget_breach` on the **first** token-bearing step of an episode:
   `total >= self.budget` is `10 >= 0`, always true.
2. `token_budget_warn_fraction=0.0` (or any value `<= 0.0`) fires a score-0.8
   `token_runaway` warning on the first token-bearing step:
   `threshold = budget * 0.0` is `0`, so `total >= 0` is always true.

Both reach the detector through `Config` unchanged. `Config.__post_init__`
range-checks the wall-clock analogue (`warn_fraction` is restricted to
`(0, 1]` and `max_episode_wall_seconds` must be positive when set, in
`_validated_horizon`) and the stagnation knobs (`_validated_stagnation`), but
the token envelope has no equivalent check.

A value above range is the mirror problem: `token_budget_warn_fraction=1.5`
makes `threshold = budget * 1.5 > budget`, so the pre-breach warning is
structurally unreachable and the episode silently loses its early signal —
the warning can never fire before the breach.

### Expected behavior

An out-of-range token-envelope knob is a configuration error and should fail
loudly at construction / `Config.resolve()` time with a message naming the
knob — the policy already established by `log_format` (#119), the horizon
knobs (#92) and the stagnation knobs (#132): *"an out-of-range value is a
configuration error and fails loudly at construction/resolve time instead of
being silently ignored downstream."*

Specifically:

- `episode_token_budget` must be `None` (disabled, as documented) or a
  positive integer. `0` is not a meaningful budget — a run cannot be allotted
  zero tokens — and it is far more likely a mistyped/unset "disable".
- `token_budget_warn_fraction` must be within `(0.0, 1.0]`, exactly like the
  wall-clock `warn_fraction`. `0.0` makes the condition fire unconditionally;
  values above `1.0` make it unreachable.

### Steps to Reproduce

```python
from snagline import Monitor
from snagline.sinks.base import AlertSink


class Collect(AlertSink):
    def __init__(self):
        self.risks = []

    def emit(self, risk):
        self.risks.append(risk)


def run(budget, warn_fraction):
    sink = Collect()
    cfg = Config(
        token_runaway_enabled=True,
        episode_token_budget=budget,
        token_budget_warn_fraction=warn_fraction,
    )
    m = Monitor.default(config=cfg, sinks=[sink])
    m.ingest(
        StepEvent(
            step_id="1",
            episode_id="ep",
            timestamp=1.0,
            action_type="tool_call",
            action_signature=make_signature("tool_call", "t", "1"),
            tool_name="t",
            tokens_in=10,
        )
    )
    return [(r.trigger, r.score, r.detail) for r in sink.risks]


print(run(1000, 0.0))  # critical warning on the FIRST step
print(run(0, 0.8))     # critical breach on the FIRST step
print(run(1000, 1.5))  # warning can never fire at all
```

Also reachable from the environment, no code needed:

```
SNAGLINE_TOKEN_RUNAWAY_ENABLED=true SNAGLINE_EPISODE_TOKEN_BUDGET=0
```

### Relevant log output

```
budget=1000 warn_fraction=0.0 -> [('token_runaway', 0.8, 'episode at 1% of its 1000-token budget')]
budget=0    warn_fraction=0.8 -> [('budget_breach', 1.0, 'episode exceeded its 0-token budget (10 observed)')]
budget=1000 warn_fraction=1.5 -> []
```

One step, ten tokens, and two of the three configurations emit a **critical**
risk. The control (`budget=1000, warn_fraction=0.8`) correctly emits nothing.

## Root Cause

`src/snagline/detectors/token_runaway.py` consumes both knobs with no bounds
checking:

```python
if self.budget is not None:              # line 73: 0 passes this guard
    total = self._totals.get(ep, 0) + step_tokens
    self._totals[ep] = total
    if not self._breached.get(ep, False):
        if total >= self.budget:         # line 77: 10 >= 0 -> always true
            ...                          # score 1.0 budget_breach
        threshold = self.budget * self.warn_fraction   # line 88: = 0.0
        if not self._warned.get(ep, False) and total >= threshold:  # always true
            ...                          # score 0.8 token_runaway
```

`Config` never validates either field, so the bad value survives
`Config.__post_init__`, `Config.resolve()` (file + env layering) and
`Monitor.default()`, and only surfaces as a bogus critical alert at runtime.
The detector's own docstring says the budget is "one warning ... and a single
critical breach"; with these values the "budget" is meaningless.

## Proposed Solution

Add a `_validated_token_runaway(cfg)` helper to `src/snagline/config.py`
mirroring `_validated_horizon` / `_validated_stagnation`, and call it from
both `Config.__post_init__` and the end of `Config.resolve` (where env/file
layering via `setattr` bypasses `__post_init__`):

- raise `ValueError` when `episode_token_budget is not None and
  episode_token_budget <= 0`;
- raise `ValueError` when `token_budget_warn_fraction` is outside
  `(0.0, 1.0]`, reusing the exact wording of the `warn_fraction` check.

The knobs are opt-in (`token_runaway_enabled` defaults to `False`), so stock
configurations never hit the new checks — same "defaults are always valid"
property the existing validators rely on.

## Acceptance Criteria

- [ ] Problem is reproducibly fixed.
- [ ] Regression test added.
- [ ] Existing behavior remains intact (valid budgets, including
      `episode_token_budget=None`, are unchanged).
- [ ] Tests pass.
- [ ] Lint/format checks pass.
- [ ] No unrelated changes.

abyyxhek:
## Summary of Changes

`TokenRunawayDetector`'s budget envelope accepted two knobs that make it emit
**critical**-severity false positives, and a third value that silently disables
its early warning:

- `episode_token_budget=0` (or negative) → `total >= budget` is `10 >= 0`, so a
  score-**1.0** `budget_breach` fires on the *first* token-bearing step.
- `token_budget_warn_fraction=0.0` (or negative) → the warning threshold is `0`,
  so a score-**0.8** `token_runaway` warning fires on the first step.
- `token_budget_warn_fraction>1.0` → the warning threshold sits above the budget,
  so the pre-breach warning can never fire before the breach silences it.

All three now fail loudly at configuration time with an error naming the knob,
instead of paging at critical severity at runtime.

## Justification

A mis-typed threshold is an operator mistake, not a production incident — but
the detector turned it into a critical alert, which is exactly the failure mode
this library exists to avoid false signals in. The knobs reach the detector
through `Config`, `Config.resolve()` (file + env layering) and
`Monitor.default()`, and none of them range-checked either field.

This closes the last gap in the configuration-validation policy the project
already established: `log_format` (#119), the horizon knobs (#92) and the
stagnation knobs (#132) are all range-checked, and the wall-clock
`warn_fraction` — the direct analogue of `token_budget_warn_fraction` — is
already restricted to `(0, 1]`. The token envelope simply had no equivalent
check. Everyone who turns on `token_runaway_enabled` benefits.

## Kind of Contribution

Bug Fix, Tests

## Root Cause

`src/snagline/detectors/token_runaway.py` consumed both knobs with no bounds
checking:

```python
if self.budget is not None:               # 0 passes this guard
    total = self._totals.get(ep, 0) + step_tokens
    if total >= self.budget:              # 10 >= 0 -> always true
        ...                               # score 1.0 budget_breach
    threshold = self.budget * self.warn_fraction   # 0.0 * budget == 0.0
    if not self._warned.get(ep, False) and total >= threshold:  # always true
        ...                               # score 0.8 token_runaway
```

`Config.__post_init__` and `Config.resolve()` never validated either field, so
the bad value survived every layering step and only surfaced as a bogus
critical alert.

## Changes

- `src/snagline/config.py` — new `_validated_token_runaway(cfg)` mirroring the
  existing `_validated_horizon` / `_validated_stagnation` validators:
  `episode_token_budget` must be `None` (documented: disables the envelope) or
  positive; `token_budget_warn_fraction` must be within `(0, 1]`, reusing the
  `warn_fraction` wording. Wired into `__post_init__` **and** the end of
  `resolve()`, because env/file layering applies values via `setattr`, which
  bypasses `__post_init__`. Field comments now state the ranges.
- `src/snagline/detectors/token_runaway.py` — the constructor duplicates the
  check, following the `StagnationDetector` precedent (#132): direct
  construction with explicit kwargs skips the `Config` check entirely.
- `tests/test_config_loader.py`,
  `tests/detectors/test_token_runaway.py` — regression coverage (see below).
- `CHANGELOG.md` — entry under `[Unreleased] / Fixed`.

The knobs are opt-in (`token_runaway_enabled` defaults to `False`) and their
defaults (`None`, `0.8`) are valid, so stock configurations never hit the new
checks — the same "defaults are always valid" property the existing validators
rely on.

## Unit Tests

Added 5 tests, all verified to **fail before** the fix and pass after:

- `tests/test_config_loader.py::test_token_runaway_envelope_out_of_range_aborts_startup`
  — both knobs at construction, via `Config.resolve()` env layering, and via
  direct detector construction; plus a positive case pinning that valid values
  (including `episode_token_budget=None` disabling the envelope) still survive
  every layer.
- `tests/detectors/test_token_runaway.py::test_zero_budget_does_not_page_at_critical_on_the_first_step`
  — end-to-end through `Monitor.default()`: the rejected config dispatches no
  risk at all.
- `tests/detectors/test_token_runaway.py::test_nonpositive_budget_is_rejected`
- `tests/detectors/test_token_runaway.py::test_warn_fraction_out_of_range_is_rejected`
- `tests/detectors/test_token_runaway.py::test_none_budget_still_disables_the_envelope`
  — regression guard that the new check does not tighten the documented
  "None disables the envelope" contract.

Pre-existing envelope behaviour is unchanged:
`test_envelope_warns_once_then_breaches_once`, `test_reset_clears_envelope_and_cusum`,
`test_state_round_trip_preserves_behavior` and `test_events_without_tokens_are_ignored`
still pass unmodified.

## Execution Tests

The original reproduction from the issue, one 10-token step under three
mis-configurations plus a valid control:

| configuration | before | after |
| --- | --- | --- |
| `budget=1000, warn_fraction=0.0` | `token_runaway` 0.8 "at 1% of budget" | rejected: `token_budget_warn_fraction must be within (0, 1]` |
| `budget=0, warn_fraction=0.8` | `budget_breach` 1.0 | rejected: `episode_token_budget must be positive when set` |
| `budget=1000, warn_fraction=2.0` | (silent false negative) | rejected: `token_budget_warn_fraction must be within (0, 1]` |
| `budget=1000, warn_fraction=0.8` (control) | no risk | no risk |

Also verified through the environment, no code needed:
`SNAGLINE_TOKEN_RUNAWAY_ENABLED=true SNAGLINE_EPISODE_TOKEN_BUDGET=0` now
aborts startup with the same clear error instead of paging.

## Testing

Commands actually executed (Python 3.13.1, Windows 11, against upstream
`master` at `90b3519`):

- `python -m pytest tests/ -q` — **PASS** (834 passed, 2 skipped, coverage 87.94% ≥ 80% gate; baseline was 829 passed)
- `ruff check src tests` — **PASS** (All checks passed)
- `ruff format --check src tests` — **PASS** (137 files already formatted)
- `mypy src tests` — **PASS** (1 pre-existing error in `tests/test_hook_bridge.py:779`, `signal.SIGKILL` is not available on Windows; unrelated to this change and present on `master`)
- `python benchmarks/detection_accuracy.py --fixtures benchmarks/fixtures --format table` — **PASS** (CI false-positive gate: **0 healthy controls fired**, macro-F1 0.951, exit 0)
- `git diff --check` — **PASS**

## Checklist

- [x] Issue reproduced
- [x] Root cause identified
- [x] Fix implemented
- [x] Regression test added/updated
- [x] Tests pass
- [x] Lint passes
- [x] Formatting checked
- [x] Type checking passes
- [x] Build passes
- [x] No unrelated changes
- [x] No conflicts

Closes #317

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid token budget settings are now rejected during configuration, including zero or negative episode budgets and warning fractions outside the range `(0, 1]`.
  - Configuration errors identify the specific setting that needs correction.
  - Valid settings continue to work across file, environment, and direct configuration, while omitting the episode budget preserves the option to disable that limit.

- **Documentation**
  - Updated release documentation to describe the new validation rules.


Cyrax321 (approved):

This PR validates token-runaway budget knobs at configuration time. All CI checks pass with 834 tests. The fix adds proper range validation matching existing pattern. Well implemented with good test coverage.

thank again

PR #319, opened 2026-09-19
fix(state): warn when redis backend is requested without a URL (#308)

### LINKED ISSUE #308: [state] SNAGLINE_STATE_BACKEND=redis with no URL silently degrades to in-memory with no warning
## Description

`default_state_backend()` in `src/snagline/state.py:146` reads `SNAGLINE_STATE_BACKEND`, and the `redis` arm is gated on `SNAGLINE_STATE_REDIS_URL` being non-empty:

```python
if kind == "redis":
    url = os.environ.get("SNAGLINE_STATE_REDIS_URL")
    if url:
        try:
            return RedisStateBackend(url)
        except ImportError:
            logger.warning("snagline: redis backend requested but redis not installed; "
                           "falling back to in-memory state")
    ...
return MemoryStateBackend()
```

When the URL is missing, the inner `if url:` is false and control falls straight through to `return MemoryStateBackend()` — with **no log line at all**. Every other degradation path in this function is announced (`redis not installed` → warning); this one is silent.

That silence is the dangerous part. The redis backend exists to coordinate episodes across *processes*; in-memory state is per-process. So a misconfigured deployment silently gets N workers each holding their own lock — no coordination, no deadlock, no error, nothing in the logs. The `except ImportError` arm shows the project already intends to fail loud on this exact class of misconfiguration; the missing-URL case just was not given the same treatment.

README:315-316 documents the pair, and the URL column reads *"Redis URL when the backend is `redis`"* — i.e. required in that mode, not optional.

## Current Behavior

```python
>>> os.environ["SNAGLINE_STATE_BACKEND"] = "redis"   # no SNAGLINE_STATE_REDIS_URL
>>> from snagline.state import default_state_backend
>>> type(default_state_backend()).__name__
'MemoryStateBackend'      # no warning logged
```

## Expected Behavior

Emit a warning for the same reason the `ImportError` arm does — the user asked for `redis` and got something else — before falling back. Something like:

```python
if kind == "redis":
    url = os.environ.get("SNAGLINE_STATE_REDIS_URL")
    if not url:
        logger.warning(
            "snagline: redis backend requested but SNAGLINE_STATE_REDIS_URL is "
            "unset; falling back to in-memory state"
        )
    else:
        ...
```

Keep the fallback (failing hard would break a working memory-mode deployment that merely sets the backend name), just stop being quiet about it.

A test belongs in `tests/test_state_backend.py` next to `test_default_state_backend_redis_when_configured` (which only covers the ImportError path, since `# pragma: no cover` on `RedisStateBackend` shows redis is not installed in CI): assert `SNAGLINE_STATE_BACKEND=redis` with no URL returns a `MemoryStateBackend` *and* emits a record at WARNING via `caplog`.

abyyxhek:
## Summary

`default_state_backend()` now logs a warning when `SNAGLINE_STATE_BACKEND=redis`
is requested but `SNAGLINE_STATE_REDIS_URL` is unset, instead of silently
falling back to in-process memory.

## Related Issue

Closes #308

## Root Cause

In `src/snagline/state.py`, the `redis` arm of `default_state_backend()` gated
on the URL being non-empty:

```python
if kind == "redis":
    url = os.environ.get("SNAGLINE_STATE_REDIS_URL")
    if url:                                     # <-- false when unset
        try:
            return RedisStateBackend(url)
        except ImportError:
            logger.warning("redis backend requested but redis not installed; ...")
return MemoryStateBackend()                     # <-- reached with no log line
```

When the URL was missing, `if url:` was false and control fell straight
through to the `MemoryStateBackend()` return with **no log line at all** —
even though the `except ImportError` arm right beside it deliberately warns on
the exact same class of misconfiguration.

That silence is the dangerous part. The redis backend exists to coordinate
episodes *across processes*; in-memory state is per-process. So a misconfigured
deployment silently gets N workers each holding their own lock — no
coordination, no deadlock, no error, and nothing in the logs. `README.md`
documents the pair with the URL described as *"Redis URL when the backend is
`redis`"*, i.e. required in that mode rather than optional.

## Solution

Split the `redis` arm so a missing URL takes the warning path explicitly,
mirroring the wording of the existing `ImportError` warning:

```python
if not url:
    logger.warning(
        "snagline: redis backend requested but SNAGLINE_STATE_REDIS_URL "
        "is unset; falling back to in-memory state"
    )
else:
    try:
        return RedisStateBackend(url)
    except ImportError:
        ...  # unchanged
```

Fallback behaviour is otherwise byte-identical: the same `MemoryStateBackend()`
is returned, so no deployment changes shape. The warning is scoped to an
explicit `redis` request, so the documented `memory` default and an unset
`SNAGLINE_STATE_BACKEND` stay quiet and never add noise to a plain
`Monitor.default()`.

## Testing

- [x] Unit tests — `python -m pytest tests/ -q`
- [x] Lint — `ruff check src tests`
- [x] Format — `ruff format --check src tests`
- [x] Type check — `mypy src tests`
- [x] Manual reproduction of the original report

Results:

- **Tests:** `831 passed, 2 skipped` (was `829 passed, 2 skipped` on
  pristine `90b3519`), coverage `87.92%` — the 80% gate is met.
- **Ruff check / format:** clean (the one reformat ruff requested was applied
  to the new test).
- **mypy:** the single reported error,
  `tests/test_hook_bridge.py:779: Module has no attribute "SIGKILL"`, is
  pre-existing on this Windows host — I confirmed it reproduces identically on
  a pristine `master` checkout (`signal.SIGKILL` does not exist on win32). It
  is in a file this PR does not touch.

Reproduction before/after (issue #308's report):

```
$ SNAGLINE_STATE_BACKEND=redis, SNAGLINE_STATE_REDIS_URL unset
before:  backend = MemoryStateBackend, logs = <<nothing>>
after:   backend = MemoryStateBackend, logs = "snagline: redis backend
         requested but SNAGLINE_STATE_REDIS_URL is unset; falling back to
         in-memory state"
```

The new `test_default_state_backend_warns_when_redis_url_missing` fails on
unmodified `master` (`AssertionError: []` — no warning) and passes with this
change. A second test, `test_default_state_backend_quiet_when_redis_not_requested`,
guards the warning's scope so the happy path can never grow noisy.

## Scope

Two files changed, both narrow:

- `src/snagline/state.py` — 12 lines: the warning arm, plus a comment on why.
- `tests/test_state_backend.py` — two new tests alongside the existing
  `default_state_backend` tests.

No public signature, config field, fallback behaviour, README, or unrelated
module was changed. This issue was already filed and verified by another
contributor (#308), so no duplicate issue was created.

## Conflict Check

Branch is based on the latest upstream `master` (`90b3519`, rebased and
confirmed unchanged during development) with no unresolved conflicts.


Cyrax321 (approved):

This PR adds a warning when redis backend is requested without a URL. CI tests pass with 831 tests. The change is narrow and prevents silent misconfiguration. Well documented and well tested.

thank again

Issue #313, opened 2026-09-18
perf: make_signature spends 0.88 us of its 1.21 us in json.dumps


Adhi1-2:
## the finding

`make_signature` is the other half of the per-step cost the published number does not count — adapters call it once per step, before constructing the `StepEvent`. It is **1.21 µs/step**, and **0.88 µs of that is `json.dumps`**, which is called purely to produce a canonical, unambiguous serialization of the parts before hashing.

Measured on this machine (median of 7 × 300k, same 3-part input):

| | cost |
|---|---|
| `json.dumps` alone | 0.884 µs |
| full `make_signature` (json + encode + sha256 + hexdigest) | **1.214 µs** |
| same, length-prefixed encoding | **0.719 µs** (−41%) |

The SHA-256 work itself is ~0.25 µs and irreducible — the avoidable part is the JSON encoder.

## why json is there

The docstring (`src/snagline/events.py:79`) is explicit and correct: JSON array encoding is unambiguous, `"a", "b||c"` can never be confused with `"a||b", "c"`. The earlier delimiter-joined scheme was replaced for exactly this reason (issue #15). Any replacement must preserve that property.

A length-prefixed encoding (`b"%d:%s" % (len(part), part)` per part, concatenated) is equally unambiguous and needs no escaping — I verified the two collision cases above stay distinct. But **it changes every signature value**, which is a compatibility break: restored snapshots carry old hashes in their loop windows, and a mixed old/new population would silently mis-detect.

## options

1. **Accept the cost.** Canonical JSON is defensible: it is obviously correct, already audited, and 1.2 µs is small next to any real agent step.
2. **Length-prefixed encoding** (−0.5 µs/step). Needs a snapshot-format version bump and a migration story for on-disk state, or a clean break documented in the changelog.
3. **Cache by parts tuple** — signatures repeat heavily in real traffic (a small tool palette over a long episode). A bounded LRU keyed on the parts tuple would make hot signatures ~0.05 µs and leave the wire format untouched. Costs a lock or a per-episode cache, and turns a stateless hash into something with an eviction policy. Probably the best ratio of win to risk if you want it.

Not PRing without a steer — the correctness risk in options 2/3 sits on the detection contract, which is yours to weigh.

For scale: with `StepEvent` construction (#312), this is ~2 µs/step of caller-side overhead against `ingest`'s ~1.5 µs.

Issue #312, opened 2026-09-18
perf: StepEvent's frozen dataclass __init__ costs 0.52 us/step, 13% of the integrator's per-step budget


Adhi1-2:
## the finding

Every adapter builds a `StepEvent` per agent step, then calls `Monitor.ingest`. The published per-step number (median 1.7–2.4 µs) only measures `ingest` — it excludes construction, which the integrator also pays for on every step.

`StepEvent` is `@dataclass(frozen=True, slots=True)`. A frozen dataclass's generated `__init__` routes every one of its 13 field assignments through `object.__setattr__(self, name, value)` so it can enforce immutability. That is roughly 4× the cost of a normal attribute store.

Measured on this machine (median of 7 × 100k constructions, same args, back to back):

| constructor | cost/step |
|---|---|
| `@dataclass(frozen=True, slots=True)` (current) | **0.847 µs** |
| `@dataclass(slots=True)` (mutable) | **0.331 µs** |
| hand-written `__slots__` class | **0.136 µs** |

For context, `Monitor.ingest` is ~1.5 µs/step after #299. So `StepEvent` construction is **~35% of what `ingest` costs**, and ~13% of the integrator's total per-step budget (construct + ingest ≈ 3.9 µs).

## why it is frozen

Immutability is a deliberate contract, not an accident — `README.md:639` says *"The system is built on immutable dataclasses"*, and the frozen bit is what makes a `StepEvent` hashable and stops an adapter from mutating an event detectors have already seen. That is a real guarantee, so this is a tradeoff to decide rather than a bug to fix silently.

## options

1. **Accept the cost.** The guarantee is worth 0.52 µs to you. Fair.
2. **Drop `frozen`, keep `slots`.** –0.52 µs/step, but events become mutable and unhashable. Nothing in `src/` depends on either: `grep` finds no `asdict`/`replace`/`fields`/`is_dataclass` use against `StepEvent` (those are all on `Config`), no `StepEvent` in a set or as a dict key, and no test asserts `FrozenInstanceError`. The README line would need updating and adapters would need a "don't mutate" note.
3. **Hand-written `__slots__` class** with the same field order and defaults (–0.71 µs). Biggest win, biggest diff; loses the free `__repr__`/`__eq__` unless re-added.
[...]

Issue #310, opened 2026-09-18
[docs] The hand-maintained test count has gone stale again (7 places, incl. project.md at 385)


Adhi1-2:
## Description

This is the recurrence of #96, which fixed the numbers but not the cause: the test count is hand-maintained in prose, so it drifts on every PR that adds tests. It has drifted again, and further this time.

Every place a count is hardcoded:

| Location | Says | Actual (master, today) |
|---|---|---|
| `README.md:18` | badge `tests-715 passed` | 827 passed |
| `README.md:48` | "715 tests, zero required deps" | 827 |
| `README.md:331` | "715 passed, 2 skipped" | 827 passed, 1 skipped |
| `README.md:767` | "715 tests passing, 2 skipped, 87.50% line coverage" | 827 passed, 1 skipped |
| `README.md:795` | "715 passed, 2 skipped" | 827 passed, 1 skipped |
| `project.md:917` | "385 tests passing, 1 skipped" | 827 passed, 1 skipped |

`project.md` is two stale-counts behind (385 → 715 → 827), which is the clearest proof the pattern doesn't self-correct.

The count is also **environment-dependent**, which is why even a corrected number is fragile: I measure 827 passed / 1 skipped / **3 failed** in this checkout, but all three failures are local-environment artifacts that pass in CI — one needs an installed console script (`pip install -e .`), one needs the `openai` package (CI does not install it), and one is #296 against a CONTINUUM checkout. So "the number" is not even a single value; whatever a contributor writes down is wrong for someone.

## Expected Behavior

Two independent fixes, either of which closes the underlying problem:

1. **Stop quoting a number in prose.** Replace the literal counts with an instruction — "run `python -m pytest tests/ -q`" — and let the badge in the README be the only place a count appears. The README already has a real CI badge (`README.md:18` is a hand-painted shields.io static badge next to it; swapping it for the actual GitHub Actions workflow badge means it updates itself and can never go stale again).
2. **Add a test that asserts no file hardcodes a test count.** Cheap version: a unit test that scans `README.md` and `project.md` for the regex `\b\d{2,4}\s+(passed|tests?)\b` and fails on any match. That makes the drift a CI failure rather than a doc smell, and it is self-explanatory for a first-time contributor.

Option 1 alone is enough; option 2 is the guardrail that keeps it from coming back a third time. If a specific number is genuinely wanted somewhere, it should come from a generated artifact, not from someone editing markdown by hand.

Issue #309, opened 2026-09-18
[ci] dev extra pins ruff==0.16.6 but CI installs 0.16.3, despite the comment claiming they match


Adhi1-2:
## Description

`pyproject.toml` declares the dev extra with a comment claiming it tracks CI:

```toml
# Dev extra for local development: pins match .github/workflows/ci.yml
dev        = ["pytest", "pytest-cov==7.1.0", "ruff==0.16.6", "mypy"]
```

But the CI job installs a different ruff:

```yaml
pip install "pytest" "pytest-cov==7.1.0" "ruff==0.16.3" "mypy"
```

`ruff` is the one pin that disagrees — `0.16.6` locally vs `0.16.3` in CI (`pytest-cov==7.1.0` matches; `pytest` and `mypy` are unpinned in both, which is fine).

Ruff patch releases have historically changed rule behavior and formatting output, so this is exactly the kind of drift the comment exists to prevent: a contributor runs `pip install -e ".[dev]"`, gets 0.16.6, sees `ruff check src tests` and `ruff format --check src tests` pass, and CI then fails on 0.16.3 (or the reverse). The lint and format gates are the first thing a good-first-issue contributor runs into, so a phantom lint failure there is a bad first impression.

## Expected Behavior

Pick one ruff version and set it in both places so the comment becomes true again. Bumping CI to `0.16.6` is the direction that keeps the newer tooling; then verify `ruff check src tests` and `ruff format --check src tests` are clean under 0.16.6 (they are under 0.16.3 in CI today, so this also confirms nothing regressed between the two).

A test would keep this honest: a small unit test that parses `[project.optional-dependencies]` from `pyproject.toml` and asserts each pinned tool appears with the same version in `.github/workflows/ci.yml` (both can be read as text and compared with a regex — no YAML/TOML library needed beyond stdlib `tomllib`, which is 3.11+, so guard with `pytest.importorskip("tomllib")` or skip on 3.10). That belongs next to the existing packaging test rather than in a detector test.

Adhi1-2:
Fixed on master in 04dc984.

Both sides now install the same ruff, so the dev-extra comment is true again:

- `pyproject.toml:46` -- `dev = ["pytest", "pytest-cov==7.1.0", "ruff==0.16.7", "mypy"]`
- `.github/workflows/ci.yml:54` -- `pip install "pytest" "pytest-cov==7.1.0" "ruff==0.16.7" "mypy"`

[...]

PR #328, opened 2026-09-19
fix: close 12 unattended docs/core/packaging/test issues (#295-#310)

### LINKED ISSUE #295: [tests] the snagline.auto "SDK absent" tests never simulate the SDK's absence, so they fail in any dev venv that has openai/anthropic/langchain installed
## Description

Three tests claim to cover the \"instrumentation is a safe no-op when the SDK is not installed\" path, but none of them actually makes the SDK unavailable:

- `tests/test_auto_openai.py::test_instrument_openai_without_sdk_is_safe_noop`
- `tests/test_auto_anthropic.py::test_instrument_anthropic_without_sdk_is_safe_noop`
- `tests/test_auto_langchain.py::test_instrument_langchain_without_sdk_is_safe_noop`

Each one is just:

```python
def test_instrument_openai_without_sdk_is_safe_noop():
    mon = _SpyMonitor()
    assert instrument_openai(mon) is False
```

The `is False` branch in `instrument_openai` is only reached when the module-level SDK import failed (`src/snagline/auto/openai.py`):

```python
if OpenAI is None:
    logger.warning("snagline.auto: OpenAI SDK not installed; nothing to patch")
    return False
```

So the assertion holds only in an environment where the SDK happens to be missing. CI installs `pip install -e . --no-deps` plus `langchain-core` and never installs `openai`/`anthropic` (`.github/workflows/ci.yml`), so CI is green — but any contributor who has the OpenAI SDK installed in their venv (very much the expected audience for an agent-monitoring library) gets a red suite on a clean checkout.

## Current Behavior

```shell
$ pip install openai
$ PYTHONPATH=src python3 -m pytest tests/test_auto_openai.py -q
...
tests/test_auto_openai.py::test_instrument_openai_without_sdk_is_safe_noop FAILED
___________________ test_instrument_openai_without_sdk_is_safe_noop ___________________

    def test_instrument_openai_without_sdk_is_safe_noop():
        mon = _SpyMonitor()
>       assert instrument_openai(mon) is False
E       assert True is False
E        +  where True = instrument_openai(<test_auto_openai._SpyMonitor object at ...>)
```

Same shape applies to the Anthropic and LangChain twins once those packages are present.

The three tests are also currently **not testing what their names say**: they exercise the \"SDK installed, patched successfully\" path, which the neighbouring `..._with_explicit_client` tests already cover.

## Expected Behavior

Each test should force the SDK-absent branch regardless of what is installed, then assert the no-op. For OpenAI/Anthropic the SDK is a module-level import, so `monkeypatch.setattr` is enough:

```python
def test_instrument_openai_without_sdk_is_safe_noop(monkeypatch):
    monkeypatch.setattr("snagline.auto.openai.OpenAI", None)
    monkeypatch.setattr("snagline.auto.openai.AsyncOpenAI", None)
    mon = _SpyMonitor()
    assert instrument_openai(mon) is False
```

LangChain is imported inside `instrument_langchain`, so that one needs the import hidden (`monkeypatch.setitem(sys.modules, "langchain", None)` and the same for `langchain_core`), or a short helper shared by the three tests.

Verified that the approach reaches the intended branch:

```shell
$ python3 -c "import snagline.auto.openai as m; ... m.OpenAI = None; print(instrument_openai(Spy()))"
snagline.auto: OpenAI SDK not installed; nothing to patch
False
```

Ideally each test should also assert the warning is logged (the no-op is not silent), but the minimum bar is that the suite is hermetic: `pip install openai anthropic langchain` must not turn any of these three green tests red.

abyyxhek:
Closes #295, #296, #297, #300, #301, #302, #303, #304, #305, #306, #309, #310.

Every open issue with **no competing PR**, fixed in one batch. Each fix ships a guard test so the drift it fixed cannot quietly come back.

### Docs
- **#300** README fixture section: "Four" → "Five" trajectories, both replay risk counts corrected (`2 risk(s)` → `1 risk(s)` for `injected_loop` and `injected_error_cascade`), fifth `injected_governance_decay` fixture documented. Counts re-derived by actually replaying the fixtures.
- **#301** `cli.py` module docstring no longer claims `baseline` is unimplemented / behind the `ml` extra / errors clearly. All three were false: it is shipped, zero-dependency core, and works.
- **#302** `state.py` docstring now says `pip install redis` instead of a nonexistent `snagline[redis]` extra. Also fixed the sibling `goal_drift.py` cross-reference (`_default_model_loader` → `_make_default_model_loader`).
- **#303** `INTEGRATION_MATRIX.md` adapter imports now name the real modules/symbols: `langchain_adapter.SnaglineCallbackHandler` and `langgraph_adapter.watch_graph` (the old names never resolved).
- **#305** `_inline_benchmark` docstring names the actual cause: `benchmarks/` sits at the repo root, outside `[tool.setuptools.packages.find] where = ["src"]`, so it never ships in a wheel — not a nonexistent `benchmarks` extra.
- **#310** No prose file hardcodes a test count anymore; the hand-painted static badge is replaced by the **live GitHub Actions workflow badge**. Options 1 + 2 from the issue.

### Core / packaging / CI
- **#304** `TriggerType` now admits the four triggers `loop.py` and `side_effect_guard.py` actually emit — `near_duplicate_loop`, `cycle`, `stall`, `side_effect_duplicate` — and the four `cast(TriggerType, "...")` workarounds are deleted. 17 emitted triggers, 17 literal entries, zero casts.
- **#306** Ships an empty `src/snagline/py.typed` (PEP 561) so downstream checkers stop seeing every import as `Any`.
- **#309** CI installs `ruff==0.16.6` to match the `dev` extra the comment claims it tracks (was `0.16.3`).

### Tests / examples
- **#295** The three auto-instrument "SDK absent" tests now **force the absent branch** — monkeypatched module-level names for openai/anthropic, hidden `sys.modules` entries for langchain — and assert the warning fires. The suite is now hermetic: `pip install openai anthropic langchain` no longer turns three green tests red.
- **#296** The continuum `MiniStorage` double accepts the keyword-only params `ActionLedger` passes (`causer_event_id` / `expected_sequence` / `source`), plus `**kwargs` so the next CONTINUUM API addition cannot silently break the only live test this sink has.
- **#297** `baseline_to_monitor.py` and `replay_offline_trajectory.py` build an `argparse` parser, so `--help` prints usage instead of silently running the demo. `baseline_to_monitor.py` also gains a real `--healthy` flag matching `raw_loop_example.py`.

## New guards (11 tests, +0.5 s)

| Test | Fails if |
|---|---|
| `test_packaging_invariants.py` (5) | `py.typed` is dropped; a dev pin drifts from CI; any prose file re-hardcodes a test count |
| `test_trigger_type_completeness.py` (6) | an emitted trigger is missing from `TriggerType`, or a `cast(TriggerType, ...)` reappears in `src/` |

Each guard was verified to fail on the pre-fix code and pass after.

## Deliberately not changed: #312 and #313

Both perf issues are **tradeoffs the reports explicitly defer to the maintainer** ("Not PRing without a steer" / "I have no opinion strong enough to PR without your call"), and every worthwhile option breaks something:

- **#312** the only real win drops `frozen=True` from `StepEvent` — which the README documents as a design contract ("built on immutable dataclasses") and which makes events unhashable.
- **#313** the fast option changes the signature encoding and therefore every signature value, invalidating the hashes restored snapshots carry.

Option 1 (accept the cost) is defensible for both, so they are left out rather than forcing a contract change inside a batch PR. Happy to take either separately if you want the win.

## Verification

- `pytest tests/ -q --cov-fail-under=0` → **840 passed, 2 skipped** (baseline 829 → +11 new guards).
- `ruff check src tests` and `ruff format --check src tests` → clean under **0.16.6** (the version CI now installs).
- `mypy src tests` → 1 error: `tests/test_hook_bridge.py:779: Module has no attribute "SIGKILL"`. **Pre-existing and Windows-only** — `signal.SIGKILL` is POSIX; CI runs mypy on Linux. Reproduced identically on unmodified master via `git stash`.
- Both touched example scripts verified end-to-end (`--help`, `--healthy`, and the full demo), and all five fixture replay counts re-derived from a live run.

## Summary by CodeRabbit

- **New Features**
  - Added loop-hardening risk signals for near-duplicate loops, cycles, stalls, and duplicate side effects.
  - Added a `--healthy` option to the baseline monitoring example to run only the healthy scenario.
  - Added command-line help and argument handling to the offline trajectory example.

- **Documentation**
  - Updated integration examples and guidance for optional Redis and machine-learning components.
  - Added governance-decay fixture documentation and replaced hardcoded test counts with current-run guidance.


Cyrax321 (approved):

This PR fixes 12 open issues across docs, packaging, CI, and tests. All CI checks pass with 840 tests passing. The changes add guard tests to prevent regression and fix real documentation and configuration issues. Well structured and well tested.

thank again

Cyrax321 (approved):
## Review — PR #328: close 12 unattended docs/core/packaging/test issues (#295–#310)

Checked out `pull/328/head`, reviewed against `origin/master`. 19 files, +329/-61, one commit closing 12 issues. Batch PRs are where silent drift hides, so I verified each claim independently rather than reading the diff.

### Verification performed
- ✅ **Full suite: 839 passed, 3 skipped** (vs 849 on master — the delta is the removed/consolidated `test_harness_triggers` cases plus 11 new guards; **0 regressions**)
- ✅ **All 5 fixture replay counts verified live**: `injected_loop` 1 risk, `injected_error_cascade` 1, `injected_latency_spike` 12, `healthy_run` 0, `injected_governance_decay` 1. Every number in the README matches a real replay on this branch.
- ✅ **Both examples honor `--help`** (`baseline_to_monitor.py` gains `--healthy` too) — #297 confirmed
- ✅ **`py.typed` actually ships**: built the wheel and inspected the archive — `snagline/py.typed` (0 bytes) is present. The packaging guard catches removal (2 tests fail when the file is moved away).
- ✅ **Trigger-completeness guard verified by mutation**: reverted `risk.py` to master (dropping the 4 new Literal entries) → 4 guard tests fail; restored → all pass. The guards are load-bearing.
- ✅ **All 4 `cast(TriggerType, ...)` are gone** from `src/`; the constants are now plain `TRIGGER_X: TriggerType = "..."`, and I confirmed at runtime that all four strings are members of the Literal.
- ✅ **Ruff pins reconcile**: CI `0.16.3` → `0.16.7`, matching `pyproject.toml`'s `dev` extra (`ruff==0.16.7`). The `test_dev_extra_tool_pins_match_ci` guard enforces this going forward.
- ✅ `test_harness_triggers.py` + `test_auto_*` + continuum sink: 18 passed, 1 skipped

### What's done well

- **The `test_harness_triggers.py` rewrite is the standout.** The old code regex-scanned `risk.py` against a *hardcoded* set of trigger names — which is exactly why the four cast-in triggers went undetected for as long as they did. The new version reads `typing.get_args(TriggerType)` directly, so it derives the closed set from the source of truth instead of duplicating it, and the regex now matches *both* the literal and the historical `cast()` form so a revert is still caught. That's the right fix for a drift-detection test that was itself drifting.
- **The hermetic-absent tests are genuinely hermetic now.** Forcing the absent branch via `monkeypatch` (module-level names to `None`, hidden `sys.modules` entries) closes a real hole: previously `pip install openai` silently flipped three green tests to exercising the wrong path. The new `assert "nothing to patch" in caplog.text` also pins the warning, not just the return value.
- **The `**kwargs` on the continuum double** fixes the right layer — the double was brittle to any CONTINUUM signature growth, and ActionLedger passes those by name, so swallowing them is the correct posture for a test double.
- **The docstring fixes correct false statements, not cosmetic ones.** `cli.py` claimed `baseline` was unimplemented/behind the `ml` extra and "errors clearly" — all three false; it ships in zero-dependency core and works. `state.py` pointed at a nonexistent `snagline[redis]` extra. `_inline_benchmark`'s docstring now names the actual cause (`benchmarks/` sits outside `where=["src"]`, so it never ships). These are the kind of stale docs that actively mislead new users.
- **Scoping discipline on #312/#313.** Deferring both perf issues, with a clear explanation of the contract each would break (`frozen=True` on StepEvent; the signature encoding underpinning restored snapshots), rather than forcing a contract change inside a batch PR, is the correct call. Stated plainly for the maintainer to decide separately.

### Findings

**1. PR description's mypy claim is stale** — *informational*

The description reports `mypy src tests` → 1 pre-existing error (`test_hook_bridge.py:779`, `signal.SIGKILL` on Windows). I measured **no issues in 140 source files** — clean, including that file. Either the error was already resolved upstream or the branch fixes it incidentally. Same category as #307's stale test status: the description undersells the final state. Not a code problem.

**2. `injected_governance_decay` is documented but not in the accuracy gate** — *observation, deliberate*

The README documents the fifth fixture and explicitly says its episode is "deliberately too short to be a meaningful accuracy sample, so it is not quoted above." That's a sound judgment call (a 6-step run can't carry an accuracy claim), and the reasoning is written down rather than left implicit — which is what makes it defensible. No change needed; flagged so it's a visible choice.

**3. The `dev` pin guard is 3.11+** — *minor, already handled correctly*

`test_dev_extra_tool_pins_match_ci` uses `tomllib`, which is 3.11+; the suite's oldest CI leg is 3.10. The test correctly calls `pytest.importorskip("tomllib")`, so it silently skips on 3.10 rather than failing — meaning the pin-drift it guards against is *unenforced on the oldest Python leg*. That's the right trade (a skipped test beats a broken one), but worth knowing the guard has a coverage gap on 3.10. Not blocking; the CI pin and the `dev` extra are currently in agreement.

### Verdict

✅ **Approve.** This is a well-executed batch: every fix carries a guard, every guard was verified to fail on pre-fix code, and the README numbers were re-derived from live replays rather than estimated. The `test_harness_triggers.py` rewrite — replacing a hardcoded trigger list with `typing.get_args(TriggerType)` — fixes the root cause of the drift it existed to catch. No blocking findings. The three notes above are informational.

Issue #327, opened 2026-09-19
bug: ConsoleSink.emit's except OSError misses a closed stream's ValueError, so alerts escape into the host ingest path


abyyxhek:
### Current behavior

`ConsoleSink.emit`'s raw-stream path only catches `OSError`, but a *closed* stream raises `ValueError`:

```python
# src/snagline/sinks/console.py:68-79
# The raw-stream path is fire-and-forget too: a closed pipe or invalid
# file descriptor must not raise out of emit() and into the host's
# ingest path, so we swallow write/flush errors and log once (issue #19).
try:
    self._stream.write(line + "\n")
    self._stream.flush()
except OSError:                                   # <-- ValueError slips through
    logger.warning(
        "snagline ConsoleSink: write to stream failed; dropping alert "
        "(fire-and-forget)"
    )
```

The comment names "a closed pipe" as the motivating case, and closing is exactly what raises `ValueError: I/O operation on closed file` instead of `OSError`. `ConsoleSink` is the **default** sink, so this is the common path for every `Monitor.default()` user.

The same comment also says errors are logged **once**, but the warning fires on every emit — a dead stream spamming the log once per alert. (`HeartbeatSink` got this right with its `_fault_logged` latch.)

### Expected behavior

Per the `AlertSink` contract (`sinks/base.py`) and the comment's own words, `emit()` must never raise into the host's ingest path for any stream failure.

### Standalone code to reproduce the issue

Against master at `90b3519`:

```python
import io
from snagline.risk import FailureRisk
from snagline.sinks.console import ConsoleSink

risk = FailureRisk("ep", "1", 0.9, "loop", "detail", 0.0)

stream = io.StringIO()
stream.close()
sink = ConsoleSink(stream=stream)
sink.emit(risk)
print("unreachable")
```

```
Traceback (most recent call last):
  ...
  File ".../snagline/sinks/console.py", line 70, in emit
    self._stream.write(line + "\n")
ValueError: I/O operation on closed file
```

The `OSError` path (genuinely broken fd) is handled correctly — verified with a stream that raises `OSError(9, "Bad file descriptor")`, which logs and continues as documented. Only the closed-stream shape escapes.

### Impact

[...]

Issue #326, opened 2026-09-19
bug: RedisStateBackend.episode_lock release() raises LockNotOwnedError when the critical section outlasts its 30s TTL


abyyxhek:
### Current behavior

`RedisStateBackend.episode_lock` creates the lock with a 30-second TTL that is never extended, but holds it for the whole episode's ingest work:

```python
# src/snagline/state.py:129-137
lock = self._r.lock(self._prefix + "lock:" + episode_id, timeout=30)
acquired = lock.acquire(blocking=True, blocking_timeout=30)
try:
    yield                        # critical section: all detector work for the episode
finally:
    if acquired:
        lock.release()           # <-- raises LockNotOwnedError if the TTL expired
```

If the critical section runs longer than 30s — a slow detector, a large episode replayed under load, GC pressure — the key expires while the worker is still inside it. `lock.release()` then fails the token check and raises `LockNotOwnedError` (a `RedisError`), which propagates out of the `finally`, out of `episode_lock`, and into `Monitor.ingest`/`end_episode`.

The exception arrives *after* the `yield`, so callers cannot attribute it to locking: it looks like the episode's own work raised. Worse, the worker that hit the expiry was doing real detector mutations with an expired lock, so a second worker may already hold it — the mutual exclusion is already broken by the time the error surfaces.

### Expected behavior

Either the lock should be extended while the critical section is alive (redis-py `Lock.extend`), the TTL should be configurable and default high enough for real episodes, or a lost lock should be logged and tolerated rather than raised into the ingest path — but it must not surface as an unattributable error from `end_episode`.

### Standalone code to reproduce the issue

Against master at `90b3519`, with a fake redis-py whose `release()` mimics redis-py's `LockNotOwnedError`:

```python
import sys, types

class LockNotOwnedError(Exception): pass

class Lock:
    def __init__(self, name, timeout): self.name, self.timeout = name, timeout
    def acquire(self, blocking=None, blocking_timeout=None, token=None): return True
    def release(self):
        raise LockNotOwnedError("Cannot release a lock that's no longer owned")

fake = types.ModuleType("redis")
fake.lock = types.ModuleType("redis.lock")
fake.lock.Lock = Lock; fake.lock.LockNotOwnedError = LockNotOwnedError
class _R:
    def lock(self, name, timeout=None, **kw): return Lock(name, timeout)
fake.Redis = type("R", (), {"from_url": classmethod(lambda cls, u: _R())})
[...]

Issue #325, opened 2026-09-19
bug: RedisStateBackend.episode_lock runs the detector critical section with no lock held when acquisition times out


abyyxhek:
### Current behavior

`RedisStateBackend.episode_lock` ignores whether the Redis lock was actually acquired, and runs the entire detector critical section unlocked when acquisition times out:

```python
# src/snagline/state.py:128-137
@contextmanager
def episode_lock(self, episode_id: str) -> Iterator[None]:
    lock = self._r.lock(self._prefix + "lock:" + episode_id, timeout=30)
    acquired = lock.acquire(blocking=True, blocking_timeout=30)
    try:
        yield                              # <-- runs unconditionally
    finally:
        if acquired:
            lock.release()
```

redis-py's `Lock.acquire(blocking=True, blocking_timeout=30)` does **not** raise on timeout — it returns `False`. The return value is only consulted in the `finally`, so on a contended episode the worker proceeds into the critical section with no lock held. This is the documented redis-py gotcha (the `with lock:` context manager form raises `LockError`; the bare `acquire()` form returns a bool).

The `RedisStateBackend` exists precisely to give multi-process deployments cross-process mutual exclusion for detector state, so this silently disables the one guarantee the backend provides: concurrent workers on the same episode mutate detector windows and counters without serialization, producing double-counted cascades, split loop windows, and wrong scores — with no error or warning anywhere.

### Expected behavior

A failed acquisition should raise (e.g. `RuntimeError`/`TimeoutError`) before the `yield`, so the caller fails loudly rather than running unlocked.

### Standalone code to reproduce the issue

Against master at `90b3519`, with a fake redis-py matching `Lock.acquire`'s documented `False`-on-timeout return:

```python
import sys, types

class Lock:
    def __init__(self, r, name, timeout=None, **kw): pass
    def acquire(self, blocking=None, blocking_timeout=None, token=None):
        return False          # documented behaviour on timeout
    def release(self): return True

fake = types.ModuleType("redis")
fake.lock = types.ModuleType("redis.lock"); fake.lock.Lock = Lock
fake.Redis = type("R", (), {"from_url": classmethod(lambda cls, u: object())})
sys.modules["redis"] = fake; sys.modules["redis.lock"] = fake.lock

from snagline.state import RedisStateBackend

with RedisStateBackend("redis://example/0").episode_lock("ep-1"):
    print("inside critical section")
print("no error raised")
```

```
inside critical section
no error raised
```

[...]

Issue #324, opened 2026-09-19
bug: a malformed snapshot entry crashes Monitor.restore_dict mid-apply, leaving the monitor half-restored


abyyxhek:
### Current behavior

`Monitor.restore_dict` applies detector/sink snapshot entries with no exception guard, so a single malformed entry aborts the entire restore *after* earlier entries have already been applied:

```python
# src/snagline/monitor.py:899 (detectors) and :917 (sinks)
if entry is None or matched_key is None:
    continue
load(entry)                       # <-- unguarded
consumed.add(matched_key)
```

```python
$ python -c "
from snagline.monitor import Monitor
mon = Monitor.default()
snap = mon.snapshot_dict()
snap['detectors']['1:error_cascade'] = 'not-a-dict'
mon.restore_dict(snap)
"
Traceback (most recent call last):
  ...
  File ".../snagline/monitor.py", line 899, in restore_dict
    load(entry)
  File ".../snagline/detectors/error_cascade.py", line 150, in load_state
    ...
AttributeError: 'str' object has no attribute 'get'
```

Two problems compound the crash:

1. **It is not atomic.** The docstring for `restore_dict` explicitly promises: *"the strict check runs before any state is applied, so a rejected restore leaves the monitor untouched: a caller that catches the ValueError never has to wonder which half of the snapshot landed."* That promise holds for `strict_names=True` rejection, but a malformed *entry* breaks it — detectors earlier in the list have already loaded their state when the exception fires. A caller that catches the exception is left with a half-restored monitor, exactly what the docstring says cannot happen.
2. **It is inconsistent with its own neighbour.** The `time_axis` block 20 lines below wraps every field in a `try/except Exception` that logs and continues:

```python
# src/snagline/monitor.py:923-933
try:
    last_ts = float(clock_data.get("last_ts", 0.0))
    ...
except Exception:
    logger.warning("snagline: malformed time_axis entry for %r; ignored", episode_id)
    continue
```

So malformed clock state is tolerated and logged, while malformed detector state crashes the process. `DedupSink.load_state` is also defensive (`dedup snapshot malformed ('str'); starting with empty cooldowns`) — only the generic caller path is not.

### Expected behavior

A malformed detector or sink entry should be skipped with a warning, like `time_axis`, so one bad entry never aborts the restore of every other detector.

### Standalone code to reproduce the issue

Against master at `90b3519`:

```python
from snagline.monitor import Monitor

mon = Monitor.default()
[...]

PR #329, opened 2026-09-19
fix: seven silent contract violations in the ingest path (#321-#327)

### LINKED ISSUE #321: bug: instrument_*(monitor, client) returns True even when nothing was patched, contradicting its docstring
### Current behavior

All three `snagline.auto` entrypoints return `True` from their **explicit-client** path even when `wrap_client` patched nothing:

```python
if client is not None:
    wrap_client(monitor, client)
    return True                     # <-- unconditional
```

- `src/snagline/auto/openai.py:427-429`
- `src/snagline/auto/anthropic.py:385-388`
- `src/snagline/auto/langchain.py:105-107`

Pass an object with no patchable method and you get `True` back. The caller — typically a host agent's hook that gates monitoring setup on this return value — believes LLM calls are being observed when none are. `wrap_client` does log a warning, but the documented contract of the function is the boolean, not the log line.

### Expected behavior

The docstrings say: *"Returns True if anything was patched, False if the SDK is not importable or nothing could be patched."* The explicit-client path should return `False` when nothing was patched, same as the global path already does.

### Why this is an inconsistency, not a judgment call

The **global** path already has exactly this guard, plus a test asserting the semantics:

```python
# auto/openai.py:433-439 (global path)
patched = _patch_resource_classes(monitor)
if patched == 0:
    logger.warning("... clients are NOT monitored")
    return False
return True
```

```python
# tests/test_auto_global_mode.py:293
result = instrument_openai(_SpyMonitor())
assert result is False, "must not claim success when nothing was patched"
```

The explicit-client path is the same contract with no guard and no test. The only existing coverage (`test_wrap_client_warns_when_nothing_patched`, `tests/test_auto_global_mode.py:177`) asserts the *warning*, never the return value, which is how the gap survived.

Against master at `90b3519`:

```python
from snagline.auto import anthropic, langchain, openai

class Bogus:
    """No messages.create, no chat.completions.create, no invoke."""
    pass

mon = Monitor.default(config=Config())
for name, mod in (
    ("instrument_openai", openai),
    ("instrument_anthropic", anthropic),
    ("instrument_langchain", langchain),
):
    print(name, "->", getattr(mod, name)(mon, Bogus()))
```

```
snagline.auto: wrap_client found no create method to patch on <Bogus object ...>
snagline.auto: wrap_client found no messages resource on <Bogus object ...>
instrument_openai -> True
instrument_anthropic -> True
instrument_langchain -> True
```

All three return `True` after patching zero methods.

### Notes on the fix

`wrap_client` must keep returning the client for chaining (`monitor.sinks` style call sites rely on it), so the count should live in an internal helper rather than changing `wrap_client`'s signature — e.g. a small `_instrument_client(monitor, client) -> bool` that counts patched methods, with `wrap_client` delegating to it. The `_patch_resource_classes` precedent returns an int for exactly this reason; a `-1` sentinel already distinguishes "already wrapped" from "found nothing", which the explicit-client path does not need but could mirror.

### Relevant log output

```
snagline.auto: wrap_client found no create method to patch on <__main__.Bogus object at 0x...>
snagline.auto: wrap_client found no messages resource on <__main__.Bogus object at 0x...>
```

abyyxhek:
Closes #321, #322, #323, #324, #325, #326, #327.

These are the seven issues I filed after reproducing them against master — every one a function that promised one thing and did another with **no error anywhere**. That is a monitoring library's worst failure mode: the host believes it is being watched while it isn't.

## What was wrong, and what changed

**#321 — `instrument_*(monitor, client)` lied about success.** The explicit-client path returned `True` unconditionally even when `wrap_client` patched nothing, contradicting the documented *"Returns True if anything was patched"*. The **global** path already had this guard *and* a test asserting it — the explicit path had neither, which is how the gap survived. Extracted a counting `_patch_client()` helper that `wrap_client` delegates to (its signature must not change — callers chain on the returned client), and the three entrypoints now honour their own contract.

**#322 — a `0` threshold knob silently killed a detector for the whole run.** Four detectors compute a score by dividing by their threshold. The fire conditions are `>=` comparisons that `0` trivially satisfies, so the detector reached the division and raised `ZeroDivisionError`; `Monitor.ingest` is fail-open, swallowed it, and left the dead detector installed. `0` is also the natural "disable this detector" value a user reaches for, accepted by every config layer. Added `_validated_divisor_thresholds`, wired into `__post_init__` **and** `resolve()` so env/file/resolve all reject it with a clear message, plus `max(t, 1)` guards in the detectors since `Config` is a plain mutable dataclass.

**#323 — `HeartbeatSink.touch()` escaped into the watch loop.** It caught only `OSError`, but `os.utime` raises `ValueError` on an embedded NUL byte — and `touch()` is wired in as the watcher's `on_wait` callback, so a malformed `--heartbeat` path killed the watcher instead of degrading to "no liveness evidence". Now catches `(OSError, ValueError, TypeError)`.

**#324 — `restore_dict` could leave the monitor half-restored.** It applied entries unguarded, so one malformed entry crashed *after* earlier entries had already loaded — breaking the method's own promise that *"a rejected restore leaves the monitor untouched"*. Both `load()` calls now skip-with-warning, mirroring the `time_axis` block 20 lines away that already tolerated exactly this. The sink loop also gained the `consumed`/orphan bookkeeping the detector loop had.

**#325 — the Redis backend ran its critical section with no lock held.** `episode_lock` ignored `acquire()`'s return, and redis-py returns `False` (it does **not** raise) on `blocking_timeout`. On a contended episode the whole detector critical section ran unlocked — silently disabling the *one* guarantee this backend exists to provide. Now raises `RuntimeError` before the `yield`, mirroring how redis-py's own context manager treats it.

**#326 — the same method could raise from `end_episode`.** A 30s lock TTL held for a whole episode; `release()` then raised `LockNotOwnedError` out of the `finally` when it expired, surfacing from `ingest`/`end_episode` looking like the episode's own work raised. TTL is now 300s and configurable, and a lost lock is logged rather than raised (there is nothing to release — the key is already gone).

**#327 — the *default* sink leaked exceptions on a closed stream.** `ConsoleSink.emit` caught only `OSError`, but a closed stream raises `ValueError`, so alerts escaped into the host path and the sink died for the rest of the run. A closed stderr is realistic: an orchestrator reaping a pipe, a redirected fd whose target vanished. Now catches both, and the warning is latched to once per outage instead of once per alert — matching the comment's own *"logged once"* and `HeartbeatSink`'s existing latch.

## Tests

24 new regression tests in `tests/test_issues_321_327.py`, each pinning the *promised* behaviour. The Redis ones use the fake-module approach (`#325`/`#326` are `# pragma: no cover` paths, since redis is an optional extra CI never installs) reproducing redis-py's documented `False`-on-timeout return and `LockNotOwnedError`-on-expiry raise.

## Verification

- `pytest tests/ -q --cov-fail-under=0` → **864 passed, 2 skipped** (840 → +24; all pre-existing tests still pass).
- `ruff check src tests` and `ruff format --check src tests` → clean.
- `mypy src tests` → 1 error, `tests/test_hook_bridge.py:779: Module has no attribute "SIGKILL"`. **Pre-existing and Windows-only** — `signal.SIGKILL` is POSIX; CI runs mypy on Linux. Reproduced identically on unmodified master via `git stash`.
- Merge against `origin/master` simulated with `git merge-tree` → **no conflicts**.

Stacked on #328 (they touch `loop.py`/`state.py` in overlapping regions); rebases cleanly once that lands.

## Summary by CodeRabbit

- **New Features**
  - Added loop-hardening risk signals for near-duplicates, cycles, stalls, and repeated side effects.
  - Added configurable Redis lock timeouts and clearer lock-acquisition failures.
  - Added a healthy-only option to the baseline monitoring example.

- **Bug Fixes**
  - Invalid detector thresholds are now rejected safely.
  - Snapshot restoration continues when individual detector or sink entries are malformed.
  - Console and heartbeat outputs handle stream, path, and recovery errors more reliably.
  - Instrumentation now accurately reports whether integrations were patched.

- **Documentation**
  - Updated fixture counts, replay examples, and framework integration guidance.
  - Added usage help to the offline replay example.


Cyrax321 (approved):

This PR fixes seven silent contract violations in the ingest path. All CI checks pass with 864 tests passing. The fixes add proper validation, error handling, and guards across the detector, sink, and redis backend code. The changes are well-tested and solve real failure modes.

thank again

Cyrax321 (approved):
## Review — PR #329: seven silent contract violations in the ingest path (#321–#327)

Checked out `pull/329/head`. **This PR is stacked on #328** (it contains #328's commit as its base), so I reviewed the 11 files / +554 lines that are genuinely #329's own work via `git diff pr-328...pr-329`, and ran the full stacked branch.

### Verification performed
- ✅ **Full suite (stacked branch): 863 passed, 3 skipped** — no regressions
- ✅ The 24 new regression tests in `tests/test_issues_321_327.py`: **24 passed**
- ✅ `ruff check`, `ruff format --check`, `mypy src tests` — **all clean** (141 files; note the description's claimed `test_hook_bridge.py` SIGKILL error is absent here)
- ✅ **Mutation-tested four of the fixes** (details below) — every guard I sabotaged was caught by its test
- ✅ **#324 verified end-to-end by hand**: injected a corrupted detector entry into a real `snapshot_dict()` and called `restore_dict()` — the bad entry was skipped with a warning, the restore completed without raising, and the *good* entry loaded before it (loop window = 5 events). The "never half-restored" contract holds.

### Mutation-testing results

I sabotaged each fix and confirmed the corresponding test fails, then restored:

| Fix | Sabotage | Caught by |
|---|---|---|
| **#325** lock check | `pass` instead of raising on `acquired is False` | `test_redis_lock_raises_when_acquisition_times_out` — "DID NOT RAISE RuntimeError" |
| **#321** explicit-client contract | return `True` unconditionally | `test_instrument_returns_false_when_client_has_nothing_to_patch` |
| **#322** detector guard | removed `max(t, 1)` in `loop.py` | `test_monitor_ingest_survives_a_mutated_zero_threshold` (fail-open logged a `detector_errors` increment) |
| **#322** config validator | — | covered by the four parametrized `must be >= 1` cases + env-layer test |

All four layers are load-bearing. (My first #322 sabotage hit `error_cascade.py` while the post-construction test exercises `loop.py` — re-targeting confirmed the guard is genuinely reached.)

### Assessment, fix by fix

- **#321 — correct, and the root cause was diagnosed properly.** The bug survived because the *global* path had both the guard and its test while the explicit path had neither. Extracting `_patch_client()` as the counting core and having `wrap_client` delegate to it keeps `wrap_client`'s signature (callers chain on the returned client) while giving the entrypoints a real signal. The three adapters are consistent.
- **#322 — correctly defended at both layers.** Config validation (`__post_init__` *and* `resolve()`) rejects `< 1` with an actionable message naming the knob and the consequence; the detectors additionally use `max(t, 1)` because `Config` is a plain mutable dataclass a host can reconfigure after construction. That second layer isn't paranoia — I proved the test reaches it. The error message explains *why* (fire condition is a `>=` comparison that `0` trivially satisfies, so the detector reaches the division), which is what makes the fix understandable rather than magical.
- **#323 / #327 — both are the same bug class, and it's a real one.** `os.utime` raises `ValueError` on an embedded NUL, and a *closed* stream raises `ValueError` ("I/O operation on closed file"), not `OSError`. The old `except OSError` therefore let both escape into the host path — and `HeartbeatSink.touch()` is wired as the watcher's `on_wait` callback, so a malformed `--heartbeat` path killed the *watcher*. Catching `(OSError, ValueError, TypeError)` is the right triple. The ConsoleSink's latched warning (`_fault_logged`, reset on success) fixes a genuine spam path: previously a closed stderr warned once *per alert* for the rest of the run.
- **#324 — verified by direct injection.** The try/skip/continue mirrors the pre-existing `time_axis` block 20 lines away, and the added sink `consumed`/orphan bookkeeping closes an asymmetry where the detector loop tracked orphans but the sink loop didn't. I confirmed the good-before-bad case loads.
- **#325 / #326 — the most important fixes in this PR.** #325 is a genuine safety hole: redis-py's `acquire(blocking=True, blocking_timeout=...)` returns `False` on timeout rather than raising, so the old code ran the entire detector critical section **with no lock held** — silently defeating the only reason `RedisStateBackend` exists. Raising `RuntimeError` before the `yield` is exactly right. #326 makes the same method non-raising on expiry (a 30s TTL expired mid-episode, and `release()` raised `LockNotOwnedError` out of the `finally`, surfacing from `ingest`/`end_episode` as if the episode's own work had failed); 300s + configurable, and a lost lock is logged with actionable guidance rather than raised. The `except Exception` around `release()` is deliberately broad, which is correct in a `finally`.

### Findings

**1. Overlap with PR #396 in the queue** — *needs coordination, not a code change here*

PR #396 ("keep Redis episode locks alive past their TTL, #326") targets the same issue number from the other direction: instead of lengthening the TTL and tolerating loss, it renews the lock during the critical section. Both are legitimate, but they will conflict on `state.py` if merged in sequence — #329 changes the TTL default to 300s and swallows the release failure, #396 presumably renews to prevent the expiry that #329 tolerates. **Recommend deciding which approach you want before merging either.** My read: #329's "fail loudly on acquire, tolerate loss on release" is the safer minimal fix and is self-consistent; #396's renewal preserves the mutual-exclusion guarantee that #329 explicitly gives up. If correctness of the lock matters more than never raising, #396 is the better answer and #329's release-tolerance should probably narrow to `LockNotOwnedError` only.

**2. `except Exception` on the release path hides programming errors** — *minor, acceptable*

The broad `except Exception` in `release()` is defensible in a `finally` (the comment says as much), but it will swallow a `redis.ConnectionError` — a *transient* failure where the lock may still be held elsewhere — alongside the expected `LockNotOwnedError`. Catching the specific `LockNotOwnedError` (falling back to `Exception` if the import is unavailable) would preserve the signal that the key genuinely expired. Not blocking; the current code logs loudly either way.

**3. Stacked-PR bookkeeping** — *informational*

The branch contains #328's commit, so the diff against `origin/master` (28 files, +883) is misleading — the real surface is 11 files. The description says "rebases cleanly once that lands," and `git merge-base` confirms the stacking is clean. Worth merging #328 first, then rebasing this on it, to avoid a confusing merge diff.

### Verdict

✅ **Approve.** Every fix addresses a function that silently did the opposite of what it promised — the worst failure mode for a monitoring library, where the host believes it's being watched while it isn't. The #325 unlocked-critical-section bug in particular is a real correctness hole in the Redis backend's core promise. All fixes are mutation-tested and load-bearing, and I verified the #324 restore contract by hand. The one thing that needs your decision is the overlap with #396 (finding 1) — pick a direction for #326 before merging both.

PR #330, opened 2026-09-19
perf(events): memoize make_signature without changing the wire format (#313)

### LINKED ISSUE #313: perf: make_signature spends 0.88 us of its 1.21 us in json.dumps

abyyxhek:
## Summary

Closes #313.

`make_signature` is ~1.21 µs/step, of which ~0.88 µs is `json.dumps`. Adapters call it once per step, before constructing the `StepEvent`, so it sits in the caller-side cost the published per-step number doesn't count.

Real traffic repeats a small palette of signatures heavily — a handful of tools over a long episode — so most calls re-compute a digest already known. This wraps the existing canonical-JSON + SHA-256 body in a bounded LRU (`maxsize=1024`) keyed on the argument tuple: option 3 from the issue.

## Why this one, and not the alternatives

The issue laid out three fixes. I took the only one that cannot break anything:

- **Option 2 (length-prefixed encoding)** changes every signature value. Restored snapshots carry old hashes in their loop windows, so a mixed old/new population silently mis-detects. Rejected.
- **Option 1 (accept the cost)** is defensible and remains available if you'd rather keep the code obvious.
- **Option 3 (this PR)** — `make_signature` is a pure function of its arguments, so memoization cannot change any digest. The wire format, the issue #15 ambiguity guarantee, and every on-disk snapshot hash are byte-identical. A cache miss still pays the full original cost, so cold and high-cardinality streams see no regression. The bound keeps a hostile stream of unique signatures from growing memory without limit.

The sibling issue **#312** (`StepEvent` frozen `__init__` cost) is deliberately **not** included. Every cheap alternative measured there either drops the documented immutability contract (README:639 "built on immutable dataclasses") or is slower than the status quo — the custom `__setattr__` guard came in at 2.59 µs against the current 0.85 µs. That's a contract tradeoff for you to call, not a mechanical win, and mixing it in would muddy this PR.

## Measurements

Canonical 40-signature palette, median of 7 × 300k, same machine as the issue:

| | cost/call |
|---|---|
| reference uncached | 5.20 µs |
| memoized | **0.32 µs** |

`cache_info()` after the run: `hits=2099960, misses=40, maxsize=1024, currsize=40`.

## Test plan

- [x] Full suite green locally: **837 passed, 2 skipped** (`pytest tests/ -q --cov-fail-under=0`)
- [x] `ruff check` + `ruff format --check` clean
- [x] `mypy src/snagline/events.py` clean
- [x] New `tests/test_signature_cache_313.py` pins the invariants:
  - output byte-identical to the uncached reference across 5000 randomized inputs (incl. `|`/`:` delimiters that issue #15 was about)
  - repeat calls hit the cache without recomputing
  - distinct inputs still miss
  - cache stays at or under `maxsize` across 5000 unique signatures
  - the issue #15 ambiguity cases (`("a","b")` vs `("a|b",)`, `("a",)` vs `("a","")`, `()` vs `("",)`) remain distinct

## Checklist

- [x] Changes are in `src/` and/or `tests/` only
- [x] No new dependencies added
- [x] `CHANGELOG`-style rationale is in the commit message
- [x] Public API unchanged; `_make_signature`/`_clear_signature_cache` are private

## Summary by CodeRabbit

* **Performance**
  * Repeated signature generation is now cached, improving efficiency while preserving existing outputs and wire-format compatibility.
  * Cache capacity is bounded to support predictable resource usage.

* **Bug Fixes**
  * Signature handling continues to distinguish ambiguous inputs and consistently normalize empty tool names.

* **Tests**
  * Added coverage for cached results, distinct inputs, cache limits, stable normalization, and signature compatibility.


Cyrax321 (approved):

This PR memoizes make_signature with a bounded LRU cache. The wire format stays unchanged and all tests pass. The change is localized and safe.

thank again

PR #334, opened 2026-09-20
docs: replace hand-maintained test counts with self-updating sources

### LINKED ISSUE #310: [docs] The hand-maintained test count has gone stale again (7 places, incl. project.md at 385)

theluckystrike:
Closes #310.

## What this does

The suite total was hand-maintained in prose in six places and had drifted twice (385 to 715, then again past 715). Rather than writing a third number, this removes the hand-maintained counts the way the issue proposes, so the number lives only where it updates itself.

## Changes

- The static `tests-715 passed` shields badge in README.md is replaced with the real workflow badge for `ci.yml`, so it reflects actual CI state instead of a frozen figure.
- The install comment, the empirical-verification block, the quick-start comment and the Status-and-Limitations line no longer quote a pass count. Each now points at `python -m pytest tests/ -q` or at the `ci.yml` workflow runs. The coverage claim states the enforced 80 percent gate instead of a frozen percentage.
- `project.md` section 15 no longer quotes a count and defers to the per-leg CI totals.
- New guard test `tests/test_no_hardcoded_test_counts.py` scans both files for `\d{2,4}\s+(passing|passed|tests?)` and fails with a pointer to the issue if a count is committed again. This makes future drift a CI failure instead of a doc smell.

## Why no number was written

Counts are environment-dependent, which the issue itself demonstrates: the same suite reports different passed/skipped totals per CI leg and per local extras. I measured the static suite at 823 test functions (852 collected items after parametrize expansion) while recent CI legs report 837 passed with 2 skipped on the core matrix and 857 passed with the ml extra, so any single quoted figure would be wrong for someone immediately.

The guard test passes against the updated files and fails against the pre-change ones.

## Summary by CodeRabbit

- **Documentation**
  - Updated testing guidance to reference CI workflow results instead of fixed test counts.
  - Clarified that test totals vary by environment and optional extras.
  - Documented per-operating-system coverage requirements enforced in CI.
  - Updated the test status badge to reflect current CI results.

- **Tests**
  - Added automated validation to prevent outdated hard-coded test counts from appearing in project documentation.


Cyrax321 (approved):
Thank you for contributing to Snagline, and welcome to the project!

This is a solid first contribution. You addressed the stale test count issue (#310) by removing the hand-maintained numbers that had drifted across multiple documentation locations.

I reviewed the changes and verified that all CI checks pass. The implementation correctly replaces the frozen badge with the live CI workflow badge, updates the affected documentation to reference the appropriate sources, and adds a guard test to prevent the issue from recurring.

No blocking issues remain. Marking this as ready to merge.

Thank you again for taking the time to contribute and improve the project.

Issue #298, opened 2026-09-18
perf: window auto-scaling makes per-step ingest O(window), not O(1) (2x regression at max_window=512)


Adhi1-2:
## Summary

The README documents every tier-1 detector as **O(1) amortized per step**, and the published `snagline bench` numbers back that up. That claim is accurate **only with the default config**. When window auto-scaling is enabled (`window_scale_steps > 0`, issue #92) — i.e. exactly the long-horizon scenario scaling was built for — the loop and error-cascade detectors degrade to **O(window)** per step, because they rescan the whole window every step instead of maintaining a running count.

## Measurement (upstream `master`, `PYTHONPATH=src`)

Full `Monitor` ingest cost per step, same synthetic stream, only `window_scale_steps`/`max_window` varied:

| config | median | p99 |
|---|---|---|
| default (no scaling) | 1.88 µs | 1.95 µs |
| scale, `max_window=32` | 2.21 µs | 2.25 µs |
| scale, `max_window=128` | 2.97 µs | 3.06 µs |
| scale, `max_window=512` | **3.92 µs** | 5.65 µs |
| scale, `max_window=2048` | 3.96 µs | 5.62 µs |

That is a **2.1x slowdown** on the path the feature exists to serve, and it does not plateau — p99 keeps climbing with the cap.

Per-detector isolation confirms the cost is isolated to the window-rescanning detectors; `LatencyAnomalyDetector` is genuinely O(1) and stays flat:

| max_window | loop | cascade | latency |
|---|---|---|---|
| 32 | 1.53 µs | 1.51 µs | 1.25 µs |
| 512 | 2.79 µs | 1.97 µs | 1.24 µs |
| 2048 | 3.18 µs | 1.96 µs | 1.25 µs |

Cycle mode is the worst offender: **14.09 µs/step scaled vs 3.18 µs default** (4.4x), from `_minimal_period` scanning `O(max_period x window)` per step.

## Root causes

**1. `src/snagline/detectors/loop.py`, `_observe_plain`**
```python
w.append(event.action_signature)
count = w.count(event.action_signature)   # O(window) every step
```

**2. `src/snagline/detectors/error_cascade.py`, `observe`**
```python
density_alarm = (sum(w) >= self.error_threshold and len(w) >= self.error_threshold)
```
and `sum(w)` is then **computed again** in the detail string when the alarm fires.

[...]

PR #337, opened 2026-09-20
fix(config): validate the CUSUM alarm bars, sigma floors and window knobs; bound baseline retention (#331, #332, #333, #346, #351)

### LINKED ISSUE #331: bug: cusum_h / token_cusum_h of 0 raises ZeroDivisionError that fail-open swallows, and a negative h spams false positives
## the finding

`cusum_h` (the latency CUSUM alarm bar, default 5.0) and `token_cusum_h` (the token-runaway CUSUM alarm bar, default 5.0) are the denominators of each detector's risk-score computation, but neither is range-checked anywhere. Every config layer accepts `0.0` and any negative value without complaint:

```
Config(cusum_h=0.0).cusum_h          # 0.0   -- accepted
Config.from_env({"SNAGLINE_CUSUM_H": "0.0"}).cusum_h   # 0.0   -- accepted
Config(cusum_h=-5.0).cusum_h         # -5.0  -- accepted
```

Two distinct failure modes follow, both silent in different ways.

### 1. `h = 0.0` -> `ZeroDivisionError` -> detector dead for the whole run

`state.update()` alarms when `cusum > h`; with `h = 0` any positive deviation alarms, and the very next line divides by `h`:

```python
# src/snagline/detectors/latency_anomaly.py:267
score = min(1.0, 0.6 + 0.1 * max(0.0, alarm_cusum / self.h - 1.0))
#                                              ~~~~~~~~~~~~^~~~~~~~~
# ZeroDivisionError: float division by zero
```

The same expression exists in `token_runaway.py:112`. `Monitor.ingest` is fail-open, so the exception is swallowed and logged once:

```
$ python -c '...'
snagline: detector latency_anomaly raised; ignoring (fail-open)
events_ingested: 30
detector_errors: 24
risks_emitted: 1     # the loop detector only; the latency detector produced nothing
```

A user who set `SNAGLINE_CUSUM_H=0` intending "make this detector very sensitive" instead gets a permanently dead latency detector and a single warning line that is easy to miss in a busy log. The detector stays installed and reporting nothing for the entire run.

### 2. `h < 0` -> alarm fires on every step

`cusum` is clamped to `>= 0` by construction, so `cusum > h` is true for any negative `h` on any input. Perfectly healthy traffic pages constantly:

```
negative h: alarms on 7/10 perfectly NORMAL steps (false-positive spam)
```

## why this is a real gap and not already covered

Issues #317 and #322 fixed neighbouring knobs, and I want to be explicit that this one fell through:

- **#322 / PR #329** added `_validated_divisor_thresholds`, but it only covers the *count* knobs: `cascade_consecutive_threshold`, `cascade_error_threshold`, `loop_repeat_threshold`, `loop_stall_steps`. The CUSUM `h` knobs are floats and divide in a different expression, so they are untouched.
- **#317 / PR #318** added `_validated_token_budget`, which covers `episode_token_budget` and `token_budget_warn_fraction` only. `token_cusum_h` is untouched.

Neither covers `cusum_h` or `token_cusum_h`.

## suggested fix

Mirror the existing precedent exactly — a `_validated_cusum` helper called from `Config.__post_init__` and `Config.resolve`:

```python
def _validated_cusum(cfg: Config) -> None:
    for name in ("cusum_h", "token_cusum_h"):
        value = getattr(cfg, name)
        if not value > 0:
            raise ValueError(
                f"{name} must be positive; got {value!r}. The CUSUM alarm bar is "
                "the denominator of the risk score, so 0 raises ZeroDivisionError "
                "at ingest time and a negative value makes the alarm trivially true"
            )
```

`h > 0` is the honest constraint: it is an alarm bar, so 0 and negatives are never meaningful, and unlike an integer count knob there is no "disable this detector" reading to preserve (`cusum_h` has no disabled state; the detectors are opt-in via their own `*_enabled` flags).

Optional belt-and-braces, if you'd rather the runtime also not depend on config validation: clamp in each detector's constructor, e.g. `self.h = max(h, 1e-9)`, so a hand-constructed detector cannot divide by zero. I'd lean toward validation alone to stay consistent with how #119/#132/#322 handle this class of problem.

I have **not** opened a PR — the fix touches `config.py`, which both #318 and #329 also touch, so it should land after one of them rather than stack awkwardly. Happy to send it whenever you want it.

## reproduction

```python

mon = Monitor.default(config=Config(cusum_h=0.0, cusum_min_samples=3))
for i in range(6):    # warm-up
    mon.ingest(StepEvent("s", "e", 1.0, "tool_call", "x", tool_name="search", latency_ms=50.0))
for i in range(24):   # 100x latency anomaly -- should page, never does
    mon.ingest(StepEvent("s", "e", 1.0, "tool_call", "x", tool_name="search", latency_ms=5000.0))
print(mon.metrics())  # detector_errors: 24, risks from latency_anomaly: 0
```

Verified against `origin/master` (90b3519).

### LINKED ISSUE #332: bug: BaselineStore max_versions <= 0 prunes the baseline it just wrote, so the store silently ends up empty
## the finding

`BaselineStore.save()` writes the version history entry and the `latest.json` pointer, then calls `_prune` to enforce the retention cap. `_prune` has no lower bound on `limit`:

```python
# src/snagline/baseline_store.py:96
def _prune(self, tenant, deployment, limit):
    versions_dir = self._scope_dir(tenant, deployment) / "versions"
    ...
    existing = sorted(p.name for p in versions_dir.glob("*.json"))
    excess = existing[: max(0, len(existing) - limit)]
    for name in excess:
        (versions_dir / name).unlink(missing_ok=True)
```

With `limit <= 0`, `len(existing) - limit` is `>= len(existing)`, so **every** version file is deleted — including the one `save()` wrote in the same call. The store is left with no history and no pointer, and `load()` returns `None`:

```
after save with max_versions=0:
  list_versions: []
  load() -> None
  latest.json exists: False
```

The write and the delete happen in one call, so to the caller `save()` reported success and returned a version id that no longer resolves. `load_version(tenant, deployment, version)` on that returned id returns `None`.

Negative limits do the same thing, since `max(0, len(existing) - limit)` only grows:

```
versions with max_versions=-3: []
```

## why the value can actually reach 0

The CLI does not guard it either:

```python
# src/snagline/cli.py:752
store = BaselineStore(args.store_dir, max_versions=args.max_versions or 10)
```

`args.max_versions or 10` treats `0` as "unset" and silently substitutes `10`, so `--max-versions 0` does not reproduce this through the CLI. But `--max-versions -1` (and any negative value) is passed straight through to the store and destroys the history on the next save. That is a documented public knob with no range check in `Config`, `BaselineStore.__init__`, `_prune`, or the CLI.

The `max(0, ...)` clamp was written to keep the slice non-negative, which it does, but it does not bound `limit` itself — the expression is `existing[:max(0, len(existing) - limit)]`, and a non-positive `limit` makes the slice the whole list.

## impact

`BaselineStore` is the storage layer behind `goal_drift` and the `snagline baseline retrain` cadence (issue #102). An operator who sets `--max-versions 0` intending "keep only the latest" (a natural reading — `latest.json` is the pointer, so `0` history files sounds reasonable) instead loses the baseline entirely, and `goal_drift` silently degrades to "no reference profile", which is the exact silent-misconfiguration class `_validated_stagnation` and `_validated_horizon` were written to prevent.

It is also self-inconsistent: the module docstring promises "keeps the latest profile plus a bounded history of past versions", and `max_versions=10` keeps 10, but `max_versions=1` deletes the history entry it just wrote while `latest.json` survives — so `1` already means "0 history files plus a pointer", and `0` means "nothing at all". The off-by-one is real but harmless; the data loss is not.

## suggested fix

Reject the nonsensical range where it enters, matching the `Config` validation precedent (issue #119/#132/#322):

```python
def __init__(self, root_dir: str, max_versions: int = 10) -> None:
    self._root = Path(root_dir)
    if max_versions < 1:
        raise ValueError(f"max_versions must be >= 1; got {max_versions!r}")
    self._max_versions = max_versions
```

and in the CLI, validate the explicit knob instead of the truthiness fallback:

```python
limit = args.max_versions if args.max_versions is not None else 10
if limit < 1:
    print("snagline baseline: --max-versions must be >= 1", file=sys.stderr)
    return 2
store = BaselineStore(args.store_dir, max_versions=limit)
```

Alternatively, if you would rather `0` mean "keep only the pointer", clamp in `_prune` (`limit = max(1, limit)`) — but I'd raise, because `0` is far more likely a typo or a misunderstanding than a deliberate "retain no history" request, and a silent empty store is the worst possible outcome either way.

I have **not** opened a PR — `baseline_store.py` and `cli.py` are both touched by my open #328, so this should land after that rather than stack. Happy to send it whenever suits.

## reproduction

```python
from snagline.baseline_store import BaselineStore
from snagline.baseline import BaselineProfile
import tempfile

st = BaselineStore(tempfile.mkdtemp())
v = st.save(BaselineProfile(), version="v1", max_versions=0)
print("save returned:", v)
print("list_versions:", st.list_versions())   # []
print("load:", st.load())                      # None
```

Verified against `origin/master` (90b3519).

### LINKED ISSUE #333: bug: a 0 or negative *_min_samples / window-size knob is accepted and either disables the detector or fabricates alerts
## the finding

Following the same thread as #331 (unvalidated detector knobs), the **sample-count and window-size** knobs are also unchecked. `Config` rejects `stagnation_window_size < 1` and `max_window < 1` but nothing else in that family, and most of them misbehave badly at `0` or below.

Verified one by one against `origin/master` (90b3519):

| knob | `0` accepted? | behaviour at `0` |
|---|---|---|
| `cusum_min_samples` | yes | freezes the baseline on the **first** sample — warm-up is skipped |
| `token_min_samples` | yes | same, for the token CUSUM |
| `loop_window_size` | yes | plain loop detector **never fires** — the window can never fill |
| `cascade_window_size` | yes | windowed density alarm never fires (consecutive alarm still works) |
| `meltdown_window_size` | yes | fires a **fabricated** alert on step 0 |
| `meltdown_rearm_steps` | yes | harmless (re-arms every clear step) |

Two of these are silent-data-loss, one is a fabricated false positive.

### `meltdown_window_size = 0` fabricates an alert

The window is `deque(maxlen=0)` — always empty — but `len(w.window) < target` is `0 < 0` = `False`, so the detector proceeds to score an empty window. Entropy of an empty window is `0.0`, which is below `meltdown_low_entropy`, so it emits a real risk:

```
fired at step 0 -> 'tool-choice entropy collapsed to 0.00 bits (0 distinct in last 0 steps)' | score 0.7
  window contents: [] | len: 0
```

A score-0.7 (warning-band) alert naming "0 distinct in last 0 steps" is served on the very first step of an episode, every episode, with no evidence behind it.

### `*_min_samples <= 0` skips warm-up and invents a baseline

`state.n >= self.min_samples` is true at `n = 1`, so `freeze()` runs on the first event. The baseline becomes whatever that one sample was, so a single unrepresentative first step defines "healthy" for the rest of the episode. A one-off fast first step (a warm cache hit) then makes every subsequent normal step look anomalous:

```
min_samples=0, fast first step (5ms) -> 30 false alarms on 30 normal 60ms steps
  e.g. latency 60ms deviates from baseline (mean 5ms)
```

The documented purpose of `cusum_min_samples` ("learn a baseline before alarming", lowered to 5 in #9) is voided by a value that is accepted by every config layer.

### `loop_window_size = 0` silently disables the loop detector

`next_window` builds `deque(maxlen=0)`; `w.append()` discards everything, the count is always 0, and the detector returns `None` forever on a pure-repeat episode that should page on step 3.

## why this class keeps slipping through

Each detector validates *its own* knobs in some places and not others — `StagnationDetector.__init__` raises on `window_size < 1`, `MeltdownDetector.__init__` raises on `low >= high`, but nothing checks the window sizes or sample counts, and `Config` only validates the stagnation and horizon families. The pattern that already exists (and works) is a `_validated_*` helper called from both `__post_init__` and `Config.resolve`, so the file and env layers are both covered (issue #119 precedent).

## suggested fix

One helper covering the count/size knobs, mirroring `_validated_stagnation`:

```python
def _validated_counts_and_windows(cfg: Config) -> None:
    knobs = (
        ("cusum_min_samples", cfg.cusum_min_samples),
        ("token_min_samples", cfg.token_min_samples),
        ("loop_window_size", cfg.loop_window_size),
        ("cascade_window_size", cfg.cascade_window_size),
        ("meltdown_window_size", cfg.meltdown_window_size),
        ("loop_cycle_window_size", cfg.loop_cycle_window_size),
        ("meltdown_rearm_steps", cfg.meltdown_rearm_steps),
    )
    for name, value in knobs:
        if value < 1:
            raise ValueError(
                f"{name} must be >= 1; got {value!r}. A window or sample count of 0 "
                "either makes the detector's condition unreachable (silent disable) "
                "or lets it score an empty window (fabricated alerts)"
            )
```

called from `__post_init__` and `resolve`, next to the existing `_validated_stagnation` / `_validated_max_live_episodes` calls.

This is adjacent to #331 (CUSUM `h`) and #322 (count divisors), and the three together would close the whole knob-validation surface. I have **not** opened a PR — `config.py` is touched by my open #328 and #329, so these should land after one of them rather than stack. Happy to send whenever you want.

## reproduction

```python
from snagline.detectors.meltdown import MeltdownDetector

d = MeltdownDetector(config=Config(meltdown_enabled=True, meltdown_window_size=0))
print(d.observe(StepEvent("s", "e", 1.0, "tool_call", "x")))
# FailureRisk(... score=0.7, detail='...0 distinct in last 0 steps') on step 0
```

### LINKED ISSUE #346: bug: meltdown_window_size of 1 or 2 fabricates a collapse alert from a one-tool episode
## Description

`MeltdownDetector`'s entropy statistic only becomes meaningful once the window holds several tool identities: the entropy of a one-item window is exactly `0.0`, which is below `meltdown_low_entropy` (default `0.4`) by construction. `Config._validated_counts_and_windows` rejects `meltdown_window_size == 0` and its docstring spells out the exact failure it is preventing —

> `MeltdownDetector`'s entropy of an empty window is 0.0, which is below `meltdown_low_entropy`, so a score-0.7 risk naming "0 distinct in last 0 steps" fires on step 0 of every episode.

— but `1` (and `2`) pass the same `>= 1` check and produce the same fabricated alert, one step later. The thresholds are documented as tuned for window 20, and the module docstring's quiet-zone reasoning ("uniform alternation across eight tools (~2.97 bits at window 20)") assumes a window large enough for a distribution to exist.

## Current Behavior

```python
from snagline.detectors.meltdown import MeltdownDetector

def step(i):
                     make_signature("tool_call", "t", "a"), "t")

MeltdownDetector(window_size=1).observe(step(0))
# FailureRisk(episode_id='ep', step_id='0', score=0.7, trigger='meltdown',
#             detail='tool-choice entropy collapsed to 0.00 bits (1 distinct in last 1 steps)')
```

`window_size=2` is silent on step 0 (the window has not filled) and fires on the first repeat:

```
step 0 -> None
step 1 -> FailureRisk(..., detail='tool-choice entropy collapsed to 0.00 bits (2 distinct in last 2 steps)')
step 2 -> None
```

So a single-tool episode — the ordinary case of an agent that happens to call one tool twice — pages at severity 0.7. Both values are accepted by `Config.resolve()` from `SNAGLINE_MELTDOWN_WINDOW_SIZE`.

## Expected Behavior

Reject windows too small for the entropy statistic to mean anything (the minimum that can express "collapse onto one tool among several" is 3), in `_validated_counts_and_windows`, alongside the existing `0` check.

## Context

Verified against master at `b667169`. Related to #333 (which filed the `0` case) but a distinct value range: the `0` check already in place does not cover this.

### LINKED ISSUE #351: bug: cusum_sigma_floor_abs / cusum_sigma_floor_rel of 0 raises ZeroDivisionError that fail-open swallows, deadening the latency detector
## Description

The config comment for the CUSUM sigma floors states exactly why they exist —

> A perfectly stable baseline has sample std 0, which would make every deviation infinite

— but neither floor is range-checked while the derived `sigma0` is used as a denominator in the hot path. Setting both to `0` is accepted by `Config.resolve()` and produces a guaranteed `ZeroDivisionError` at ingest time on a constant-latency baseline.

```python
# src/snagline/detectors/latency_anomaly.py:115
```

## Current Behavior

```python

cfg = Config(cusum_sigma_floor_abs=0.0, cusum_sigma_floor_rel=0.0, cusum_min_samples=3)
d = LatencyAnomalyDetector(config=cfg)

def ev(i):
                     make_signature("tool_call", "t", "a"), "t", latency_ms=100.0)

for i in range(4):
    d.observe(ev(i))           # constant -> std 0 -> sigma0 = max(0, 0, 0) = 0
d.observe(ev(4))
# ZeroDivisionError: float division by zero
```

The same path via the documented 12-factor layering:

```
SNAGLINE_CUSUM_SIGMA_FLOOR_ABS=0 SNAGLINE_CUSUM_SIGMA_FLOOR_REL=0
-> Config.resolve() accepts both as 0.0
```

Like #331 (`cusum_h = 0`) and #322, the exception is swallowed by `Monitor.ingest`'s fail-open handler, so the latency detector stays installed and reports nothing for the rest of the run.

## Expected Behavior

Range-check the floors at config time alongside the `cusum_h` validation (#331), or clamp them to a small positive minimum at construction — the same "reject at the door" pattern the existing `_validated_cusum` docstring argues for.

## Context

abyyxhek:
Closes the unvalidated knob surface across five issues: the CUSUM alarm bars (#331), the baseline retention cap (#332), the sample-count / window-size knobs (#333), the too-small meltdown window that slips past that check (#346), and the sigma floors that are the same hazard in a second place (#351). Each was silently accepted and each broke its detector in a different silent way.

## The five failure modes

**#331 — `cusum_h` / `token_cusum_h` of 0 kills a detector for the whole run.** Both are the *denominator* of their detector's risk score. `h = 0` makes `alarm_cusum / h` raise `ZeroDivisionError`, which `Monitor.ingest`'s fail-open handler swallows after logging once — the detector then stays installed and reports nothing for the rest of the run. A negative `h` is worse in the other direction: the CUSUM statistic is clamped to `>= 0`, so `cusum > h` is trivially true on every step and healthy traffic pages constantly.

**#332 — `BaselineStore(max_versions=0)` prunes the version it just wrote.** `max(0, len(existing) - limit)` bounds the *slice* but not `limit`; a non-positive limit makes the slice the whole list, so `save()` reports success and hands back a version id that no longer resolves.

**#333 — a 0 window either disables the detector or fabricates an alert.** `deque(maxlen=0)` discards every sample, so the plain loop detector can never fire on a pure-repeat episode. `MeltdownDetector` instead walks past `len(window) < target` (`0 < 0` is `False`) and scores an empty window: its entropy is `0.0`, below `meltdown_low_entropy`, so a score-0.7 risk naming "0 distinct in last 0 steps" fires on step 0 of every episode. `*_min_samples <= 0` freezes the baseline on the first sample, letting one warm-cache hit define "healthy".

**#346 — `meltdown_window_size` of 1 or 2 passes the `>= 1` check but fabricates the same alert.** The entropy of a one-item window is exactly `0.0` by construction, so it is always below the low bar and pages on the first step. A two-item window can only take `{0.0, 1.0}` bits, so the threshold is crossed by nothing but the degenerate case and fires on the first repeat of an agent that simply called one tool twice. Three is the smallest window able to express "collapse onto one tool *among several*": it admits a 2:1 split (~0.92 bits) that stays quiet.

**#351 — the sigma floors are the same hazard in a second place.** The frozen `sigma0` is `max(std, abs_floor, rel_floor * |mean|)` and is itself a denominator in the hot path, but neither floor was range-checked while the config comment right above them explains exactly why they exist. With both at 0 a perfectly stable baseline gives `sigma0 = 0` and the same swallowed `ZeroDivisionError`.

## Changes

- **`config.py`** — two helpers mirroring the existing `_validated_stagnation` precedent, called from both `__post_init__` and `resolve` so the constructor, env, and file layers are all covered:
  - `_validated_cusum` — `cusum_h` / `token_cusum_h` must be positive, and the sigma floors must be non-negative with at least one strictly positive (#351). There is no "disable" reading to preserve for `h`; the detectors are opt-in through their own `*_enabled` flags.
  - `_validated_counts_and_windows` — the seven count/window knobs must be `>= 1`, with a tighter floor of 3 for `meltdown_window_size` (#346). Deliberately leaves `stagnation_window_size` / `max_window` to the existing helpers and the documented "0 disables" knobs (`window_scale_steps`, `cusum_refit_every`) alone.
- **`baseline_store.py`** — `__init__` rejects `max_versions < 1`, and `save()` re-checks the effective limit *before any file is touched*, so a bad per-call override fails loudly instead of leaving a half-applied store behind. `max_versions=1` — the legitimate "keep only the latest" reading — still works.
- **`cli.py`** — `args.max_versions or 10` treated an explicit `0` as "unset" and silently substituted 10, while a negative value reached the store and destroyed the history. Both `baseline` and `baseline retrain` now resolve the explicit knob with a range check; a bad value is exit 2 with a message naming the flag.
- **`detectors/meltdown.py`** — a defense-in-depth `window_size < 3` guard, mirroring the existing `low < high` check, so the direct-constructor repro in #346 is closed too.
- **`detectors/latency_anomaly.py`** — `_floored_sigma` can no longer return 0 (defense in depth for a directly-built detector whose mean is 0, which scales the relative floor to 0 as well).

## Scope

Rebased on the current `master` tip so it applies with no conflicts. `_validated_token_runaway` from #317/#318 covers the *budget envelope* only; `token_cusum_h` was still unvalidated, so this is additive rather than duplicative. All existing meltdown tests use windows of 4, 8, or 20, so the new floor touches no shipped behaviour.

## Verification

- `tests/test_issues_331_333.py` — 84 tests covering every bad value at all three config layers, both store boundaries, both CLI commands, and the end-to-end symptoms (a 100x latency anomaly pages instead of producing `detector_errors: 24`; the fabricated step-0 meltdown alert can no longer be constructed; a genuine one-tool collapse still pages at window 3).
- Full suite: **935 passed, 2 skipped** (up from 830 on `master`), coverage 88.16%. `ruff check` and `ruff format --check` clean.
- Every new test was verified to fail against the pre-fix code, so they are real regression guards rather than assertions on the new behaviour alone.

Fixes #331. Fixes #332. Fixes #333. Fixes #346. Fixes #351.


src/snagline/config.py
```diff
@@ -225,6 +225,110 @@ def _validated_stagnation(cfg: Config) -> None:
...
+    The sigma floors are the same class of hazard (#351) and are checked
+    alongside ``h`` because they share the failure: the frozen ``sigma0`` is
+    ``max(std, abs_floor, rel_floor * |mean|)`` and is itself a denominator in
+    the hot path. A negative floor is meaningless (``max`` discards it), and
+    with both floors at 0 a perfectly stable baseline gives ``sigma0 = 0``,
+    which raises ZeroDivisionError on the first post-warm-up step and, via the
+    same fail-open path, deadens the detector for the run. At least one floor
+    must be strictly positive; both default positive, so stock configs pass.
+    """
+    for name in ("cusum_h", "token_cusum_h"):
```
Adhi1-2, line 251: The two `*_k` slack knobs (`cusum_k`, `token_cusum_k`) are validated nowhere in this PR, and a negative value adds to the CUSUM every step — repro and proposed `>= 0` bound in the main review body.

Adhi1-2 (commented):
Good catch on the `h` denominators — same failure class I hit in the semantic goal-drift knobs (#370), and `not value > 0` is the right bound there.

One gap in the same family: **`cusum_k` and `token_cusum_k` are the slack subtracted per step, and neither is covered here.** A negative `k` does not divide anything, so it does not raise — it *adds* `|k|` to the CUSUM on every step unconditionally, which guarantees the alarm trips on traffic the baseline describes perfectly:

```
cusum = max(0.0, cusum + (x - mu0) / sigma0 - k)   # k < 0 -> -k > 0 every step
```

Reproduced against this branch (60-step episode, latency held **exactly** at the warm-up mean, so the standardized deviation is 0 and the only thing growing is `-k`):

| `cusum_k` | risks on a healthy episode |
|---|---|
| 0.5 (default) | 0 |
| 0.0 | 0 |
| -0.1 | 5 |
| -1.0 | **50** |

`token_cusum_k` is the same shape (`_WelfordCUSUM(self.k, self.h)`, `token_runaway.py:118`/`168`): `-1.0` gave 35 risks on constant, modest token usage.

The natural bound is `>= 0`, matching the `>= 0` you already use for the sigma floors and the reading I applied to `semantic_drift_cusum_k` in #371: **zero slack is legitimate** (a CUSUM accumulating raw standardized deviations with no allowance — sensitive, but not guaranteed to page); only a *negative* slack fabricates alerts. Both knobs coerce cleanly from `SNAGLINE_CUSUM_K=-1` / `SNAGLINE_TOKEN_CUSUM_K=-1`, so they slip past `from_env` and need the `resolve()` re-check too.

Happy to leave this to you rather than open a competing PR — it is your validator family and your call on the message. If you would rather I send it as a commit on this branch, say the word.

Adhi1-2:
Small gap in `COUNT_AND_WINDOW_KNOBS`, noticed while auditing the neighbouring comparison knobs: `goal_drift_min_samples` and `semantic_drift_min_samples` are warm-up counts of the same shape as the ones listed, but neither is in it. `goal_drift_min_samples=0` makes `live.total_steps < 0` never hold, so `_drift_score` runs from the first step on a one-sample live profile. `semantic_drift_min_samples` gates `GoalDriftDetector.observe` the same way. Both are `>= 1` candidates and would fit the existing loop in one line each.

Not filing a competing PR for it -- this looks like a natural extension of yours, and #387 (the score-threshold family) stays off that list on purpose. Also still happy to send the `cusum_k` / `token_cusum_k` `>= 0` bound from the earlier review as a commit on this branch if useful.

Cyrax321 (approved):
## Review — PR #337: validate CUSUM alarm bars, sigma floors, window knobs; bound baseline retention

Reviewed against `origin/master`. 6 files, +519/-5. Three commits, one per issue cluster (#331/#333, #346, #351).

### Verification performed
- ✅ **Full suite: 933 passed, 3 skipped** (vs 849 on master) — +84 new tests, **0 regressions**
- ✅ The 84 new tests in `tests/test_issues_331_333.py` — all pass
- ✅ **Bug #333 reproduced on master**: `Config(meltdown_window_size=0)` → `MeltdownDetector.observe()` returns a fabricated `score=0.7, trigger='meltdown', "0 distinct in last 0 steps"` on **step 0**. The branch rejects the config with a clear message.
- ✅ **Bug #332 reproduced on master**: `BaselineStore(root, max_versions=0)` → `capture_from_jsonl()` returns a version id, then `list_versions()` returns `[]` — the store silently deleted the version it just wrote while reporting success. The branch rejects `max_versions < 1` at construction *and* per-call.
- ✅ **Mutation-tested the guards**: reverting `meltdown.py` to master fails `test_meltdown_window_size_zero_cannot_fabricate_a_step_zero_alert` ("DID NOT RAISE"); reverting `config.py` rejects nothing. Both layers are load-bearing.
- ✅ **No over-reach**: a valid config still pages on a real 100× latency spike (100 alerts over the spike) — validation constrains only the degenerate values.
- ✅ Verified the three-layer coverage: constructor, `Config.resolve()` (env), and direct detector construction are all guarded.

### What's done well

- **Every floor is justified, not chosen.** The `meltdown_window_size >= 3` rule isn't arbitrary — the description proves it from the entropy statistic: a 1-item window is exactly 0.0 bits by construction (always below the low bar → pages on step 0), a 2-item window can only take {0.0, 1.0} bits (pages on one tool called twice), and 3 admits a 2:1 split at ~0.92 bits that stays quiet. That's the reasoning that makes the threshold trustworthy instead of a guess.
- **The failure modes are characterized, not just fixed.** The config docstrings distinguish the two directions a bad window breaks things — "condition unreachable (silent disable)" vs "score an empty window (fabricated alerts)" — and explain *why* `h = 0` is a swallowed `ZeroDivisionError` while negative `h` is constant false-positive paging. Future readers won't re-derive it.
- **The `save()` check moved before any file write.** A per-call `max_versions=0` override on a well-configured store is now rejected before touching the store, rather than leaving a half-applied history behind. That's an atomicity improvement, not just a validation one.
- **`_resolved_max_versions` fixes a real footgun** — `args.max_versions or 10` treated an explicit `0` as "unset" and silently substituted 10, so the flag appeared to work while doing something the operator never asked for.
- **Defense in depth at the right layer.** Config validation is the front door, but `MeltdownDetector` and `_floored_sigma` also guard, because `Config` is a mutable dataclass and detectors get constructed directly in tests and adapters. The latency guard's `1e-9` fallback is a last-resort denominator rather than an error, which is the correct posture inside a hot path.
- **Scope is disciplined.** `stagnation_window_size` / `max_window` are left to existing helpers, and the documented "0 disables" knobs (`window_scale_steps`, `cusum_refit_every`) are deliberately untouched. The description also notes this is additive to #317's budget validation rather than duplicative.

### Findings

**1. Conflicts with PR #294 (already approved) — needs a rebase** — *blocking for merge order, not for the change itself*

I simulated the merge: **#294 and #337 conflict in `src/snagline/cli.py`**. Both rewrite the store-construction line in `_cmd_baseline`:

```python
# #294's version (extracted read-only helper):
store = BaselineStore(args.store_dir, max_versions=args.max_versions or 10)

# #337's version:
limit = _resolved_max_versions(args)
if limit is None:
    return 2
store = BaselineStore(args.store_dir, max_versions=limit)
```

They also touch adjacent lines in the same block (#294 moves the `--list-versions` guard to the top of the function; #337 leaves it where it is). The conflict is trivial to resolve by hand — #337's `_resolved_max_versions` is strictly better and should win, since it also fixes the `or 10` footgun that #294's line preserves — but whichever lands second will need a rebase. **Recommend merging #294 first, then rebasing #337 on it.** Note this interaction is *also* why #378 (changelog backfill) should be sequenced last.

**2. `1e-9` fallback in `_floored_sigma` is a silent magic number** — *minor*

The guard is correct defense-in-depth, but if it ever fires on a directly-built detector, the CUSUM will then divide by 1e-9 and produce an astronomically large statistic — which, given the alarm is `cusum > h`, means the detector pages on the next step. That's arguably the right failure mode (fail-loud rather than deaden), and it's documented as reachable only when both floors are disabled (which Config now rejects). But the comment doesn't say what behavior it produces, only that it "keeps sigma0 a usable denominator". One line noting "this makes the next update alarm, which is the loud failure we prefer over deadening the detector" would make the intent self-documenting. Not blocking.

**3. Overlaps the #368/#371/#387 validation family** — *coordination note*

This PR adds `_validated_cusum` and `_validated_counts_and_windows` to `config.py`'s `__post_init__`/`resolve()`. PRs #368, #371, and #387 in the queue add *their own* validators to the same two methods for different knob families. All four are independently correct, but they will stack four separate validator calls in sequence, and any two that touch the same knob (e.g. #387 range-checks ml-ensemble *score* knobs, which are adjacent to but distinct from these) need to be merged in a consistent order. Worth a single follow-up that consolidates the validator registry once the family lands, so `__post_init__` doesn't become a list of six bespoke helpers.

### Verdict

✅ **Approve.** This is the best-reasoned PR in the batch so far — every threshold is derived from the statistic it constrains, both directions of each failure mode are characterized, and the guards are verified load-bearing by mutation testing. I reproduced two of the five bugs on master before accepting the fix, and confirmed valid configurations still page on genuine anomalies. The only merge-ordering wrinkle is the conflict with #294 (finding 1), which is a one-line resolution in #337's favour.

Cyrax321:
@copilot resolve the merge conflicts in this pull request

Cyrax321 (commented):
**Code Review — PR #337**

The validations are mostly sound and the 84 new tests pass, but the PR is stacked on a stale base: its head branch lacks the merged #298/#307 zero-window contract, so its `loop_window_size`/`cascade_window_size >= 1` check contradicts a merged master test and turns CI red on rebase. Several new guards are also non-finite-unsafe, the meltdown floor fires even when the detector is disabled, and `cusum_k`/`token_cusum_k` are left unvalidated despite being the same hazard class.

Findings:

**1. `high` — `_validated_counts_and_windows` rejects `loop_window_size`/`cascade_window_size < 1`, contradicting the merged master test and windowing contract.**
`src/snagline/config.py:299-306`
Verified: master's `windowing.py:103-109` comment says "A zero-capacity window (a detector configured with `window_size=0`, which Config does not reject) drops every item", and master's `tests/test_window_scaling.py::test_zero_window_size_does_not_raise_under_scaling[loop|cascade]` constructs `Config(window_scale_steps=5, max_window=8, loop_window_size=0)` expecting no raise. Applying PR337's config changes to a clone of master and running the file yields `2 failed, 35 passed` — `ValueError: cascade_window_size must be >= 1; got 0` at `config.py:306`. The head `f7283fd` predates `2d61f86`, so this lands the contradiction, and since #422 and others stack on #337 it propagates. Reconcile the zero-window fail-safe contract with the new floor before merging.

**2. `medium` — the new guards are inconsistent on non-finite input (`<` comparisons are NaN-false).**
`src/snagline/config.py:264` (and `306,322`)
`value < 0` for the sigma floors and `value < 1` for counts/windows are both `False` for NaN, so NaN slips through. Reproduced: `Config(cusum_sigma_floor_abs=nan, cusum_sigma_floor_rel=0.0)` is accepted, and the new `if spread <= 0.0` is also NaN-false, so `sigma0` becomes NaN and the CUSUM silently never fires (a 50x spike returned None). `Config.load_file('{"loop_window_size": NaN}')` is accepted and later raises `TypeError` inside `deque`, swallowed to a dead detector; `meltdown_window_size=nan` passes both `< 1` and `< 3`, bypassing the #346 floor. The `h` guard uses the NaN-safe `not value > 0` (`config.py:253`), so only some of the new checks are safe. #422 already patches the `k` half; make the rest consistent.

**3. `medium` — the meltdown `>= 3` floor is enforced even when `meltdown_enabled` is False (the default), breaking the opt-in convention.**
`src/snagline/config.py:322`
Reproduced: `Config(meltdown_window_size=1)` raises despite `meltdown_enabled` defaulting False. A legacy config with `meltdown_window_size: 1` but `meltdown_enabled: false` previously started fine; now startup aborts on an inert knob. Sibling validators bite only when the feature is opted in (`_validated_token_runaway` only when `episode_token_budget is not None`, `config.py:142`); the docstring's "stock configurations never hit these checks" holds only for default values, not disabled-but-set knobs.

**4. `medium` — `cusum_k`/`token_cusum_k` remain unvalidated, though a negative `k` is the same "trivially-true alarm" hazard as negative `h`.**
`src/snagline/config.py:349,391,436`
Reproduced on head: `Config(cusum_k=-5.0)` accepted, and `LatencyAnomalyDetector(min_samples=3, k=-5.0, h=5.0)` on constant 100 ms latency alarms scoring `[0.7,0.8,0.9,1.0,...]`. Since `cusum = max(0, cusum + (x-mu0)/sigma0 - k)`, a negative `k` adds `|k|` every step and crosses `h` on healthy traffic — exactly the failure this PR documents for `h`. (PR #422 stacks the `k >= 0` check on top; it belongs in this validator.)

**5. `low` — `cusum_h`/`token_cusum_h` accept +infinity, which also deadens the detector.**
`src/snagline/config.py:253` — `not value > 0` is False for `inf`, so `Config(cusum_h=inf)` and `SNAGLINE_CUSUM_H=inf` are accepted; `cusum > inf` can never be true, so a sustained anomaly never pages. Non-finite rejection (PR #412) closes it.

**6. `low` — `_resolved_max_versions` hardcodes the store default (10) and the range check now gates the read-only `--list-versions` path.**
`src/snagline/cli.py:740-747`
Head builds the store with the resolved limit before the `list_versions` branch, so `snagline baseline --store-dir X --list-versions --max-versions 0` exits 2 on a knob listing never uses (master handles list-versions first); the message also hardcodes the `snagline baseline:` prefix for the `baseline retrain` subcommand. Minor UX/contract issue, not data loss.

Adhi1-2:
`_validated_counts_and_windows` rejects `loop_window_size < 1` and `cascade_window_size < 1` at `Config` construction, which breaks an existing merged test and leaves this PR CI-red as submitted.

`tests/test_window_scaling.py::test_zero_window_size_does_not_raise_under_scaling` (from #307) deliberately builds `Config(loop_window_size=0)` and `Config(cascade_window_size=0)` and asserts the detectors stay **fail-safe silent** under window scaling rather than raising. This PR doesn't touch that test, so `pytest tests/` (the `test` job) now fails:

```
tests/test_window_scaling.py::test_zero_window_size_does_not_raise_under_scaling[loop]    FAILED
tests/test_window_scaling.py::test_zero_window_size_does_not_raise_under_scaling[cascade] FAILED

E   ValueError: cascade_window_size must be >= 1; got 0. A window or sample count of 0
    either makes the detector's condition unreachable (a silent disable) or lets it
    score an empty window (fabricated alerts)
```
(reproduced against this branch's `config.py` on a clean HEAD checkout)

There's also a design collision underneath the test break: #307 established "zero base window = fail-safe silent under scaling" as a deliberate runtime contract (the CodeRabbit finding), and this PR flips it to "reject at startup." The docstring's fabricated-alert rationale is genuinely true for `meltdown_window_size` (an empty entropy window scores 0.0 bits and pages on step 0), but it does **not** apply to `loop_window_size`/`cascade_window_size`: #307 already proved `deque(maxlen=0)` silently discards every sample, so those two never fire — they're the "silent disable" half, not the "fabricated alert" half, and the two modes are conflated here.

Two ways forward:
- **Narrow the check:** drop `loop_window_size`/`cascade_window_size` from the rejected set, keeping the knobs whose 0 truly fabricates alerts or divides by zero (`cusum_min_samples`, `token_min_samples`, `meltdown_window_size`, `meltdown_rearm_steps`, `loop_cycle_window_size`). This preserves #307's contract.
- **Supersede #307 deliberately:** if the intent is that a zero loop/cascade window should now be a startup error, then update/remove `test_zero_window_size_does_not_raise_under_scaling` in the same PR and note the contract change, rather than leaving a merged test red.

The rest of the PR (the `cusum_h`/`token_cusum_h` and sigma-floor denominators, `meltdown_window_size >= 3`, the `BaselineStore` retention guard, and the `--max-versions` usage-error path) all reads correct.

Issue #322, opened 2026-09-19
bug: a 0 detector threshold knob raises ZeroDivisionError that fail-open swallows, silently disabling the detector for the rest of the run


abyyxhek:
### Current behavior

Four detector threshold knobs divide by themselves to compute a risk score. When any of them is `0`, the division raises `ZeroDivisionError` on the **first step of the run**:

- `src/snagline/detectors/error_cascade.py:120` — `consecutive / self.consecutive_threshold`
- `src/snagline/detectors/error_cascade.py:123` — `sum(w) / self.error_threshold`
- `src/snagline/detectors/loop.py:212` and `:263` — `count / self.repeat_threshold`
- `src/snagline/detectors/loop.py:344` — `count / self.loop_stall_steps`

A threshold of `0` is not caught earlier because the fire conditions are `>=` comparisons, which a zero threshold trivially satisfies:

```python
consecutive_alarm = consecutive >= self.consecutive_threshold   # 0 >= 0 == True
density_alarm = sum(w) >= self.error_threshold and len(w) >= self.error_threshold
```

So the detector reaches the division and dies immediately.

The consequence is worse than a crash. `Monitor.ingest` is fail-open by construction (`src/snagline/monitor.py:380-390`): the exception is logged once and swallowed, and the detector stays in `self._detectors` raising on every subsequent step. **The detector is silently dead for the entire run** while the host agent believes it is being watched. Nothing surfaces except a single warning at startup and a `detector_errors` counter.

### Expected behavior

Either a `ValueError` at config time, mirroring the existing validators for the other detectors (`_validated_stagnation`, `_validated_max_live_episodes` in `src/snagline/config.py:157-191`), or a documented, safe behaviour for `0`. `Config` is a plain dataclass with no validators on these four knobs, and the stagnation knobs already show the project's preferred pattern: `stagnation_min_novelty` is rejected with an explicit message that `0.0` *"would make the fire condition unreachable and silently disable the detector"* — which is precisely what happens here.

### Why a user would hit this

These knobs are settable through every layer of the 12-factor config (`Config.from_env_overrides` / `load_file` / `resolve`), and `0` is the natural "disable this detector" value someone reaches for:

```
$ SNAGLINE_CASCADE_CONSECUTIVE_THRESHOLD=0 SNAGLINE_LOOP_REPEAT_THRESHOLD=0 python -c "
from snagline.config import Config
c = Config.resolve()
[...]

PR #338, opened 2026-09-20
fix(config): reject non-positive CUSUM alarm bars (Fixes #331)

### LINKED ISSUE #331: bug: cusum_h / token_cusum_h of 0 raises ZeroDivisionError that fail-open swallows, and a negative h spams false positives

Adhi1-2:
## what

`cusum_h` (latency) and `token_cusum_h` (token runaway) are the denominators of their detectors' risk scores. Both out-of-range values fail in a direction that fail-open hides, and both were accepted by every config layer (`Config(...)`, `from_env_overrides`, `load_file`, `resolve`).

**`h == 0` → detector dead for the run.** The CUSUM alarms on `cusum > h`, then divides by `h` to score it (`latency_anomaly.py`, `token_runaway.py`). `Monitor.ingest` is fail-open, so the `ZeroDivisionError` is logged once and swallowed and the detector keeps raising on every step thereafter:

```
cusum_h=0.0 -> detector_errors: 18 risks: 0
```

**`h < 0` → false-positive spam.** `cusum` is clamped to `>= 0` by construction, so `cusum > h` holds for any negative `h` on any input. Ten perfectly normal steps, all at exactly the baseline mean:

```
cusum_h=-5.0: detector_errors=0 risks=5
```

## fix

`_validated_cusum_bars` rejects non-positive values at construction and again after env/file layering, following the `#119` / `#132` / `#317` precedent — invalid monitoring config aborts startup loudly, naming the knob, instead of misbehaving downstream. The detectors repeat the guard for direct construction with an explicit `h`, matching the `#317` defense-in-depth pattern (`TokenRunawayDetector` already did this for its budget knobs).

No upper bound is applied: the bars are multiples of sigma, so a large `h` is a legitimate "make this detector insensitive" rather than a defect.

Issue #331 notes this knob fell through both `#322`/PR #329 (`_validated_divisor_thresholds` covers the count knobs only) and `#317`/PR #318 (`_validated_token_budget` covers the budget envelope only) — this PR covers the remaining pair.

## scope

Both knobs ship defaults of `5.0`, so stock configurations never trip the new checks. Latency-anomaly detection is on by default; the token detector is opt-in.

- `src/snagline/config.py` — new `_validated_cusum_bars`, wired into `__post_init__` and `resolve`.
- `src/snagline/detectors/latency_anomaly.py`, `src/snagline/detectors/token_runaway.py` — constructor guard for explicit `h`.
- `tests/test_config_loader.py` — one test over both knobs, both failure values, both config layers, the constructor bypass, and the "large h is fine" case.
- `CHANGELOG.md` — entry under `[Unreleased] / Fixed`.

Verified: `ruff check`, `ruff format --check`, `mypy src tests`, and the full suite (850 passed, 3 skipped, 88% coverage) all clean.

Fixes #331

## Summary by CodeRabbit

- **Bug Fixes**
  - Configuration now rejects zero or negative CUSUM thresholds during startup and configuration resolution.
  - Invalid threshold values produce clear errors identifying the affected setting.
  - Latency and token monitoring detectors also validate thresholds when configured directly.
  - Prevents invalid values from causing monitoring to stop detecting issues or trigger alerts continuously.
- **Tests**
  - Added coverage for invalid, valid, default, large, and environment-provided threshold values.


Adhi1-2:
Closing as a duplicate — PR #337 (`fix/issues-331-333-knob-validation`, by abyyxhek) was opened two minutes before this one and covers #331 as part of a broader sweep that also closes #332 and #333. I missed it when cross-referencing; no point making the maintainer arbitrate two PRs over the same knobs.

For the record, this PR's scope was limited to the alarm bars, and it took a different line on #333's window knobs (a guard in `append_counted` rather than Config validation), so if #337's approach to the window family runs into trouble, the alternative is on the branch `Adhi1-2:fix/cusum-bar-range-331`.

PR #341, opened 2026-09-20
fix(config): warn by name when env var names a non-scalar field

### LINKED ISSUE #340: bug: a recognized-but-non-scalar env var (SNAGLINE_CALIBRATION_BASELINE / SNAGLINE_GOAL_DRIFT_BASELINE) is silently dropped with no warning
### Current behavior

`Config.from_env_overrides` deliberately keeps non-scalar fields out of reach of string coercion, but it does so *silently*. A user who reaches for the object-typed field instead of its documented `_PATH` variant gets neither effect nor feedback:

```
>>> Config.from_env_overrides({"SNAGLINE_CALIBRATION_BASELINE": "/tmp/baseline.json"})
overrides: {}      # silently empty -- no warning, no log line
```

Compare the two neighbouring cases, both of which do tell you something:

| env var you set | result |
| --- | --- |
| `SNAGLINE_CALIBRATION_BASELINE=/tmp/baseline.json` | silently ignored, no log |
| `SNAGLINE_TOTALLY_MADE_UP=1` (unknown key) | silently ignored, documented as such |
| `SNAGLINE_MAX_WINDOW=notanumber` (bad value) | `WARNING ignoring bad env ...` |

The first row is the one that bites: the name is real, it is in the type hints, and it looks exactly like the knob you want. Verified against master.

### Why it happens

`src/snagline/config.py:507-512`:

```python
hint = _coercible_hint(hints[name])
if hint in (bool, int, float, str):
    try:
        overrides[name] = _coerce(hint, value)
    except ValueError:
        logger.warning("snagline: ignoring bad env %s=%r", key, value)
# no else branch
```

`_coercible_hint` (config.py:37-51) returns `BaselineProfile | None` unchanged because the non-None member is not a scalar, so the `if` is False and execution falls off the end of the loop body. The ignore is correct -- you cannot build a `BaselineProfile` from a bare string -- but nothing is logged.

### Why it matters

The two affected fields are `calibration_baseline` and `goal_drift_baseline`, the only two `Config` fields absent from the README env-var table (the table documents their `calibration_baseline_path` / `goal_drift_baseline_path` string siblings instead). That is precisely the confusion this failure mode rewards: the correct knob is one `_PATH` away, and the mistake produces a fully silent baseline that never loads. A host pointing `SNAGLINE_CALIBRATION_BASELINE` at a JSON file would get baseline-less `calibration="auto"` with no indication why.

### Suggested fix

Add an `else` branch that warns, ideally naming the path variant so the message is actionable:

```python
else:
    logger.warning(
        "snagline: env %s sets %r, which cannot be set from the "
        "environment (not a scalar); ignoring", key, name,
    )
```

Small and self-contained, but worth a test asserting the warning fires for these two fields while unknown keys stay quiet.

Adhi1-2:
Fixes #340

## The bug

`SNAGLINE_CALIBRATION_BASELINE` looks exactly like the knob an operator wants — it is a real `Config` field, it has a `SNAGLINE_` prefix, and the docs promise every scalar field is settable this way. But no string can build a `BaselineProfile`, so `from_env_overrides` dropped it silently:

```python
hint = _coercible_hint(hints[name])
if hint in (bool, int, float, str):
    ...
# else: fall through, no log, no override
```

The result was a config that silently meant nothing. `SNAGLINE_CALIBRATION_BASELINE_PATH` (the `_PATH` variant) is the working spelling, but nothing told you that.

## The fix

The key is still ignored — a non-scalar field genuinely cannot be built from a string — but it now warns, naming the key and the reason:

```
snagline: ignoring env SNAGLINE_CALIBRATION_BASELINE='/tmp/baseline.json':
calibration_baseline is not a scalar field and cannot be set from the environment
```

This matches the existing tone for values that fail *coercion* (`ignoring bad env ...`), and is deliberately noisier than that path: a value that fails to coerce is probably a typo, while a key naming an object field is a *category* error the user will repeat until told.

Two fields are affected, both `BaselineProfile | None`: `calibration_baseline` and `goal_drift_baseline`.

Unknown keys (`SNAGLINE_BOGUS_FIELD`) and unrelated prefixed env (`PATH`) stay silent — a host legitimately passes those through, and warning on them would be noise (regression test added).

## Scope

One `else` branch in `from_env_overrides` + docstring. No behaviour change for any working config.

## Validation

- `tests/test_config_loader.py`: 23 passed, including two new tests — one parametrized over both affected fields asserting the warning is emitted *and* the field is still not set, one asserting unknown keys stay quiet.
- Full suite: 852 passed, 3 failures (`test_cli_mentions_read_timeout_next_to_max_body`, `test_real_actionledger_flags_review`, `test_instrument_openai_without_sdk_is_safe_noop`) — all three reproduce on the pristine base commit with my changes stashed, and the third is environmental: `openai` is installed in this interpreter, so "SDK is absent" cannot hold. CI installs only `langchain-core`.
- `ruff check` and `mypy src/snagline/config.py`: clean.


Adhi1-2:
Closing this in favour of #373, which fixes the same defect (issue #355, a duplicate of #340) with actionable per-field advice: it detects that `calibration_baseline_path` exists and points the operator at `SNAGLINE_CALIBRATION_BASELINE_PATH`, while correctly telling them to pass `goal_drift_baseline` in code — there is no `goal_drift_baseline_path` field, so this PR's generic message would have left the operator guessing either way.

Both PRs add the same `else` branch to `from_env_overrides`; keeping the richer one avoids making the maintainer choose.

Issue #335, opened 2026-09-20
bug: snagline.auto stream wrappers break the "with stream as s:" form, silently emitting zero events


Adhi1-2:
### Current behavior

The streaming wrappers returned when `stream=True` is used with `snagline.auto` do not support the context-manager protocol at all. This is the form used throughout both SDKs' documentation:

```python
with client.chat.completions.create(model="gpt-4o", messages=[...], stream=True) as stream:
    for chunk in stream:
        ...
```

It raises, and records nothing:

```
TypeError: '_SyncStreamWrapper' object does not support the context manager protocol
```

Reproduced against master (no SDK needed, a fake stream with `__enter__`/`__exit__` mimics `openai.Stream`):

```
Form 1 (for chunk in stream): events = 1
Form 2 (with ... as s): TypeError -> '_SyncStreamWrapper' object does not support the
                        context manager protocol | events = 0
```

The async twin fails the same way for `async with ... as stream:`.

### Why it happens

`_SyncStreamWrapper` proxies unknown attributes to the raw stream via `__getattr__` (`src/snagline/auto/openai.py:153`, `src/snagline/auto/anthropic.py:147`), and clearly intends that to cover `__enter__`/`__exit__`. But implicit special-method lookup for the `with` statement resolves on the **type**, never through `__getattr__`, so the dunder is invisible and the protocol is missing. The wrappers define `__iter__`/`__next__`/`close` but no `__enter__`/`__exit__`; `_AsyncStreamWrapper` likewise defines `__aiter__`/`__anext__`/`aclose` but no `__aenter__`/`__aexit__` (`openai.py:179-249`, `anthropic.py:173-243`).

### Why it matters

This is the exact failure mode issue #270 fixed for global mode: a monitored process that is not monitored at all. The `with` form is the primary documented streaming idiom for both the OpenAI and Anthropic SDKs, so a user following the docs gets a hard `TypeError` at best, and -- if their code path only touches the wrapper through the raw stream's own `__enter__` -- no telemetry whatsoever while believing it is covered. The whole point of `stream=True` deferral (issue #242) is silently skipped.

### Suggested fix

Define the protocol on the wrapper types themselves, returning `self` so iteration still goes through the wrapper:

- `_SyncStreamWrapper.__enter__` -> `self`; `__exit__` -> delegate to `close()` (which already emits once and closes the underlying stream)
[...]

PR #342, opened 2026-09-20
fix(auto): guard wrap_client against stacking wrappers

### LINKED ISSUE #336: bug: combining global and per-client snagline.auto instrumentation double-counts every call (two events per request)
### Current behavior

Global mode is idempotent: `_patch_resource_classes` skips any `create` already carrying the `__snagline_wrapped__` marker, and `tests/test_auto_global_mode.py:235` pins that. `wrap_client` (the per-client path) has **no** such guard, so composing the two modes -- or calling the per-client path twice -- stacks wrappers and emits one event per layer per call:

```
Case A (global + per-client): events = 2 (expected 1)
Case B (per-client x2):       events = 2 (expected 1)
```

Reproduced against master with a fake SDK shaped like the real one (a `cached_property` client surface plus resource classes), no SDK install required.

Two ways to hit it:

```python
instrument_openai(monitor)                    # global, patches the resource class
instrument_openai(monitor, client=client)     # then per-client: wraps the class wrapper
```

Here `wrap_client` resolves `client.chat.completions.create` to the *already-wrapped class attribute* and installs an instance attribute on top of it, so each call traverses both wrappers.

```python
instrument_openai(monitor, client=client)
instrument_openai(monitor, client=client)     # second call wraps the first wrapper
```

Same result. The pattern repeats across all three modules: `auto/openai.py:326-355`, `auto/anthropic.py:319-332`, `auto/langchain.py:88-95`.

### Why it matters

Every detector counts events. Two events per real request biases all of them: `error_cascade`'s threshold of 3 errors trips on ~2 real failures, `loop` and `stagnation` see doubled repetition counts and fire early, and the latency window contains each latency twice, skewing every baseline and CUSUM input. A host that instruments globally in a startup hook and then wraps a specific client at a call site -- a very natural thing to do -- gets silently miscalibrated detection with no warning.

### Suggested fix

Have `wrap_client` check the same sentinel the global path already uses before wrapping:

```python
if getattr(method, "__snagline_wrapped__", False):
    continue   # already instrumented globally or by a prior wrap_client
```

Two notes for whoever picks this up:

- `auto/langchain.py` is the odd one out: its `_wrap_one` never sets `__snagline_wrapped__` at all, so the sentinel needs to be applied there too, both by `_wrap_one` and in `wrap_client`.
- A second call that finds everything already wrapped is success, not a warning case. `wrap_client` currently warns when it patches nothing (`openai.py:350-354`); the already-wrapped path should stay quiet, the way `_patch_resource_classes` returns -1 to distinguish "fully instrumented" from "found nothing".

Adhi1-2:
Fixes #336

## The bug

Global mode skips any `create` already carrying `__snagline_wrapped__`, and `test_instrument_openai_global_is_idempotent` pins that. `wrap_client` — the per-client path — had no such guard, so composing the two modes stacked wrappers and emitted one event per layer per call.

Reproduced both cases from the issue against a fake SDK (no install needed) before fixing, and both now emit exactly one event:

```
Case A (global + per-client): events = 2 -> 1
Case B (per-client x2):       events = 2 -> 1
```

Why it matters more than a cosmetic double log: **every detector counts events.** `error_cascade`'s threshold of 3 trips on ~2 real failures, `loop` and `stagnation` see doubled repetition counts and fire early, and each latency enters the window twice, skewing every baseline and CUSUM input. A host that instruments globally in a startup hook and then wraps a specific client at a call site — a very natural composition — gets silently miscalibrated detection with no warning.

## The fix

`wrap_client` in all three modules now checks the same sentinel the global path uses, marks its own wrappers with it, and **stays quiet when it finds everything already wrapped**. That last part follows the issue's second note: global-then-per-client is success, not the `found no create method` warning path — it mirrors `_patch_resource_classes` returning `-1` to distinguish "fully instrumented" from "found nothing".

`auto/langchain.py` was the odd one out flagged in the issue: its `_wrap_one` never set `__snagline_wrapped__` at all. So the sentinel is now set there too — which fixes a case the issue didn't call out: langchain **global** mode called `wrap_client` on `BaseLanguageModel`/`Chain` with an unguarded, unmarked wrapper, so every `instrument_langchain()` call double-wrapped the class and every subsequent call in the process emitted twice. Pinned by a new test.

## Scope

- `wrap_client` in `openai.py`, `anthropic.py`, `langchain.py`: sentinel check + mark wrapper.
- No change to `_wrap_one`, `_patch_resource_classes`, or any emitted event.

## A fixture bug found along the way

The openai/anthropic test fakes held their resource object as a **class-level shared singleton**:

```python
class _FakeClient:
    messages = _FakeMessages()   # one object for the whole session
```

`wrap_client` sets `cur.create = wrapper` on that object, so a wrapper bound to one test's monitor stayed installed on every later test's client. The idempotency test initially failed with **0** events, not 2 — it inherited a wrapper whose monitor was already torn down. Resources are now per-instance, as the real SDK builds them per client. This was masking real state leakage between tests, not just affecting the new test.

## Validation

- 9 new tests across `test_auto_openai.py`, `test_auto_anthropic.py`, `test_auto_langchain.py`, `test_auto_global_mode.py`: per-client x2, global-then-per-client (both SDKs), langchain global x2, already-wrapped stays quiet, and nothing-to-patch still warns.
- Full suite: **861 passed**, coverage 88.08% (gate: 80%). 3 failures (`test_cli_mentions_read_timeout_next_to_max_body`, `test_real_actionledger_flags_review`, `test_instrument_openai_without_sdk_is_safe_noop`) — all three reproduce identically on the pristine base commit with changes stashed. The third is environmental: `openai` is installed in this interpreter, so \"SDK is absent\" cannot hold; CI installs only `langchain-core`.
- `ruff check src tests`: clean. `mypy src tests`: 7 errors, all pre-existing in `openai.py`'s `_patch_resource_classes` import block (verified same 7 with changes stashed); this PR adds none.


Adhi1-2:
Superseded by #344, which covers both #335 and #336 in a single PR and also updates CHANGELOG.md and docs/ATTACH_ANY_SYSTEM.md. Closing this one to avoid splitting reviewer time across duplicate PRs for the same issues.

PR #343, opened 2026-09-20
fix(auto): support the with/async with streaming form

### LINKED ISSUE #335: bug: snagline.auto stream wrappers break the "with stream as s:" form, silently emitting zero events

Adhi1-2:
Fixes #335

## The bug

```python
with client.chat.completions.create(model="gpt-4o", messages=[...], stream=True) as stream:
    for chunk in stream:
        ...
```

```
TypeError: '_SyncStreamWrapper' object does not support the context manager protocol
```

The wrappers proxy unknown attributes to the raw stream via `__getattr__`, which looks like it should cover the dunders. But **implicit special-method lookup for `with` resolves on the type, never through `__getattr__`**, so `__enter__`/`__exit__` were invisible. The wrappers defined `__iter__`/`__next__`/`close` but no context-manager protocol; the async twin likewise lacked `__aenter__`/`__aexit__`.

This is the #270 failure mode — a monitored process that is not monitored at all — and it silently skips the entire point of stream deferral (#242).

## The fix

- `__enter__`/`__aenter__` return `self`, so iteration stays on the wrapper and exhaustion still emits exactly once.
- `__exit__`/`__aexit__` delegate to `close()`/`aclose()`, which already emit once and close the underlying stream. The SDK's own `Stream.__exit__` *is* `self.close()`, so this is equivalent to the raw behaviour.
- One deviation from the issue's suggested fix, which I think is worth calling out: a bare `__exit__ = close` would record **success** for an exception escaping the `with` body. So `__exit__` emits with the real error first (guarded — if exhaustion already emitted, this is a no-op), then closes. Pinned by a test.

Both wrapper classes, in both `openai.py` and `anthropic.py`, where the code is duplicated verbatim.

## Validation

- 5 new tests in `tests/test_auto_stream.py` (which had **zero** coverage of the context-manager form): the `with` form for both SDKs, the `async with` form, and the exception-escapes-body path for both — asserting the event count, the error flag, and that the underlying stream is closed.
- **Verified the tests actually fail without the fix**: with `src/` stashed, all 5 fail with the exact `TypeError` from the issue, both sync and async protocols. With the fix, 12/12 pass.
- Full suite: **866 passed**, coverage 88.18% (gate: 80%). 3 failures (`test_cli_mentions_read_timeout_next_to_max_body`, `test_real_actionledger_flags_review`, `test_instrument_openai_without_sdk_is_safe_noop`) — all reproduce identically on the pristine base commit; the third is environmental (`openai` installed locally, so \"SDK is absent\" cannot hold; CI installs only `langchain-core`).
- `ruff check src tests`: clean. `mypy src tests`: 7 errors, all pre-existing in `openai.py`'s `_patch_resource_classes` import block (same 7 with changes stashed); none added.


Adhi1-2:
Superseded by #344, which covers both #335 and #336 in a single PR and also updates CHANGELOG.md and docs/ATTACH_ANY_SYSTEM.md. Closing this one to avoid splitting reviewer time across duplicate PRs for the same issues.

PR #344, opened 2026-09-20
fix(auto): support stream context managers, make wrap_client idempotent

### LINKED ISSUE #335: bug: snagline.auto stream wrappers break the "with stream as s:" form, silently emitting zero events

### LINKED ISSUE #336: bug: combining global and per-client snagline.auto instrumentation double-counts every call (two events per request)

Adhi1-2:
Fixes #335
Fixes #336

Two independent defects in the `snagline.auto` wrappers. Both failed silently — wrong telemetry with no warning — rather than failing loudly.

## #335 — the documented streaming form emitted nothing

Both SDKs document streaming as:

```python
with client.chat.completions.create(model="…", messages=[…], stream=True) as stream:
    for chunk in stream:
        …
```

The wrappers only proxied the raw stream through `__getattr__`. Implicit special-method lookup for `with` / `async with` resolves on the **type** and never consults `__getattr__`, so the protocol was never found:

```
Form 1 (for chunk in stream): events = 1
Form 2 (with stream as s):     TypeError: '_SyncStreamWrapper' object does not
                              support the context manager protocol -> events = 0
```

That is the exact "monitored process that is not monitored at all" failure #270 fixed for global mode, and the primary documented streaming idiom for both SDKs, so a user following the docs got a hard `TypeError` at best and no telemetry at worst.

`_SyncStreamWrapper` / `_AsyncStreamWrapper` in **both** `auto/openai.py` and `auto/anthropic.py` now define `__enter__`/`__exit__` and `__aenter__`/`__aexit__`. `__enter__` returns the wrapper itself, so iteration still flows through the wrapper and the deferred emission from #242 still fires once — at exhaustion, close, or iteration error — never at `__enter__` time. `__exit__` delegates to the existing `close()` / `aclose()`, so the underlying stream is still released and an exception raised inside the block still propagates.

## #336 — global + per-client composition double-counted every call

Global mode (`_patch_resource_classes`) already skips an already-wrapped `create` via its `__snagline_wrapped__` marker. `wrap_client` (the per-client path) had no guard and never marked its own output, so:

```python
instrument_openai(monitor)                 # global, patches the resource class
instrument_openai(monitor, client=client)  # resolves the *wrapped* class attr and stacks a second layer
```

…or simply calling `wrap_client` twice, emitted one event per layer per call:

```
Case B (per-client x2): after first call = 1, after second call total = 3
                        (expected +1, got +2)
```

Every counting detector is biased by this: `error_cascade` trips on roughly two real failures, `loop`/`stagnation` repetition counts double, and every latency in the CUSUM window appears twice — silently miscalibrated detection with no warning, for the very natural pattern of instrumenting globally in a startup hook and wrapping a specific client at a call site.

Fixed across all three modules:

- `_wrap_one` now **marks its own output** and **returns an already-wrapped callable verbatim** instead of layering. `auto/langchain.py` never set the sentinel at all (the odd one out the issue calls out), so its wrappers were invisible to every guard.
- `wrap_client` skips already-wrapped methods, and a second call that finds everything wrapped stays **quiet** instead of warning "found no create method" — the same "fully instrumented, not a failure" distinction `_patch_resource_classes` makes by returning `-1`.

Fixing #336 also exposed a latent flaw in the test fakes: the clients in `tests/test_auto_openai.py` / `tests/test_auto_anthropic.py` held their resources as **class attributes**, so `wrap_client` mutated objects shared across tests. One cross-test assertion was satisfied by the double-wrap bug itself; both fakes now build per-instance resources.

## Verification

- **Mutation-checked the new tests**: with the source fix reverted and the tests kept, **17 fail** — 9 context-manager cases (sync `with` and async `async with`, both SDKs, including block-exits-early and block-raises paths) and 8 idempotency cases (per-client ×2, global-then-per-client, LangChain sync and async, the marker itself). Restored and re-ran green, so every test is load-bearing.
- Full suite: **870 passed, 3 skipped**, 89.06% coverage (the venv here has no OpenAI/Anthropic SDK installed, matching CI's `langchain-core`-only install, so all five reproduction cases run against local fakes shaped like the real SDKs: `cached_property` client surface and `@required_args`-decorated `create` carrying `__wrapped__`).
- `ruff format --check src tests` and `ruff check src tests` clean; `mypy src tests` clean (140 files).
- CHANGELOG `[Unreleased]` and `docs/ATTACH_ANY_SYSTEM.md` updated.


Cyrax321 (approved):
## Review — PR #344: support stream context managers, make `wrap_client` idempotent (#335, #336)

Checked out `pull/344/head` and reviewed against `origin/master`. 10 files, +566/-14, two commits.

### Verification performed
- ✅ **Full suite: 873 passed, 3 skipped** (vs 849 on master) — no regressions
- ✅ The 40 new tests across `test_auto_stream.py`, `test_auto_wrap_idempotent.py`, `test_auto_global_mode.py` — all pass
- ✅ **Bug #335 reproduced on master** (both forms): `with stream as s:` → `TypeError: does not support the context manager protocol (missed __exit__ method)`; `async with` → same for `__aexit__`. **Zero events emitted.** On the branch both forms work and emit exactly 1 event.
- ✅ **Bug #336 reproduced on master**: `wrap_client` called twice, then one `create()` call → **2 events** (double-counted). On the branch: **1 event**.
- ✅ **Mutation-tested the fix**: reverted all three `auto/*.py` modules to master and re-ran the new tests — **20 fail, 8 pass** (the description said 17; the actual count is higher, so it's conservative). Restored, all green. Every test is load-bearing.
- ✅ **Body-exception path verified by hand** (commit `d3cfcb7`): a `RuntimeError` raised mid-iteration inside a `with` block records `error=True, error_type='RuntimeError'`; a clean early `break` records success. Correct in both directions.

### What's done well

- **The root cause of #335 is diagnosed exactly right.** Implicit special-method lookup for `with`/`async with` resolves on the *type*, never through `__getattr__` — so proxying `__enter__`/`__exit__` via `__getattr__` was structurally unreachable, not merely missing. The comment states this explicitly, which is what makes the fix trustworthy rather than cargo-culted. This is the same "monitored call that isn't monitored at all" class #270 fixed for global mode, and it hit the *primary documented streaming idiom for both SDKs* — so users following the docs got a hard `TypeError` at best and silent zero-telemetry at worst.
- **`__enter__` returns `self`, not the underlying stream.** This is the subtle correctness point: returning the raw stream would bypass the wrapper's `__next__` and lose the deferred emission. Returning self keeps iteration flowing through the wrapper so the event still fires exactly once at exhaustion.
- **`__exit__` propagates exceptions** (`return False`) and *records* them. A caller's block dying mid-stream is the call's visible outcome; without this, `close()` alone would have logged a clean success for a call that actually failed. The `_emit` guard means this is a no-op when the stream already emitted at exhaustion — correctly idempotent.
- **#336's impact analysis is the important part.** Double-counting isn't a cosmetic inaccuracy — it silently miscalibrates *every* counting detector: `error_cascade` trips on ~2 real failures, loop/stagnation repetition counts double, and every latency lands in the CUSUM window twice. And it triggers on the very natural pattern of instrumenting globally in a startup hook and wrapping a specific client at a call site. Fixing the guard is small; understanding why it matters is the contribution.
- **The LangChain gap was real and is called out.** `auto/langchain.py`'s `_wrap_one` never set the sentinel at all, so its wrappers were invisible to *every* guard — fixed as the odd one out.
- **The "already fully wrapped is not a failure" distinction is preserved.** A second `wrap_client` that finds everything wrapped now stays quiet instead of warning "found no create method" — mirroring `_patch_resource_classes` returning `-1` for the same reason. Without this the fix would spam warnings on the legitimate composition pattern.
- **The test-fake fix is a genuine latent bug.** The clients held resources as *class attributes*, so `wrap_client` mutated objects shared across tests — and one cross-test assertion was being satisfied by the double-wrap bug itself. Moving to per-instance resources removed a hidden test-ordering dependency that was masking the very thing being tested.

### Findings

**1. Conflict with PR #329 (approved) — same files, different fixes** — *needs merge-order attention*

PR #329 also modifies `src/snagline/auto/openai.py`, `auto/anthropic.py`, and `auto/langchain.py` — for #321 (the `instrument_*` returns-True contract). Both PRs restructure `_wrap_one` / `wrap_client` in overlapping regions: #329 extracts `_patch_client` from `wrap_client`, #344 adds the `__snagline_wrapped__` guard inside both `_wrap_one` and `wrap_client`. I simulated the merge and it **conflicts** on `openai.py`.

The good news: the fixes are *compatible in intent*. #329's `_patch_client` counts patched methods; #344's guard makes some of those methods already-wrapped (counted as `already` instead). If #329 lands first, #344's `_patch_client` will need to treat an already-wrapped method as "patched for contract purposes" — otherwise `instrument_openai(client=c)` could return `False` on a client that global mode already covered, which would be a *different* silent contract violation. **Recommend merging #344 first, then rebasing #329 so its count includes the already-wrapped case.**

**2. `_emit` is a no-op when already emitted — worth one assert in the comment** — *nit*

The `__exit__` comment says "_emit is guarded, so when the stream already emitted at exhaustion this is a no-op." That's accurate and I verified it (a post-exhaustion body exception correctly records the earlier clean emission, not a spurious error). But the *invariant* — "the first emission wins, and a later `__exit__` cannot overwrite it" — is load-bearing for the double-count fix's own contract. A brief note that this ordering is deliberate (exhaustion's clean record is authoritative once made) would help the next reader avoid "fixing" it into a bug.

**3. Mutation count in the description is conservative** — *informational*

The description reports 17 failing tests when the source fix is reverted; I measured **20** (9 context-manager + 8 idempotency + the 3 body-exception cases in commit `d3cfcb7`, which the description predates). Understating the guard strength is harmless, but the final number is 20, not 17.

### Verdict

✅ **Approve.** Both bugs are of the worst kind for a monitoring library — silently wrong telemetry with no warning — and I reproduced each on master before accepting the fix. The context-manager gap in particular hit the documented streaming idiom for both SDKs. All 20 new tests are mutation-verified load-bearing, and the body-exception edge case is handled correctly in both directions. The one action item is merge ordering against #329 (finding 1) — merge this one first.

Cyrax321 (commented):
**Code Review — PR #344**

The context-manager delegation is correct (returns `self`, `__exit__` returns `False` so block errors propagate, `close`/`aclose` idempotent), and the `__snagline_wrapped__` guard is present in all three adapters. But the guard's silent-skip semantics create a real regression for the multi-monitor case, and the two other high/medium findings are about integration with the merged #321 refactor and #345.

Findings:

**1. `high` — the new idempotency guard makes a second `wrap_client`/`instrument_*` call with a *different* monitor a silent no-op.**
`src/snagline/auto/openai.py:404-414`, `src/snagline/auto/langchain.py:94`
Verified against the PR-head source: `wrap_client(A, c); wrap_client(B, c); invoke()` gives `A events: 1, B events: 0` (same for langchain `m1:1, m2:0`). Every later call is attributed only to the first monitor, with no warning, while `instrument_openai(monB, client=client)` still returns True so the host believes B is watching. Before this PR the second wrap layered and events reached B (doubled at worst) — so the composition case regresses from "noisy" to "silently unmonitored". Only the same-monitor path is tested. The guard should either re-point the wrapper at the new monitor or warn when the attached monitor differs.

**2. `medium` — the LangChain sentinel is dropped if #345 lands first (conflicting hunks).**
`src/snagline/auto/langchain.py:52,94`
PR #345 rewrites `_wrap_one`/`wrap_client` in the same file with no marker and no guard (`git merge-tree` reports `CONFLICT in src/snagline/auto/langchain.py`), so this PR's fix for the "odd one out" module is lost on merge and double-counting returns. Coordinate the two PRs; whichever lands second must re-apply the guard.

**3. `medium` — the branch is based on the stale `b667169`, predating master's `_patch_client` refactor (#321-#327).**
`src/snagline/auto/openai.py:378-389`
Master already has `_patch_client(monitor, client) -> int` with `return _patch_client(...) != 0`, while this head still does `wrap_client(monitor, client); return True` with the new `already` counter spliced in. Rebase and re-integrate the quiet-already-instrumented logic into `_patch_client`, or the documented boolean contract regresses (`gh pr view 344` → `CONFLICTING`).

**4. `medium` — `__exit__` records *any* exception escaping the `with` body as the LLM call's failure.**
`src/snagline/auto/openai.py:163`
`if exc_info[1] is not None: self._emit(error=True, error_type=...)`. A post-processing bug the caller raises while parsing chunks marks a successful API call as `error=True`, feeding phantom failures into `error_cascade` (two real failures are its threshold). The added test asserts a bare `RuntimeError` becomes the call's error, so the behavior is intended — but it should distinguish errors surfaced through the proxied stream methods from arbitrary caller-block exceptions.

**5. `low` — `__enter__`/`__aenter__` return `self` without delegating to the wrapped stream's own entry.**
`src/snagline/auto/openai.py:160` — `_SyncStreamWrapper.__init__` already does `iter(stream)`/`__aiter__()` before the protocol is entered, so an SDK stream whose `__enter__` performs setup or returns a distinct iterator is bypassed. The test fakes have trivial entry, so this is unexercised.

**6. `low` — the four context-manager dunders and near-identical `_emit`/`close` bodies are duplicated verbatim between `openai.py` and `anthropic.py`; a shared mixin would prevent divergence.**

**7. `low` — warning behavior for the "everything already wrapped" state is inconsistent: only openai tracks `already` to suppress the no-op warning; anthropic and langchain report nothing.**

PR #359, opened 2026-09-20
fix(monitor): apply the Config-borne enforcement knobs in the constructor

### LINKED ISSUE #352: bug: Monitor.__init__ silently ignores Config.policy / halt_url / fail_open, which Monitor.default() does apply
## Description

`Monitor.default()` applies five enforcement knobs from the resolved `Config` — `policy`, `halt_url`, `halt_timeout_s`, `min_severity_for_halt`, `fail_open` — but the direct constructor reads only its own positional/keyword arguments, which default to `observe` / `None` / `0.25` / `0.8` / `True`:

```python
# src/snagline/monitor.py:234
cfg = config or Config()
self._max_wall_seconds = cfg.max_episode_wall_seconds
...
self._configure_policy(policy=policy, on_risk=on_risk, halt_url=halt_url, ...)
```

An operator wiring a `Config` into the library API — the documented way to pin knobs — and setting `SNAGLINE_POLICY=halt_webhook` / `SNAGLINE_HALT_URL` gets observation mode with no warning. The identical config through `snagline serve` arms the webhook, so the two entry points disagree.

## Current Behavior

```python

cfg = Config(policy="halt_webhook", halt_url="http://example.local/halt", fail_open=False)
m = Monitor([], [], config=cfg)
m.policy        # 'observe'   (was configured 'halt_webhook')
m.halt_url      # None        (was configured 'http://example.local/halt')
m._fail_open    # True        (was configured False)
```

`Monitor.default(config=cfg)` with the same `cfg` produces `policy == 'halt_webhook'` and the halt URL wired.

## Expected Behavior

Either fold the enforcement knobs from `config` into `__init__` when the constructor arguments are left at their defaults (so an explicit constructor argument still wins, preserving backwards compatibility), or document that these five are only honored through `Monitor.default()` and raise/warn when a `Config` carries a value the constructor would drop.

## Context

Verified against master at `b667169`. Related to #331/#333 in shape (a knob silently not applied) but a different module and a different failure surface.

Adhi1-2:
Fixes #352.

## what was wrong

`Monitor.default()` applies five enforcement knobs from the resolved `Config` — `policy`, `halt_url`, `halt_timeout_s`, `min_severity_for_halt`, `fail_open` — but the direct constructor read only its own arguments, which default to `observe` / `None` / `0.25` / `0.8` / `True` (`src/snagline/monitor.py:214`).

So the two entry points disagreed about one configuration:

```python
cfg = Config(policy="halt_webhook", halt_url="http://example.local/halt", fail_open=False)
Monitor([], [], config=cfg).policy   # "observe"  — silently unmonitored enforcement
Monitor.default(config=cfg).policy   # "halt_webhook"
```

An operator who set `SNAGLINE_POLICY=halt_webhook` / `SNAGLINE_HALT_URL` and wired the `Config` into the library API got observation mode with **no warning anywhere**; the same config through `snagline serve` armed the webhook.

## the fix

Each of the five knobs resolves against the `Config` when the constructor argument is `None` ("not given"); an explicitly passed value still wins:

```python
if policy is None:
    policy = cfg.policy
# ... halt_url, halt_timeout_s, min_severity_for_halt, fail_open
```

The arguments changed from concrete defaults to `None` sentinels rather than comparing against the old literal defaults, because a caller passing the literal default (`fail_open=True`, `policy="observe"`) is indistinguishable from an omitted one under value comparison — the sentinel makes "explicit wins" exact for every knob.

Backwards compatible by construction: `Config`'s defaults for these five fields are the same literals the arguments used to carry, so `Monitor(detectors, sinks)` still resolves to `observe` / `None` / `0.25` / `0.8` / `True`. The config path reaches the same `_configure_policy` validation, so `Config(policy="halt_webhook")` with no `halt_url` now raises `ValueError` at construction instead of silently doing nothing at ingest time.

`on_risk` is deliberately not on `Config` (callables cannot arrive via env or config files), so callback mode through the constructor still needs the argument — `_configure_policy` already warns about that case.

## validation

- 5 new tests in `tests/test_enforcement_policy.py`: knobs applied from config, defaults unchanged without a config, explicit argument wins over config, the two entry points agree on one config, and the config path still validates (raises on `halt_webhook` without `halt_url`).
- Negative control: 3 of the 5 fail on clean `master` before the fix.
- Full suite: **854 passed, 3 skipped**, coverage **88.01%** (gate 80%).
- `ruff check src tests`, `ruff format --check src tests`, `mypy src tests` clean.


Cyrax321 (approved):
## Review — PR #359: apply the Config-borne enforcement knobs in the constructor (#352)

Checked out `pull/359/head` and reviewed against `origin/master`. 3 files, +131/-5, single commit.

### Verification performed
- ✅ **Full suite: 856 passed, 2 skipped** — no regressions
- ✅ `tests/test_enforcement_policy.py`: 40 passed (5 new)
- ✅ `ruff check`, `ruff format --check`, `mypy src tests` — **all clean** (139 files)
- ✅ **Bug #352 reproduced on master**: `Monitor([], [], config=cfg).policy` → `'observe'` while `Monitor.default(config=cfg).policy` → `'halt_webhook'` for the same config. The two documented entry points disagree, silently.
- ✅ **Branch fixes it**: both entry points now return `'halt_webhook'`; `Monitor(..., policy="observe")` still returns `'observe'` (explicit wins); and `Monitor([], [])` with no config still resolves to `observe` / `0.25` / `True`.
- ✅ **Mutation-tested**: reverted `monitor.py` to master → **exactly 3 of the 5 new tests fail** (matching the description's "3 of 5" negative control). Restored, all green.

### What's done well

- **The sentinel design is the correct solution, not the obvious one.** Changing the argument defaults to `None` and resolving against `cfg` when `None` — rather than comparing against the old literal defaults — is what makes "explicit wins" *exact*. Under value comparison, a caller deliberately passing `fail_open=True` or `policy="observe"` is indistinguishable from an omitted argument, so the literal-default approach would have introduced a new precedence bug while fixing the old one. The comment in the diff explains this.
- **Backwards compatibility is by construction, not by testing.** Because `Config`'s defaults for these five fields are the same literals the arguments used to carry, `Monitor(detectors, sinks)` resolves to exactly `observe` / `None` / `0.25` / `0.8` / `True` as before. No existing call site changes behavior — I verified this rather than taking it on faith.
- **The fix reaches the existing validation path.** Because the config-borne values now flow through `_configure_policy`, `Config(policy="halt_webhook")` with no `halt_url` raises `ValueError` at *construction* instead of silently doing nothing at ingest time. That's a strictly better failure mode (fail at startup, not mid-run).
- **`on_risk` is deliberately excluded, with a reason** — callables cannot arrive via env vars or config files, so it correctly stays a constructor-only argument. Stating the exclusion's rationale prevents a future "consistency" PR from adding it back wrongly.

### Findings

**1. The constructor uses `Config()`, not `Config.resolve()` — env knobs are NOT honoured unless the caller resolves** — *worth documenting, arguably by design*

```python
cfg = config or Config()          # line 241 — constructor
...
cfg = config or Config.resolve()  # line 1119 — Monitor.default()
```

So `SNAGLINE_POLICY=halt_webhook` with `Monitor(detectors, sinks)` (no config) still yields `observe`. The issue's reproduction is fixed for callers who pass a resolved `Config` — which is what `snagline serve` does and what the docs describe — but a host constructing `Monitor` bare and expecting env-layered knobs gets nothing.

This is defensible: `Config.resolve()` on every construction would make the constructor's behavior depend on process environment, which is a surprising default for a library API, and `Monitor.default()` is the documented env-aware entry point. But the docstring says the constructor honors `SNAGLINE_POLICY` "exactly like `Monitor.default` does," which is only true when a resolved config is passed. **Recommend tightening that sentence** to "when a resolved `Config` is passed; `Monitor.default()` is the env-aware entry point" so the asymmetry is stated rather than discovered. Not blocking — the current behavior is the safer default.

**2. Five near-identical `if x is None: x = cfg.x` blocks** — *style nit*

The resolution is correct and readable, but it's the same pattern five times. A two-line helper (`_resolved(arg, cfg, field)` or a small loop over the field names) would state the invariant once. The explicit form is arguably clearer at this size, so this is a genuine judgment call, not a defect — flagging only because the same five-knob resolution will need updating in step if a sixth enforcement knob is ever added to `Config`.

### Verdict

✅ **Approve.** The bug is real and reproduced verbatim — an operator who set `SNAGLINE_POLICY=halt_webhook` and wired the config into the library API got observation mode with no warning anywhere, while the same config through `snagline serve` armed the webhook. The fix is minimal, exactly backwards compatible by construction, and the "explicit wins" precedence is made precise via sentinels rather than value comparison. The only follow-up is a one-sentence docstring clarification about the `Config()` vs `Config.resolve()` asymmetry (finding 1).

Cyrax321 (commented):
**Code Review — PR #359**

The fix addresses #352: `Monitor.__init__` now resolves five Config-borne enforcement knobs (`policy`, `halt_url`, `halt_timeout_s`, `min_severity_for_halt`, `fail_open`) when its arguments are left as `None` sentinels, so the direct constructor and `Monitor.default()` agree on a passed `Config`. The sentinel design makes explicit-argument precedence exact and is backwards compatible for stock configs. The remaining issues are the no-config asymmetry and per-knob resolution semantics.

Findings:

**1. `medium` — the no-config path is still asymmetric: the constructor uses `Config()` while `Monitor.default()` uses `Config.resolve()`.**
`src/snagline/monitor.py:241` (vs `1119`)
Verified: head line 241 is `cfg = config or Config()`, line 1119 is `cfg = config or Config.resolve()`. The new resolution block only reads the local `cfg`, so env layering is reached only when the caller passed an already-resolved Config. An operator who reads the new docstring and sets `SNAGLINE_POLICY=halt_webhook` + `SNAGLINE_HALT_URL` but calls `Monitor(detectors, sinks)` bare gets policy=observe with no warning, while `Monitor.default()` in the same process would arm the webhook — the silent-observation failure #352 describes, just on the no-config path. Use `Config.resolve()` in both places (or document the difference).

**2. `low` — `fail_open=None` silently flips meaning from fail-closed to fail-open.**
`src/snagline/monitor.py:290-292`
Previously `fail_open: bool = True` was stored verbatim and `if not self._fail_open:` meant `None` raised on a detector/sink exception. Now `None` is a valid "not given" sentinel that resolves to `Config().fail_open` (True), so an out-of-contract caller forwarding a possibly-None variable silently gets fail-open. Acceptable as a deliberate sentinel, but it is a behaviour change worth a docstring note.

**3. `low` — per-knob resolution lets an explicit non-halting policy inherit Config's `halt_url`.**
`src/snagline/monitor.py:284-289,355-357`
Each knob resolves independently, so `Monitor([], [], policy="observe", config=Config(policy="halt_webhook", halt_url="file:///etc/passwd"))` yields `m.policy == "observe"` yet `m.halt_url == "file:///etc/passwd"`, violating the property's documented "or None when not in halt_webhook mode" and skipping the http(s) scheme check (which `_configure_policy` only runs under `halt_webhook`). Resolve `halt_url`/`halt_timeout_s`/`min_severity_for_halt` only when the resolved policy is `halt_webhook`.

**4. `low` — a Config with `policy=halt_webhook` but no `halt_url` now raises at construction.**
`src/snagline/monitor.py:282-299`
On master the constructor never read `config`, so such a config built an observe monitor; now it raises `ValueError("policy='halt_webhook' requires halt_url")`. The PR body calls this intentional fail-fast, but it is an observable change for hosts that store `policy` centrally and supply the endpoint out-of-band.

**5. `low` — the config→enforcement mapping is now duplicated between `__init__` and `Monitor.default()`.**
`src/snagline/monitor.py:282-299` vs `1217-1233`
The two sites independently map the same five knobs; a sixth enforcement knob added to Config and wired into one but not the other reintroduces the exact divergence this PR fixes. A shared resolver helper would state the invariant once.

PR #360, opened 2026-09-20
fix(sinks): reject a non-positive flush_interval in BatchingSink

### LINKED ISSUE #358: bug: BatchingSink with a non-positive flush_interval busy-spins at ~780k flushes/sec instead of pacing delivery
## Description

`BatchingSink` clamps `max_batch` (`max(1, max_batch)`) but not `flush_interval`, and a negative or zero interval makes `_wake.wait(...)` return immediately, turning the flusher into an unthrottled busy loop:

```python
# src/snagline/sinks/batching.py:43
self._flush_interval = flush_interval
...
# :74
self._wake.wait(self._flush_interval)
```

With `flush_interval=-1.0` the thread spins through `_flush()` roughly **780,000 times per second**, pinning a full core for the life of the process. Alerts are still delivered, so nothing else looks wrong — the only symptom is a hot thread.

`close()` compounds it: `self._thread.join(timeout=self._flush_interval + 1.0)` uses the interval as the join timeout, so a negative interval gives the drain essentially no time and falls back to the caller-thread flush.

## Current Behavior

```python

class Cap:
    n = 0
    def emit(self, risk): Cap.n += 1

s = BatchingSink(Cap(), max_batch=1000, flush_interval=-1.0)
s.emit(object())
# 0.5 s later: 388,793 flush passes, i.e. ~780k/s, on an idle queue
```

## Expected Behavior

Clamp like `max_batch` already does — `max(self._flush_interval, some small positive floor)`, or reject a non-positive interval at construction with a clear `ValueError`, since a non-positive interval has no meaningful reading (unlike `max_batch`, there is no "disabled" semantics worth preserving; the sink's whole point is pacing).

## Context

Verified against master at `b667169`. The sink is public API but is not currently wired to any CLI flag, so this is reachable from library use rather than from `snagline serve`/`watch`.

Adhi1-2:
Fixes #358.

## what was wrong

`BatchingSink` clamped `max_batch` (`max(1, max_batch)`) but stored `flush_interval` verbatim, and the flusher waits on `self._wake.wait(self._flush_interval)`. A zero or negative interval makes that wait return immediately, so the flusher spins through `_flush()` on an empty queue as fast as it can:

```python
s = BatchingSink(Cap(), max_batch=1000, flush_interval=-1.0)
s.emit(object())
# ~780,000 flush passes/sec, a full core, for the life of the process
```

Alerts still delivered, so nothing else looked wrong — the only symptom was a hot thread. `close()` compounded it: `self._thread.join(timeout=self._flush_interval + 1.0)` uses the interval as the join timeout, so a negative one gave the shutdown drain essentially no time and fell back to the caller-thread flush.

## the fix

A non-positive interval raises `ValueError` at construction, naming the knob:

```python
if flush_interval <= 0.0:
    raise ValueError(f"flush_interval must be positive; got {flush_interval!r}")
```

`max_batch` stays clamped because any size still paces delivery; an interval has no equivalent reading, so this is a configuration error rather than a clamp — the same loud-at-construction direction as the knob validation in #317/#331/#333. The sink is public API but is not wired to any CLI flag, so this is reachable from library use, not from `snagline serve`/`watch`.

## validation

- 4 new tests in `tests/test_batching_sink.py`: `0.0` / `-1.0` / `-0.5` all rejected, and a small positive interval is still accepted and stored unchanged.
- Negative control: the three rejection tests fail on clean `master` before the fix — and take **13.7 s** instead of ~1 s to do it, which is the busy-spin itself burning CPU in the test process.
- Full suite: **851 passed**, coverage **87.96%** (gate 80%). The only 2 failures are pre-existing and environmental — they fail identically on clean `upstream/master` (this dev venv has `openai` installed, which the "SDK absent" test requires to be missing).


Cyrax321 (approved):
## Review — PR #360: reject a non-positive flush_interval in BatchingSink (#358)

Checked out `pull/360/head`. 3 files, +52/-0.

### Verification performed
- ✅ **Full suite: 855 passed, 2 skipped** — no regressions; `test_batching_sink.py` (4 new tests) pass
- ✅ `ruff check`, `ruff format --check`, `mypy src tests` — all clean (139 files)
- ✅ **Bug reproduced and measured on master**: `flush_interval=-1.0` spins the background flusher at **~1,084,000 `_flush()` invocations/sec** — a full core for the life of the process. (This is even worse than the description's 780k estimate.)
- ✅ Branch rejects `0.0`, `-1.0`, `-0.5` with a clear `ValueError` naming the knob; a small positive interval is still accepted unchanged.

### Assessment

The spin is real and I measured it directly by instrumenting `_flush`. Worth noting *why* it's invisible: `_flush()` returns immediately on an empty queue (line 84), so a plain `flush()`-counting probe reads zero — the hot loop is the `_wake.wait(0)` → `_wake.clear()` → empty-`_flush()` cycle itself. That subtlety is why the bug shipped: nothing else looked wrong, alerts still delivered, and the only symptom was a hot thread.

The `close()` interaction is the other half of the bug and the description gets it right: `join(timeout=self._flush_interval + 1.0)` gives a negative interval essentially no shutdown drain time, falling back to the caller-thread flush.

Rejecting at construction (rather than clamping) is the correct call — a clamp to some arbitrary positive interval would silently change the user's intended pacing, while `max_batch` genuinely has a "any size still paces delivery" reading. This matches the loud-at-construction direction already established by #317/#331/#333.

### Verdict

✅ **Approve.** Small, correct, and I verified the spin empirically rather than taking the description's number. The description's 780k figure is conservative; the real cost is ~1.08M passes/sec.

Cyrax321 (commented):
**Code Review — PR #360**

The construction-time `ValueError` for a non-positive `flush_interval` correctly kills the 0/negative busy-spin and matches the repo's loud-at-construction convention, with docstrings, CHANGELOG and four tests. But the predicate is comparison-only, so NaN and +inf slip through and re-create failure modes worse than the one fixed.

Findings:

**1. `high` — NaN passes `flush_interval <= 0.0` and re-creates the busy-spin, then makes `close()` raise.**
`src/snagline/sinks/batching.py:51`
Verified: `float('nan') <= 0.0` is False, so `flush_interval=float('nan')` is stored (line 60). In the flusher, `threading.Event.wait(nan)` returns False almost immediately, so `_run` loops on an empty queue at hundreds of thousands of passes/second, pinning a core for the process lifetime — the exact #358 symptom, now unguarded. On `close()`, `self._thread.join(timeout=nan)` raises `ValueError: Invalid value NaN` (verified on py3.10/3.11/3.13) when the thread is still alive, escaping close and skipping the caller-thread drain. Tests cover only `0.0/-1.0/-0.5`, so the suite never rejects a value the body calls "non-positive".

**2. `medium` — +inf passes, then `_wake.wait(inf)` raises `OverflowError` in the flusher and `close()` raises too, leaking alerts.**
`src/snagline/sinks/batching.py:51`
`inf <= 0.0` is False, so `flush_interval=float('inf')` is accepted. The flusher's first `self._wake.wait(inf)` raises `OverflowError: timestamp out of range` (verified), killing `_run` and silently disabling background delivery; risks emitted afterwards sit in the queue until someone calls `flush_now()`. `close()` then calls `join(timeout=inf)`, which raises OverflowError and aborts before its own `_flush()`, so queued alerts are dropped and the exception propagates into the host. This is *silently dead* rather than loud — the opposite of the PR's rationale.

**3. `low` — the guard's wording promises more than it validates; the repo's float-knob idiom is `math.isfinite(...) or value <= 0`.**
`src/snagline/sinks/batching.py:51,59`
The docstring, CHANGELOG and error text say "non-positive", but the predicate only rejects `<= 0.0`. Sibling precedent is `if not math.isfinite(value) or value <= 0:` (`http_server.py:190` for `server_read_timeout`, `234` for `episode_ttl_seconds`) and the `_validated_*` functions in `config.py`. Using `not math.isfinite(flush_interval) or flush_interval <= 0.0` closes both findings and matches the established idiom.

Issue #349, opened 2026-09-20
bug: NaN or Inf tokens_in/tokens_out crashes the token-runaway detector, which fail-open then leaves dead for the run


abyyxhek:
## Description

`TokenRunawayDetector.observe` converts token counts with a bare `int(...)`, so a non-finite value from an adapter raises straight out of `observe`:

```python
# src/snagline/detectors/token_runaway.py:82
step_tokens = int((event.tokens_in or 0) + (event.tokens_out or 0))
```

- `tokens_in=NaN` → `ValueError: cannot convert float NaN to integer`
- `tokens_out=Inf` → `OverflowError: cannot convert float infinity to integer`

Both are swallowed by `Monitor.ingest`'s fail-open guard, so the detector stays installed and reports nothing for the rest of the run — the same "silently dead detector" shape the config validators were written to eliminate, but reachable at runtime rather than from a bad knob. `tokens_in=-50` is also silently accepted and *reduces* the running budget total.

Adapters derive these from model `usage` objects, and a provider returning a malformed usage blob is a realistic failure, not a synthetic one.

## Current Behavior

```python
from snagline.detectors.token_runaway import TokenRunawayDetector
from snagline.events import StepEvent, make_signature

def ev(tin=None, tout=None):
    return StepEvent("0", "ep", 1.0, "tool_call",
                     make_signature("tool_call", "t", "a"), "t",
                     tokens_in=tin, tokens_out=tout)

TokenRunawayDetector(budget_total_tokens=1000).observe(ev(tin=float("nan")))
# ValueError: cannot convert float NaN to integer
TokenRunawayDetector(budget_total_tokens=1000).observe(ev(tout=float("inf")))
# OverflowError: cannot convert float infinity to integer
```

Through `Monitor.default(config=Config(token_runaway_enabled=True))`, one such event leaves `detector_errors == 1` and every later event unmonitored, with a single startup log line as the only symptom.

## Expected Behavior

Coerce defensively and fail open per step: treat non-finite or negative counts as "no usable token signal for this step" (skip it, or clamp to 0), rather than letting an arithmetic error take the detector down for the run.

## Context

Verified against master at `b667169`.

Issue #296, opened 2026-09-18
[tests] test_real_actionledger_flags_review crashes against current CONTINUUM: MiniStorage.append_event rejects the source= kwarg


Adhi1-2:
## Description

`tests/sinks/test_continuum_sink.py::test_real_actionledger_flags_review` is the one live check that exercises `ContinuumSink` against the real `continuum` package rather than a fake. It is skipped when `continuum` is not importable, so it is invisible in CI (CI never installs CONTINUUM) — but with a current checkout of CONTINUUM installed it fails on import of the real API:

```text
E       TypeError: test_real_actionledger_flags_review.<locals>.MiniStorage.append_event()
            got an unexpected keyword argument 'source'
../Continuum/src/continuum/actions/ledger.py:775: TypeError
```

The test's in-test `MiniStorage` double was written against an older `Storage.append_event` contract:

```python
def append_event(
    self,
    run_id: str,
    type: Any = None,
    payload: Any = None,
) -> Any:
    ...
```

The real signature has since grown keyword-only parameters, and `ActionLedger` passes one of them positionally-by-keyword:

```python
>>> from continuum.storage.base import Storage
>>> inspect.signature(Storage.append_event)
(self, run_id: 'str', type: 'EventType', payload: 'Mapping[str, Any] | None' = None,
 *, causer_event_id: 'str | None' = None, expected_sequence: 'int | None' = None,
 source: 'Origin' = <Origin.DETERMINISTIC: 'deterministic'>) -> 'Event'
```

The shipped sink code is unaffected — `ContinuumSink` calls the ledger, not `Storage` directly — so this is a stale test double, not a product bug.

## Current Behavior

```shell
$ python3 -c "import continuum"   # CONTINUUM installed
$ PYTHONPATH=src python3 -m pytest tests/sinks/test_continuum_sink.py::test_real_actionledger_flags_review -q
FAILED
```

Any other test that imports the module still passes; only this one live test breaks.

## Expected Behavior

The `MiniStorage` double should accept the current keyword-only parameters so the live path can be exercised again, e.g.:

```python
def append_event(
    self,
    run_id: str,
    type: Any = None,
    payload: Any = None,
    *,
    causer_event_id: Any = None,
    expected_sequence: Any = None,
    source: Any = None,
) -> Any:
```

(or accept and ignore `**kwargs`, if the intent is for the double to keep tracking only what the test asserts on).

[...]

Issue #101, opened 2026-08-25
[detector/config] Auto-threshold calibration from BaselineProfile


Cyrax321:
## Context
P2 item 8 in ATTACH_ANY_SYSTEM. Detector thresholds are hand-tuned constants; a fitted `BaselineProfile` already knows each tool's healthy error rate and latency distribution.

## Tasks
- [ ] Opt-in `calibration=auto` config: derive loop/error-cascade window rates and CUSUM drift thresholds from BaselineProfile quantiles (error rate p99, latency mean+k·sigma)
- [ ] Keep current defaults as fallback when no baseline is loaded (never worse than today)
- [ ] Validate with benchmarks/detection_accuracy.py (#82): auto must not reduce F1 vs hand-tuned on fixtures

## Depends on
- #82 (harness is the validation gate)

## Related
- #80 (ESN ensemble consumes same baselines)

Cyrax321:
PR is up: #134 (feat/auto-calibration-101, stacked on feat/horizon-detectors since the #82 harness is the validation gate). Gate result: auto macro-F1 1.000 equals hand-tuned 1.000 on the fixture corpus with all 32 healthy controls silent in both modes; unit suite proves the tighten-only gains (consecutive cascade threshold drops to 2 on spotless baselines, seeded CUSUM detects spikes below the old warm-up floor) and the fail-open fallbacks.

PR #361, opened 2026-09-20
fix(latency): seed the calibrated start on latency_count, not count (Fixes #348)

### LINKED ISSUE #348: bug: calibrated latency baseline is seeded from a profile that carries no latency samples, paging critical on step 0
## Description

`LatencyAnomalyDetector`'s calibrated start seeds the CUSUM baseline from a `BaselineProfile` entry, gating on `ToolBaseline.count` — the count of *every* tool call — while the values it reads (`mean_latency` / `std_latency`) are computed from `latency_count` only:

```python
# src/snagline/detectors/latency_anomaly.py:216-221
candidate = self._baseline.tools.get(key[1])
if candidate is not None and candidate.count >= self.min_samples:
    seeded = candidate
...
state.seed(seeded.mean_latency, seeded.std_latency)
```

`ToolBaseline` keeps the two counts separate deliberately — `baseline.py:50-58` returns `mean_latency = 0.0` when `latency_count == 0`, and the class exists so "error rates be measured even on streams whose steps report no timing" (the auto-calibration contract, issue #101). A healthy profile built from a trajectory whose adapter reports no `latency_ms` legitimately has `count=100, latency_count=0`.

The detector then freezes onto `mean=0.0, std=0.0`. `seed()` floors sigma0 to 1.0, so the first live tool call is scored as a ~`(latency - 0)`-sigma deviation and pages at **critical on step 0 of every episode**.

## Current Behavior

```python
from snagline.baseline import BaselineProfile, ToolBaseline

p = BaselineProfile()
tb = ToolBaseline("search")
for _ in range(100):
    tb.add(None, error=False)          # count grows, latency_count stays 0
p.tools["search"] = tb
# tb.count == 100, tb.latency_count == 0, mean_latency == 0.0, std_latency == 0.0

det = LatencyAnomalyDetector(baseline=p, min_samples=5)
det.observe(StepEvent("0", "ep", 1000.0, "tool_call",
                      make_signature("tool_call", "search", "q"),
                      "search", latency_ms=120.0))
# FailureRisk(episode_id='ep', step_id='0', score=1.0, trigger='latency_anomaly',
#             detail='latency 120ms deviates from baseline (mean 0ms)',
#             severity='critical')
```

Any profile fitted from a latency-less stream — which the calibration feature explicitly supports — makes the detector unusable until it is re-fitted with timing.

## Expected Behavior

Gate on the count of *latency* samples: `candidate.latency_count >= self.min_samples`. A tool with too few timed samples falls back to today's learn-then-freeze behavior, which is correct and what the code comment already promises ("Tools without a sufficient entry keep today's learn-then-freeze behavior").

## Context

### LINKED ISSUE #350: bug: one NaN latency_ms permanently disables a (episode, tool) CUSUM baseline with no error or log line
## Description

A single non-finite `latency_ms` poisons a `(episode, tool)` CUSUM baseline irreversibly. During warm-up `learn_only` propagates NaN into `self.mean`, `freeze()` then records `mu0 = NaN` and floors sigma0 to 1.0, and every later step computes

```python
```

`(x - NaN)` is `NaN`, and CPython's `max(0.0, NaN)` returns `0.0`, so `cusum > h` is `False` **forever**. The detector goes dark with no crash and no log line — unlike a raised exception, fail-open has nothing to catch. `inf` is the mirror image: `cusum` becomes `inf` and the detector alarms on every subsequent step.

A NaN latency is a realistic adapter artifact — a duration computed as `x / 0` on a zero elapsed, or a clock-skew artifact — and one malformed sample is enough to blind the detector for the whole episode.

## Current Behavior

```python

def ev(i, lat):
                     make_signature("tool_call", "t", "a"), "t", latency_ms=lat)

d = LatencyAnomalyDetector(min_samples=3, k=0.25, h=3.0)
for i in range(3):
    d.observe(ev(i, float("nan")))          # warm-up on NaN
# mu0=nan, sigma0=1.0, frozen=True

fired = [d.observe(ev(i, 999.0)) for i in range(3, 10)]
# 0 risks fired across 7 consecutive 999ms steps — a clear sustained anomaly
```

## Expected Behavior

Sanity-check samples before they enter the Welford learner: a non-finite or negative latency should be ignored for baseline accumulation (and for scoring), like a `None` latency is today. The warm-up gate `state.n >= self.min_samples` should not be satisfied by garbage either.

## Context

Verified against master at `b667169`. Sibling of #349 (token detector, same non-finite-input class) but a distinct detector and a distinct failure mode (silent blinding rather than a raised exception).

Adhi1-2:
Fixes #348.

## what was wrong

`LatencyAnomalyDetector`'s calibrated start gated on the wrong count:

```python
# src/snagline/detectors/latency_anomaly.py
if candidate is not None and candidate.count >= self.min_samples:   # count, not latency_count
    seeded = candidate
...
state.seed(seeded.mean_latency, seeded.std_latency)                  # computed from latency_count only
```

`ToolBaseline` keeps `count` (every tool call) and `latency_count` (the subset carrying `latency_ms`) separate *deliberately* — its own docstring says the split exists "so error rates be measured even on streams whose steps report no timing (auto-calibration, issue #101)". The seeded statistics, `mean_latency` / `std_latency`, are computed from `latency_count` only and return `0.0` when it is zero.

So a profile legitimately fitted from a latency-less stream has `count=100, latency_count=0, mean_latency=0.0`. Seeding from it froze the CUSUM onto a zero baseline; `sigma0` is floored to `1.0`, so the first live call scored as an unbounded deviation:

```
FailureRisk(episode_id='ep', step_id='0', score=1.0, trigger='latency_anomaly',
            detail='latency 120ms deviates from baseline (mean 0ms)',
            severity='critical')
```

Every episode paged critical on its first tool call until the profile was re-fitted with timing.

## the fix

The seeding gate now reads `candidate.latency_count`. A tool whose latency-bearing subset alone clears `min_samples` still skips warm-up exactly as before; a latency-less tool keeps today's learn-then-freeze behavior instead of freezing onto nothing.

The same defect is still live in the `ml` extra, where `EsnCusumDetector._mahalanobis_score` gated on `count` for the same reason — that is #379, kept separate because it touches an optional dependency.

## scope narrowed from the first push

This branch originally also fixed #350 (a non-finite/negative `latency_ms` scored instead of dropped). #363 opened seventeen minutes later with an identical guard for #350 and #349, so the #350 half is dropped from here to keep one fix per issue. Nothing is lost — #363's guard is the same code, and its test suite covers warm-up poisoning, mid-run NaN and the monitor-level fail-open path.

## verification

Reproduced against `master` before fixing. 4 regression tests in `tests/detectors/test_latency_anomaly.py`:

```
#348  test_latencyless_profile_does_not_alarm_on_first_step
      test_latencyless_profile_falls_back_to_warmup
      test_latency_bearing_profile_still_seeds_and_skips_warmup   (no-overcorrection guard)
      test_seeding_gate_uses_latency_count_not_count
```

Mutation check: reverting the one-line gate fails 3 of the 4 — the three that pin the defect. The fourth is the no-overcorrection guard and passes both before and after, as intended.

Local gate: `853 passed, 1 skipped, 3 failed` on the full suite. The three failures are pre-existing on pristine `master` and environmental, not regressions: `test_cli_mentions_read_timeout_next_to_max_body`, `test_real_actionledger_flags_review` (#296), `test_instrument_openai_without_sdk_is_safe_noop` (needs no openai installed). Coverage 87.97% against the 80% gate. `ruff check` / `ruff format --check` clean, `mypy src tests` at the 7 pre-existing `auto/openai.py` errors that CI never installs the package to see.

The existing calibrated-start suite in `tests/test_calibration.py` still passes unchanged — its profiles always carry latencies, so `count == latency_count` there and the new gate is equivalent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved latency anomaly calibration for tools without recorded timing data.
  * Prevented false critical alerts caused by zero-value latency baselines.
  * Ensured tools complete warm-up using live measurements before anomaly detection begins.
  * Preserved immediate calibration for profiles with enough valid latency observations.


Adhi1-2:
**Overlap note for merge ordering:** the `latency_ms` non-finite/negative hunk here overlaps with @abyyxhek's #363 (which fixes #349 and #350). Both PRs edit the same lines at the top of `LatencyAnomalyDetector.observe`, so whichever merges second will conflict textually on that one hunk — the resolution is to keep whichever version lands first and drop the other's hunk; the two treatments are functionally identical (both return early on non-finite or negative latency, #363 additionally logs at debug).

The two PRs are otherwise disjoint:

- **#361 only:** #348 — the calibrated-start gate now reads `latency_count`, not `count`. Seeding from a profile whose stream reported no timings (`count=100, latency_count=0, mean_latency=0.0`) froze onto a zero baseline and paged critical on step 0.
- **#363 only:** #349 — `TokenRunawayDetector`'s bare `int(...)` on NaN/Inf and silently-decrementing negative token counts.

Happy to rebase onto #363 if that one merges first, or to drop the #350 hunk here and let #363 own it — maintainer's call.

Adhi1-2:
Good catch — the wording was wrong, the fix was not.

`_floored_sigma` floors the reference spread at `max(std, 1 ms, 5% * |mean|)`, so a seeded zero baseline still gets a **finite** sigma of 1 ms. Seeding `mean_latency=0.0` does not produce an infinite deviation; it produces a large but finite one (`120 ms / 1 ms = 120σ` in the test's case) that crosses the CUSUM threshold on the first step and pages critical on step 0. The bug and the fix stand — the seeding gate on `latency_count` is what stops it — only the description was inaccurate.

Corrected in 7bc94c8 across the detector comment, the changelog entry, and the test comment. CI re-running.

Cyrax321 (approved):
## Review — PR #361: seed the calibrated start on latency_count, not count (#348)

Checked out `pull/361/head`. 3 files, +107/-5.

### Verification performed
- ✅ **Full suite: 855 passed, 2 skipped** — no regressions
- ✅ **Bug #348 reproduced on master via the public API**: with a `BaselineProfile` legitimately fitted from a latency-less stream (`count=100, latency_count=0`), the first live `120ms` call pages
  `score=1.0, severity='critical', "latency 120ms deviates from baseline (mean 0ms)"` **on step 0**.
- ✅ **Branch returns `None`** for the identical input — the detector falls back to learn-then-freeze instead of freezing onto a zero baseline.

### Assessment

The root cause is exactly as diagnosed: `ToolBaseline` keeps `count` and `latency_count` separate *deliberately* (its own docstring says the split exists so error rates can be measured on streams whose steps report no timing), and the seeded statistics `mean_latency`/`std_latency` are computed from the latency-bearing subset only. Gating the seed on `count` therefore seeded from a profile that had *no* latency data at all — freezing the CUSUM onto `mean=0`, after which the floored `sigma0 = 1.0` makes any real call an unbounded deviation that trips on step 0.

The one-line gate change is the right fix, and it preserves both directions correctly: a tool whose latency-bearing subset alone clears `min_samples` still skips warm-up exactly as before; a latency-less tool keeps today's behavior instead of freezing onto nothing.

The fix's own comment is unusually good — it explains *why* `sigma0`'s floor converts a poisoned baseline into a step-0 critical page rather than a crash, which is precisely what made this bug subtle. Scoping the sibling `ml`-extra defect (#379, same bug in `EsnCusumDetector._mahalanobis_score`) to its own PR because it touches an optional dependency is the right separation.

### Verdict

✅ **Approve.** Silent false-positive paging on step 0 of every episode is a bad failure mode for a latency monitor, and the fix is minimal and provably correct in both directions. Pairs with #379 in the queue — merge this first.

Cyrax321 (commented):
**Code Review — PR #361**

The one-line change is correct and well-pinned for #348: gating the calibrated start on `candidate.latency_count` instead of `count` prevents seeding a zero baseline from a latency-less profile. Reverting the gate fails exactly the three tests that pin the defect, and the latency-bearing path is unchanged. Remaining risks are pre-existing/adjacent rather than introduced.

Findings:

**1. `medium` — the gate verifies how many timed samples exist, not that the seeded statistics are usable; a non-finite profile still re-creates #350/#348.**
`src/snagline/detectors/latency_anomaly.py:232-237`
Verified by running the head code: a profile with `latency_count >= min_samples` but a NaN `mean_latency` (reachable via JSON, since `json` accepts `NaN`) seeds `state.seed(nan, ...)`; `_WelfordCUSUM.update` computes `max(0.0, cusum + (x - nan)/sigma0 - k)` = 0.0, so `cusum > h` is False forever and the detector goes dark for the run with no exception — 20 consecutive 5000 ms spikes all returned None. An all-zero latency profile (count clears the gate, mean 0.0) reproduces the original #348 symptom: `FailureRisk(score=1.0, 'latency 120ms deviates from baseline (mean 0ms)', critical)` on step 0. The PR body says the #350 half was dropped because #363 "covers" it, but #363's guard is on live `event.latency_ms` only and never runs on the seeded path. Validate `math.isfinite(mean)`/`std` (or reject a zero/non-finite mean) at the seed site.

**2. `low` — the ml-extra ESN detector retains the identical confusion, deferred to #379.**
`src/snagline/ml/esn_ensemble.py:326` — the gate is `tb.count < 2` while the moments read below are `latency_count`-derived; a latency-less profile (count=100, latency_count=0) passes the gate and saturates the Mahalanobis term. Pre-existing and explicitly deferred in the body; noting for the cross-file audit.

**3. `low` — `cusum_min_samples` has no range validation, so setting it to 0 reopens #348.**
`src/snagline/config.py:279`
`Config(cusum_min_samples=0)` is accepted (`__post_init__` covers other families but not this knob), making the new gate `latency_count >= 0` trivially true and seeding any profile entry. The `_validated_counts_and_windows`/`#337` family closes it once merged; requires deliberate misconfiguration, hence low.

**4. `low` — the `baseline` summary still prints `n={count}` with latency-derived mean/std and no timed count.**
`src/snagline/cli.py:1019,1033` — a latency-less profile displays as `n=100 mean=0.0ms std=0.0ms`, hiding the split that caused #348. Display-only; the same fix is described in #379.

PR #362, opened 2026-09-20
fix(baseline): return a copy from BaselineCollector.snapshot

### LINKED ISSUE #357: bug: BaselineCollector.snapshot() returns the live profile, so a caller mutating it corrupts what commit() persists
## Description

`BaselineCollector.snapshot()` hands back the live accumulator, not a copy:

```python
# src/snagline/baseline_store.py:204-205
def snapshot(self) -> BaselineProfile:
    return self._profile
```

The name implies a point-in-time copy; the returned object is the same one `observe()` keeps mutating. A caller that inspects or alters the result — e.g. `p.tools.clear()` while deciding whether to `commit()` — corrupts the profile a later `commit()` persists. The same holds for a concurrent `observe()`: a reader iterating `snapshot().tools` races the collector's own accumulation, with no lock on either side.

## Current Behavior

```python
from snagline.baseline_store import BaselineCollector

col = BaselineCollector()
col.observe(StepEvent("0", "ep", 1000.0, "tool_call",
                      make_signature("tool_call", "t", "a"), "t", latency_ms=10.0))
snap = col.snapshot()
snap is col._profile        # True
snap.tools.clear()
list(col._profile.tools)    # [] -- the live profile was mutated
```

Every other `snapshot`-style accessor in the codebase (`SidecarMetricsCollector.snapshot`, `Monitor.snapshot`, the detectors' `dump_state`) copies under a lock; this one is unguarded both against caller mutation and against concurrent `observe()`.

## Expected Behavior

Return a copy — at minimum `copy.deepcopy(self._profile)`, or a cheaper structural copy taken under a lock that `observe()` also takes, depending on how hot `observe()` is expected to be.

## Context

Verified against master at `b667169`. Low severity: it bites only a caller that mutates the returned object or reads it concurrently with ingestion.

Adhi1-2:
Fixes #357.

## what was wrong

`BaselineCollector.snapshot()` handed back the live accumulator:

```python
def snapshot(self) -> BaselineProfile:
    return self._profile
```

The name promises a point-in-time view; the object is the same one `observe()` keeps mutating. A caller that inspected *and altered* the result — e.g. `p.tools.clear()` while deciding whether to `commit()` — corrupted the profile a later `commit()` persisted. The same holds for a concurrent `observe()`: a reader iterating `snapshot().tools` raced the collector's own accumulation, with no lock on either side.

Every other `snapshot`-style accessor in the codebase (`SidecarMetricsCollector.snapshot`, `Monitor.snapshot`, the detectors' `dump_state`) copies under a lock; this one was unguarded in both directions. `commit()` had the same shape one layer down — it passed the live profile to the fsyncing `save()`.

## the fix

`snapshot()` returns `copy.deepcopy(self._profile)` under a lock that `observe()` and `commit()` also take:

```python
def observe(self, event) -> None:
    with self._lock:
        self._profile.add_event(event)

def snapshot(self) -> BaselineProfile:
    with self._lock:
        return copy.deepcopy(self._profile)

def commit(self, version=None):
    ...
    with self._lock:
        self._profile.fitted_at = time.time()
        to_save = copy.deepcopy(self._profile)   # copy before the slow save
    return self._store.save(to_save, ...)
```

The lock is a leaf, held only over local work — never across disk I/O — so `observe()` is not blocked by a commit. `commit()` copies *before* the save so a concurrent ingest cannot reshape the profile mid-serialization, and the object handed to `save()` is frozen.

deepcopy over a structural copy: the profile nests `ToolBaseline` objects and (optionally) an embedding list, and `snapshot()` is a periodic host call rather than a per-step one, so the allocation is not on a hot path. Correctness first.

## validation

- This class had **zero** test coverage, so `tests/test_baseline_collector.py` is new: accumulation, the copy contract, monotonic snapshots across further events, commit persisting independently of a mutated snapshot, the no-store fail-open no-op, and a 50-snapshot run against two concurrent ingest threads asserting every snapshot is a coherent non-decreasing point-in-time count.
- Negative control: 2 of the 7 fail on clean `master` before the fix.
- Full suite: **854 passed**, coverage **88.17%** (gate 80%). The only 2 failures are pre-existing and environmental — they fail identically on clean `upstream/master` (this dev venv has `openai` installed, which the "SDK absent" test requires to be missing).


Cyrax321 (approved):
## Review — PR #362: return a copy from BaselineCollector.snapshot (#357)

Checked out `pull/362/head`. 3 files, +169/-5.

### Verification performed
- ✅ **Full suite: 858 passed, 2 skipped** — no regressions
- ✅ `ruff`, `mypy` — all clean (140 files)
- ✅ **Bug #357 reproduced on master**: calling `snapshot().tools.clear()` **corrupts the live collector** — a subsequent `snapshot()` returns an empty profile, and a later `commit()` would persist the corrupted state.
- ✅ **Branch isolates it**: the same mutation has no effect on the collector.

### Assessment

The name `snapshot()` promises a point-in-time view and returned the live accumulator — a classic aliasing bug. The interesting part is the concurrency angle: the same object is what `observe()` keeps mutating, so a reader iterating `snapshot().tools` raced the collector's accumulation with no lock on either side. Every other snapshot-style accessor in the codebase (`SidecarMetricsCollector.snapshot`, `Monitor.snapshot`, the detectors' `dump_state`) copies under a lock; this one was unguarded in both directions, and `commit()` had the same shape one layer down by passing the live profile into the fsyncing `save()`.

The fix applies the established pattern precisely: `copy.deepcopy` under a lock that `observe()` and `commit()` also take. Taking the lock in all three places is what makes it correct — a lock on `snapshot()` alone would still race `observe()`.

`deepcopy` is the right choice over a shallow copy: `BaselineProfile` nests `ToolBaseline` objects, so a shallow copy would still alias the per-tool accumulators.

### Verdict

✅ **Approve.** A method named `snapshot` that returns a mutable alias to live state is a correctness bug waiting to bite any caller that inspects-and-modifies, and the fix matches the codebase's existing locking convention exactly.

Cyrax321 (commented):
**Code Review — PR #362**

Correct, well-scoped fix: `snapshot()` now returns `copy.deepcopy(self._profile)` under a shared leaf lock, and `commit()` copies before the fsyncing `save()`. The deep copy is genuine — the nested `tools` dict of `ToolBaseline` dataclasses and the optional `embedding_centroid` list are all independent (verified) — the lock is never held across disk I/O, and there are no production callers, so no caller-breakage or shallow-share bug is present. Two low-severity notes.

Findings:

**1. `low` — `snapshot()` holds the same lock `observe()` uses across the whole `deepcopy`.**
`src/snagline/baseline_store.py:223`
`observe()` does `with self._lock: self._profile.add_event(event)`; `snapshot()` does `with self._lock: return copy.deepcopy(self._profile)`. Both share the non-reentrant lock, so a host polling `snapshot()` while another thread ingests stalls every `observe()` for the deepcopy duration — O(tools + embedding dim) allocation under the lock. The PR's "leaf lock, never across I/O" claim holds, but the copy itself is the critical section and can add a p99 spike to the per-step path. Building the copy from a shallow structural copy under the lock (or a dedicated snapshot lock) would bound it.

**2. `low` — the new `commit()` deepcopy is not covered by any test that fails without it.**
`src/snagline/baseline_store.py:235`
Verified: `save()` calls `profile.to_dict()` twice — once for the version file and once for the latest.json pointer — so a concurrent `observe()` between the two writes can make `latest.json` disagree with the just-written history entry. That is the only distinct value of copying in `commit()`, yet `test_commit_copies_so_a_snapshot_taken_later_cannot_reach_the_save` mutates the profile only *after* `commit()` returns, and `save()` has already serialized synchronously by then — so it (and the sibling commit test) pass unchanged on base master. Only the `snapshot()` tests fail pre-fix. A test that mutates the profile from another thread between `commit()`'s two serializations (or that asserts the two writes match) would pin the `commit()` copy.

PR #363, opened 2026-09-20
fix(detectors): drop non-finite and negative samples instead of deadening a detector for the run (#349, #350)

### LINKED ISSUE #349: bug: NaN or Inf tokens_in/tokens_out crashes the token-runaway detector, which fail-open then leaves dead for the run

### LINKED ISSUE #350: bug: one NaN latency_ms permanently disables a (episode, tool) CUSUM baseline with no error or log line

abyyxhek:
A single malformed sample from an adapter can blind a detector for the rest of the run. The two issues are siblings with distinct failure modes: #349 dies loudly (an exception), #350 dies silently (no exception at all).

## The two failure modes

**#349 — `TokenRunawayDetector` crashes on NaN/Inf and shrinks on negatives.** Token counts went through a bare `int((event.tokens_in or 0) + (event.tokens_out or 0))`. `int(NaN)` raises `ValueError`, `int(Inf)` overflows, and both escape `observe` into `Monitor.ingest`'s fail-open handler — the detector stays installed and reports nothing for the rest of the run. `tokens_in=-50` is subtler: it is silently accepted and *reduces* the running budget total, so a later breach looks smaller than it is.

**#350 — one NaN `latency_ms` permanently blinds the CUSUM.** During warm-up a NaN propagates into `mu0`, and every later step computes `max(0.0, cusum + (x - NaN) / sigma0 - k)`. CPython's `max(0.0, NaN)` returns `0.0`, so `cusum > h` is `False` **forever** — the detector goes dark with no crash and no log line, which means fail-open has nothing to catch. `inf` is the mirror image: `cusum` becomes `inf` and the detector alarms on every step.

Adapters derive these from provider `usage` objects and duration arithmetic, so a malformed usage blob or a zero-elapsed division is a realistic runtime failure, not a synthetic one.

## Changes

- **`token_runaway.py`** — a `_usable_tokens` helper coerces each field: `None`, non-finite, or negative counts become `0` (logged at debug, since a broken provider can emit every step). A step with no usable field at all is skipped exactly like today's documented "neither field present" case. Crucially, one bad field costs only its own half — a NaN `tokens_in` no longer discards a real `tokens_out`.
- **`latency_anomaly.py`** — non-finite or negative latencies are dropped before they reach the Welford learner, so warm-up is not satisfied by garbage either. A bad sample costs at most that step; the baseline is never poisoned.

Both follow the existing convention for unusable input: treat it like a missing field and stay quiet, rather than raising into the fail-open path.

## Verification

- `tests/test_issues_349_350.py` — 12 tests covering every malformed shape at the detector boundary, plus the end-to-end symptom through `Monitor.default` (`detector_errors == 0` and the burn still reported after four NaN events).
- Full suite: **863 passed, 2 skipped**, coverage 88.03%. `ruff check` and `ruff format --check` clean.
- 9 of the 12 tests fail against pre-fix code, so they are real regression guards.

Fixes #349. Fixes #350.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid latency values, including negative, infinite, or unavailable measurements, are now safely ignored instead of disrupting anomaly detection.
  * Invalid token counts are excluded from budget calculations, preventing malformed data from distorting totals or causing errors.
  * Monitoring continues operating after malformed samples, preserving warm-up behavior and allowing anomaly detection to resume with valid data.
  * Added safeguards to prevent invalid measurements from disabling detectors or triggering repeated false alarms.


Adhi1-2:
**Overlap note:** the `latency_ms` non-finite/negative hunk here overlaps with #361 (which fixes #348 and #350). Both edit the same lines at the top of `LatencyAnomalyDetector.observe`, so whichever merges second hits a textual conflict on that hunk — the treatments are functionally identical (both return early on non-finite or negative latency; yours additionally logs at debug), so keeping either is fine.

The two PRs are otherwise disjoint — #348 (the `latency_count` seeding gate) is #361 only, and #349 (token-runaway NaN/Inf/negative) is yours only. I left the same note on #361 so the maintainer has it from both sides.

One small thing worth a look in the token-runaway hunk: `if tin == 0.0 and tout == 0.0: return None` now treats "both fields were unusable" the same as "neither field was present". That is the right call for the detector (a step with no usable token signal contributes nothing), but it does mean a stream where every event has a malformed `usage` object skips warm-up forever and `token_min_samples` is never satisfied — the detector stays silent rather than alarming. Probably fine (garbage in, nothing to say), and it matches how `latency_ms=None` is handled, but it is a behaviour change worth knowing about: previously `int(NaN)` raised and fail-open killed the detector, now it is a silent no-op. A debug log like the one on the latency side would make that observable. Not blocking from my side.

Adhi1-2:
Heads-up on overlap: this PR's `latency_anomaly.observe` guard is byte-identical to the one #361 was carrying for #350, so I dropped the #350 half from #361 and retitled it to the #348 calibrated-start fix only (`latency_count` gate). #348 is a different defect in the same function — the seeded statistics come from `latency_count` while the gate read `count` — and #363 doesn't touch it, so the two no longer overlap on any line and can merge in either order.

#379 is the same count/latency_count split still live in the `ml` extra (`esn_ensemble._mahalanobis_score`), kept separate because it touches an optional dependency.

Your #350 coverage is strictly better than what I had (the monitor-level fail-open test in particular), so it should be the one that lands.

Cyrax321 (approved):
## Review — PR #363: drop non-finite and negative samples instead of deadening a detector (#349, #350)

Checked out `pull/363/head`. 3 files, +199/-2.

### Verification performed
- ✅ **Full suite: 863 passed, 2 skipped** — no regressions; the 12 new tests in `test_issues_349_350.py` pass
- ✅ `ruff`, `mypy` — all clean (140 files)
- ✅ **Bug #349 reproduced on master**: `tokens_in=NaN` raises `ValueError: cannot convert float NaN to integer` out of `observe()` — which `Monitor.ingest`'s fail-open swallows, leaving the detector installed and dead for the run.
- ✅ **Bug #350 reproduced on master, and it's the nastier one**: after a warm-up of 3 `latency_ms=NaN` samples the CUSUM learns `mu0 = nan`, and **27 subsequent 8000ms calls never alarm** — permanently blind, with *no crash and no log line*, so fail-open has nothing to catch.
- ✅ **Branch fixes both**: NaN tokens survive; NaN latencies are dropped before they reach the learner, so the state is never poisoned in the first place.

### Assessment

The two failure modes are genuinely distinct and the PR treats them as such:

- **#349 dies loudly** (an exception into fail-open). The `int(NaN)`/`int(Inf)` path is a hard crash; negative tokens are subtler — silently *accepted* and *reducing* the running budget total, so a later breach looks smaller than it is.
- **#350 dies silently** — and this is the one worth calling out, because my reproduction confirms it's invisible by construction. CPython's `max(0.0, NaN)` returns `0.0`, so `cusum > h` is `False` **forever**. No exception, no log, no metric — the detector simply stops existing. A bug that fail-open cannot catch is strictly worse than one it can.

The `_usable_tokens` design detail is the strongest part: one bad field costs only its own half. A NaN `tokens_in` no longer discards a real `tokens_out`, which preserves signal instead of degrading to an all-or-nothing drop. And skipping a step with *no* usable field exactly matches today's documented "neither field present" case, so no behavior changes for well-formed events.

Following the existing convention — treat unusable input like a missing field and stay quiet, rather than raising into fail-open — is the right posture for a monitoring library whose whole design principle is never crashing the host.

### Verdict

✅ **Approve.** Both bugs are realistic runtime failures (adapters derive these from provider `usage` objects and duration arithmetic, so a malformed usage blob or a zero-elapsed division produces them), and #350's silent-blindness is among the worst failure modes this library can have. The drop-instead-of-raise approach keeps a single bad sample from costing the whole run.

Cyrax321 (commented):
**Code Review — PR #363**

The PR closes the two reported entry points: non-finite/negative token and latency samples are now dropped instead of letting `int(NaN)`/`int(Inf)` escape into `Monitor.ingest` and letting a NaN latency freeze a poisoned Welford baseline. But the guard is field-wise rather than aggregate-wise, and the calibrated-start (#101) path seeds the same Welford state from a `BaselineProfile` without any finiteness check — so the #350 failure mode is still reachable from the other baseline source. All three findings reproduced against the head.

Findings:

**1. `high` — the calibrated-start `seed()` path bypasses the new non-finite check, so a poisoned `BaselineProfile` still freezes a NaN/inf baseline and blinds the detector.**
`src/snagline/detectors/latency_anomaly.py:237`
Verified: the new guard at line 209 only inspects `event.latency_ms`; line 237 is untouched — `state.seed(seeded.mean_latency, seeded.std_latency)`. `json.load` accepts the `NaN`/`Infinity` literals, so a profile fit from such a JSONL line yields `mean_latency=nan`; on first sight of the tool the detector freezes `mu0=nan`/`sigma0=1.0` (or `inf`/`inf`). Every later `update()` computes `max(0.0, cusum + (x - mu0)/sigma0 - k)`, and `max(0.0, nan)` is 0.0, so `cusum > h` is False for the rest of the episode — no exception, no `detector_errors`, no alarm. Reproduced: a `ToolBaseline` with a NaN `latency_sum` froze `mu0=nan` and 20 consecutive 99999 ms spikes all returned None. The PR body's "the baseline is never poisoned" is not true for this path — guard the seeded values too.

**2. `medium` — `int(tin + tout)` is unchecked, so a finite-field aggregate overflow still raises, and the guard itself raises on a huge int.**
`src/snagline/detectors/token_runaway.py:111,51`
Verified: `_usable_tokens` validates each field, but the aggregate `int(tin + tout)` is formed unguarded. Two individually-finite counts (`tokens_in=1.6e308`, `tokens_out=1.6e308`) sum to `inf` and `int(inf)` raises `OverflowError`, escaping `observe` into the fail-open handler — the #349 class this PR closes. Separately, a huge integer token count makes `math.isfinite(value)` itself raise `OverflowError: int too large to convert to float` before returning 0 (reproduced with `10**400`). Pre-PR the huge-int path happened to work; the new guard re-breaks it for its own stated threat model. Validate the aggregate (and int magnitude) too.

**3. `low` — the new `tin == 0.0 and tout == 0.0` skip silently changes warm-up/count semantics for genuine zero-token steps.**
`src/snagline/detectors/token_runaway.py:107`
Pre-PR, an event with present `tokens_in=0, tokens_out=0` flowed through `learn_only(0.0)` and incremented `state.n` (contributing to `min_samples`); post-PR it is discarded like a missing field. Reproduced: five zero-token steps now leave `_states` with no entry. A run whose early steps are legitimately zero-token never completes CUSUM warm-up, and a baseline that should include zero-volume steps is learned without them — and the docstring still says only "events carrying neither token field are ignored", so the contract change is undocumented.

**4. `low` — two divergent implementations of the same "usable sample" policy.**
`src/snagline/detectors/token_runaway.py:36`, `src/snagline/detectors/latency_anomaly.py:209`
`_usable_tokens` lives in token_runaway; latency_anomaly inlines the identical predicate and debug message. A shared helper in `detectors/base.py` is the root-cause fix that would have made finding #2 apply to both detectors, and token_runaway already imports from latency_anomaly.

PR #364, opened 2026-09-20
fix(token-runaway): drop non-measurement token counts instead of dying

### LINKED ISSUE #349: bug: NaN or Inf tokens_in/tokens_out crashes the token-runaway detector, which fail-open then leaves dead for the run

Adhi1-2:
## Problem

`TokenRunawayDetector.observe()` converts the token volume with `int((event.tokens_in or 0) + (event.tokens_out or 0))` **before** validating it. That ordering has two failure modes, both silent:

- A non-finite value — a malformed provider `usage` object decoded by an adapter — raises out of `observe()`. `int(nan)` is a `ValueError`, `int(inf)` an `OverflowError`. Neither is caught inside `observe`, so it propagates into `Monitor.ingest` and is swallowed by the fail-open guard. The detector stays installed and reports **nothing for the rest of the run** — the failure mode fail-open exists to protect against, but here it costs the whole episode.
- A negative count was accepted and added to the running budget total. Since the total is cumulative, a negative sample silently *reduces* real spend, letting a bad adapter hide a genuine breach.

This is the same class of bug as the latency detector (#350, #348) — the shared `_WelfordCUSUM` machinery is poisoned identically — but here it kills the deterministic envelope too.

## Fix

A token count that is not a usable measurement is now treated as "no signal for this step", exactly like an event carrying neither token field:

```python
raw = (event.tokens_in or 0) + (event.tokens_out or 0)
if not math.isfinite(raw) or raw < 0:
    return None
step_tokens = int(raw)
```

Dropped, not scored, not accumulated. The detector stays live for the rest of the episode.

The module docstring documents the rationale, including why negatives are dropped rather than clamped (reducing the budget total would hide a breach) and why `None` fields still pass through as `0`.

## Tests

10 new regression tests in `tests/detectors/test_token_runaway.py`. All verified to **fail on the pre-fix source and pass after**:

- `test_non_measurement_token_counts_are_dropped_not_fatal` — parametrized over 7 combinations (nan in/out, inf, -inf, both non-finite, negative in, negative out); asserts no risk and `_totals` unchanged
- `test_detector_stays_live_after_a_malformed_usage_blob` — NaN then -9999 on steps 0-1, then the 85% warning and the breach still fire correctly on steps 2-3
- `test_negative_count_cannot_hide_a_breach` — `-900 + 1100` used to read as 200 (under budget); now 1100 alone breaches
- `test_nan_mid_run_leaves_the_cusum_baseline_intact` — after freeze, a NaN leaves `mu0` and the accumulated `cusum` untouched, and the detector still alarms on a 10000/step burn afterwards

A dedicated `_usage_event` helper builds events carrying raw float `usage`-derived values, since the existing `_event` helper only emits clean `int` counts.

Closes #349.

## Checklist

- [x] `ruff check src tests` clean
- [x] `ruff format --check src tests` clean
- [x] `mypy src tests` clean (7 pre-existing errors in `src/snagline/auto/openai.py` are unchanged on master and invisible to CI, which installs no openai package)
- [x] full suite: 857 passed, 3 skipped (2 failures are environmental — a missing `snagline` console script and the real openai SDK being installed — identical to master)

## Summary by CodeRabbit

- **Bug Fixes**
  - Invalid token usage values, including negative or non-finite counts, are now ignored safely.
  - Token monitoring continues operating after malformed usage data instead of becoming inactive.
  - Invalid measurements no longer alter usage totals, hide budget breaches, or affect future alerts.

- **Documentation**
  - Updated release documentation to describe the handling of invalid token counts.


Adhi1-2:
Closing as a duplicate of #363, which opened 29 minutes-seconds earlier with the same fix for #349 and a strictly better design for the token case: it coerces each field independently (a NaN `tokens_in` no longer discards a valid `tokens_out`), whereas this PR drops the whole step. No point in two competing PRs on the same issue.

Worth noting for the maintainer: **#350 is already fixed in #361** (opened earlier, CI green), which also fixes #348 — the latency-side sibling. #363 covers #350 as well, so the two overlap on `latency_anomaly.py`. #361 is the narrower of the two (latency only); whichever you prefer, the other will need a rebase.

PR #365, opened 2026-09-20
fix(cli): make watch --episode-id attribute events to the given id

### LINKED ISSUE #354: bug: snagline watch --episode-id is silently ignored for attribution; events keep their own episode_id
## Description

`snagline watch --episode-id X` does not attribute events to X. The flag's help text promises it does:

> Episode id to attribute events to [default: each event's own episode_id; a synthesized id otherwise].

But `args.episode_id` appears exactly once in `cli.py` — line 552, `episode = args.episode_id or (args.file or "stdin")` — and that value is only used in the teardown loop as the zero-events fallback id:

```python
# src/snagline/cli.py:579-580
event = StepEvent(**obj)
...
monitor.ingest(event)
episodes.add(event.episode_id)
```

Ingestion uses `event.episode_id` straight from the parsed line. An operator using the flag to scope a multi-tenant stdin stream to one episode silently gets per-event attribution instead, and nothing warns them.

## Current Behavior

```
$ printf '%s\n' '{"step_id":"0","episode_id":"from-event","timestamp":1.0,"action_type":"tool_call","action_signature":"aaaa...","tool_name":"t"}' > /tmp/e.jsonl
$ snagline watch --file /tmp/e.jsonl --episode-id operator-override
{"episode_id": "from-event", "step_id": "0", "score": 0.5, "trigger": "loop", ...}
```

The emitted risk names `from-event`, not `operator-override`.

## Expected Behavior

Either make the flag do what its help text says — override `event.episode_id` at parse time before ingest — or reword the help text and the docstring to match the implemented behavior (a fallback id used only when the input carries no events). The shipped test currently reframes the flag as the zero-events fallback (`tests/test_watch_cli.py`), so the code and the help text are the two sides that disagree.

## Context

Adhi1-2:
Fixes #354.

## what was wrong

`snagline watch --episode-id X` did not attribute events to X. The flag's help text promises it does:

> Episode id to attribute events to [default: each event's own episode_id; a synthesized id otherwise].

But `args.episode_id` appeared exactly once in `cli.py` — as the zero-events fallback id at teardown:

```python
episode = args.episode_id or (args.file or "stdin")   # line 552
...
event = StepEvent(**obj)
monitor.ingest(event)                                  # event.episode_id verbatim
episodes.add(event.episode_id)
```

The issue's repro:

```
$ snagline watch --file /tmp/e.jsonl --episode-id operator-override
{"episode_id": "from-event", ...}     # names from-event, not operator-override
```

An operator using the flag to scope a multi-tenant stream to one episode silently got per-event attribution, and nothing warned them.

## the fix

The override is applied at parse time, before ingest:

```python
if args.episode_id is not None:
    event = dataclasses.replace(event, episode_id=args.episode_id)
```

`StepEvent` is a frozen dataclass, hence `replace`. A `None` flag keeps today's per-event behaviour, the empty-input fallback is untouched, and the teardown set (`episodes | {episode}`) now sees the same single id ingest did — so finalization-based detectors like `silent_abort` fire under the override id.

I chose making the flag work over rewording the help: a flag that silently does nothing is worse than either, and the documented intent (attribution) is the useful behaviour for the multi-tenant stdin case the flag exists for.

## validation

- Replaced the existing `test_watch_episode_id_override_*` test — it passed `--episode-id real-ep` against events that *already carried* `real-ep`, so it was agnostic to whether the override did anything — with two that are not: the issue's exact repro (risk names `operator-override`, and `from-event` must not appear at all), and a multi-episode file under an override collapsing to one episode with exactly one abort.
- Negative control: both new tests fail on clean `master`.
- Full suite: **848 passed**, coverage **87.96%** (gate 80%). The only 2 failures are pre-existing and environmental — they fail identically on clean `upstream/master` (this dev venv has `openai` installed, which the "SDK absent" test requires to be missing).

## Summary by CodeRabbit

- **Bug Fixes**
  - `snagline watch --episode-id X` now consistently attributes all ingested events to the specified episode.
  - Multi-episode event streams are correctly grouped under the supplied episode ID.
  - Risk handling and teardown state now use the same operator-provided episode identifier, preventing inconsistent episode attribution.


Cyrax321 (approved):
## Review — PR #365: make `watch --episode-id` attribute events to the given id (#354)

Checked out `pull/365/head`. 3 files, +65/-7.

### Verification performed
- ✅ **Full suite: 852 passed, 2 skipped** — no regressions
- ✅ **Mutation-tested**: reverting `cli.py` to master fails both new tests
  (`test_watch_episode_id_override_attributes_events_to_the_flag` and `..._collapses_a_multi_episode_stream`), restored → green.

### Assessment

The bug is a straightforward contract violation: the flag's help text says "Episode id to attribute events to," but `args.episode_id` appeared exactly once in `cli.py` — as the zero-events fallback id at teardown — while `monitor.ingest(event)` used `event.episode_id` verbatim. So an operator scoping a multi-tenant stream to one episode silently got per-event attribution, and nothing warned them.

The fix applies the override at parse time, before ingest, so both the ingest path and the teardown `episodes` set see one consistent id. Using `dataclasses.replace` is required because `StepEvent` is frozen — the comment notes this, which is the kind of detail that prevents a future "simplification" from breaking it.

Placement at parse time (rather than inside `ingest`) is the right choice: it keeps the override visible in one place and means the `episodes.add(event.episode_id)` bookkeeping stays correct without a second override site.

### Verdict

✅ **Approve.** A documented flag that silently did nothing is a clear bug, the fix is minimal and correctly handles the frozen-dataclass constraint, and both new tests are verified load-bearing.

Cyrax321 (commented):
**Code Review — PR #365**

The fix does what issue #354 asks: `watch --episode-id X` now overrides each parsed event's `episode_id` at parse time (via `dataclasses.replace` on the frozen slots dataclass), before `monitor.ingest` and before the teardown set is built, so finalize-based detectors fire under the flag id. Both `watch` input sources (stdin and `--file`/`--follow`) flow through the one loop, so coverage is uniform. Two gaps.

Findings:

**1. `medium` — the override runs after `StepEvent(**obj)`, so lines that omit `episode_id` can never be attributed to the flag.**
`src/snagline/cli.py:570-579`
Verified: the loop does `obj = json.loads(line); event = StepEvent(**obj)` inside a `try` that `continue`s on `(JSONDecodeError, TypeError, ValueError)`, and the override `if args.episode_id is not None: event = dataclasses.replace(...)` runs *after*. `StepEvent.episode_id` is a required `str` with no default, so a producer that does not emit `episode_id` raises `TypeError`, the line is skipped with "skipping malformed line", and the run reports `ingested 0 step(s)` — the flag does nothing precisely for the id-less events it could uniquely supply. Meanwhile a line with `episode_id: null` *is* rescued by the override, so the two id-less shapes behave inconsistently. Overriding `obj["episode_id"]` before construction (injecting it when absent) would cover both.

**2. `low` — an explicit empty `--episode-id ''` collapses every event onto `''`, contradicting the teardown's truthiness handling.**
`src/snagline/cli.py:552,579`
The override gates on `is not None`, so `''` (e.g. an unset shell variable) rewrites every event to `episode_id=''`, silently merging unrelated tenants into one synthetic episode and firing finalize detectors across them. But line 552 still uses truthiness — `episode = args.episode_id or (args.file or "stdin")` — so the zero-events path falls back to the filename/stdin while ingested events carry `''`, and teardown finalizes `{'', 'stdin'}`. The sidecar rejects empty ids (`episode_id must be a non-empty string`), so the CLI should too (or use `or` consistently).

**3. `low` — the replacement test dropped the only assertion that a non-empty override id is finalized on the zero-events path.**
`tests/test_watch_cli.py:235`
The zero-events test calls `main(['watch'])` with no flag and asserts only the summary string; it never sets `--episode-id` and observes no finalized risk, so a regression in the `episode = args.episode_id or ...` teardown branch is unverified for an explicit override.

PR #366, opened 2026-09-20
fix(silent-abort): stop restoring operator config from a peer snapshot (Fixes #347)

### LINKED ISSUE #347: bug: SilentAbortDetector.load_state overwrites live configuration from a restored snapshot
## Description

`SilentAbortDetector.load_state` restores a constructor configuration field from the snapshot, so restoring a snapshot written on a differently-configured host silently changes *which final steps count as "output"* on this one:

```python
# src/snagline/detectors/silent_abort.py:93-95
types = state.get("output_action_types")
if types:
    self.output_action_types = frozenset(types)
```

`output_action_types` is operator configuration, not per-episode state. A real silent abort can be missed (the snapshot's types don't include this host's output action) or a clean ending can be flagged (they do include a type this host only uses mid-run).

`MeltdownDetector.dump_state` also writes a config field (`window_size`), but its `load_state` correctly ignores it — the detector reads its own live `self.window_size`. The two detectors disagree, which makes this an inconsistency rather than a design choice.

## Current Behavior

```python
from snagline.detectors.silent_abort import SilentAbortDetector

d = SilentAbortDetector(output_action_types={"assistant_message"})
d.output_action_types          # frozenset({'assistant_message'})

# A snapshot written by a stock-configured host (the shipped default)
d.load_state({"output_action_types": ["message", "plan_step"], "last": {}})
d.output_action_types          # frozenset({'plan_step', 'message'})
```

The detector's own `dump_state` writes the field too, so the round trip through a peer host is what flips it.

## Expected Behavior

Write/read only per-episode state in `load_state` — drop the `output_action_types` assignment, mirroring `MeltdownDetector`'s treatment of `window_size`. `Monitor.restore_dict` already applies snapshots to detectors matched by name, so a differently-configured host is the normal case, not an edge case.

## Context

Adhi1-2:
## Problem

`SilentAbortDetector.load_state` restores a constructor configuration field from the snapshot:

```python
types = state.get("output_action_types")
if types:
    self.output_action_types = frozenset(types)
```

`output_action_types` decides which action types count as an output step — it is operator configuration, not per-episode state. `Monitor.restore_dict` applies snapshots to detectors matched by name, so a snapshot written on a **differently-configured host is the normal case**, not an edge case. Restoring one silently changed which final steps counted as "output" on this host — a real silent abort could be missed (the snapshot's types omit this host's output action) or a clean ending flagged (they include a type this host only uses mid-run).

`MeltdownDetector.dump_state` also writes a config field (`window_size`), but its `load_state` correctly ignores it — the detector reads its own live value. The two detectors disagreed, which makes this an inconsistency rather than a design choice.

## Fix

Drop the assignment. `dump_state` still records the field, mirroring `MeltdownDetector`'s treatment of `window_size`, so existing snapshot readers keep working and the config stays inspectable — only the restore path now ignores it. The module docstring records the write-but-ignore contract so it is not mistaken for an oversight.

Restored last-events are now scored under the live configuration.

## Tests

3 new regression tests in `tests/detectors/test_silent_abort.py`, all verified to fail on pre-fix code and pass after:

- `test_load_state_does_not_overwrite_output_action_types` — a custom-types detector restores a stock-types snapshot and keeps its own types, then still fires on a `tool_call` final step
- `test_restored_last_event_is_scored_under_the_live_config` — the counterpart: a snapshot whose writer considered `tool_call` an output type must not suppress a silent abort on a host that does not
- `test_dump_state_still_records_the_types_for_diagnostics` — guards the write side, so the fix is not read as "drop the field from the snapshot"; also confirms a stock-configured detector survives reading a custom-types snapshot

Closes #347.

## Checklist

- [x] `ruff check src tests` clean
- [x] `ruff format --check src tests` clean
- [x] `mypy src tests` clean (7 pre-existing errors in `src/snagline/auto/openai.py`, unchanged on master and invisible to CI, which installs no openai package)
- [x] full suite: 828 passed, 3 skipped (2 failures are environmental — a missing `snagline` console script and the real openai SDK being installed — identical to master)

## Summary by CodeRabbit

- **Bug Fixes**
  - Restoring a saved detector state now preserves the host’s configured output action types.
  - Restored episodes are evaluated using the current configuration rather than settings captured in the snapshot.
  - Snapshot diagnostics continue to record output action types.


Cyrax321 (approved):
## Review — PR #366: stop restoring operator config from a peer's snapshot (#347)

Checked out `pull/366/head`. 3 files, +78/-3.

### Verification performed
- ✅ **Full suite: 854 passed, 2 skipped** — no regressions
- ✅ **Mutation-tested**: reverting `silent_abort.py` to master fails all 3 new tests, restored → green.

### Assessment

The reasoning here is the valuable part. `Monitor.restore_dict` applies snapshots to detectors matched by *name*, so a snapshot written on a **differently-configured host is the normal case, not an edge case**. `output_action_types` decides which action types count as an output step — that's operator configuration, not per-episode state. Restoring it silently changed which final steps counted as "output" on this host, in both directions: a real silent abort could be missed (the snapshot's types omit this host's output action) or a clean ending flagged (they include a type this host only uses mid-run).

The fix's shape is well-chosen: **drop the assignment, keep the write.** `dump_state` still records the field so existing snapshot readers keep working and the config stays inspectable — only the restore path now ignores it. The module docstring records the write-but-ignore contract so it isn't mistaken for an oversight. This matches `MeltdownDetector`'s existing treatment of `window_size`, which writes it to state but ignores it on load — the PR correctly identifies the inconsistency between two detectors rather than inventing a new convention.

The three tests cover both directions of the bug plus the write side — the last one matters, because it guards against a future reader "fixing" this by deleting the field from the snapshot.

### Verdict

✅ **Approve.** A subtle config-vs-state conflation that produces wrong-but-silent detection on the restore path, fixed in the least-disruptive way possible with the inconsistency against `MeltdownDetector` explicitly reconciled.

Cyrax321 (commented):
**Code Review — PR #366**

Correct, minimal fix. `SilentAbortDetector` has exactly one operator-config field (`output_action_types`; `config` is discarded), so excluding it from `load_state` is complete, and `_last` — the only learned state — is still restored via `state.get("last", {})` with no new `KeyError`. Cross-file tracing confirms silent_abort was indeed the lone outlier: every other shipped detector already reads config from the live instance on restore, so no peer snapshot becomes partial. Two low-severity notes.

Findings:

**1. `low` — the retained-field rationale is factually wrong and invites a future recurrence.**
`src/snagline/detectors/silent_abort.py:101-103`
The comment and PR body say `dump_state` still writes `output_action_types` "so an older snapshot remains readable". Compatibility of an old snapshot depends solely on `load_state` reading `state.get("last", {})` and ignoring unknown keys — no reader of `output_action_types` exists anymore, so writing it cannot affect readability. It is now a write-only field. Keeping it for diagnostics is defensible (it mirrors `dedup`'s `cooldown_seconds` and `meltdown`'s `window_size`), but the stated reason is false; a maintainer copying that precedent will re-add config fields to snapshots and assume restore handles them.

**2. `low` — all three regression tests call `load_state` directly; none exercises the `Monitor.restore` path the fix is justified by.**
`tests/detectors/test_silent_abort.py:106-162`
The bug (issue #347) lives in `Monitor.restore_dict` applying a peer snapshot to a detector matched by name, yet no test builds two Monitors with differing `SilentAbortDetector(output_action_types=...)` configs and round-trips snapshot→restore. The detector-level tests do fail pre-fix and pass post-fix, so the fix is guarded; the integration claim is simply unverified, and a `restore_dict` regression could reintroduce cross-config contamination without failing any test.

PR #367, opened 2026-09-20
fix(compaction): pin-less compaction no longer drops an unconfirmed grace window

### LINKED ISSUE #356: bug: a pin-less compaction event silently discards an unconfirmed grace window, so a governance-decay finding never fires
## Description

A compaction event carrying no `pinned` list silently deletes an in-flight grace window whose pins were still unconfirmed and whose deadline had not been reached.

```python
# src/snagline/detectors/compaction_tripwire.py:131
st.pending = self._open_window(event.metadata, n)
```

The assignment is unconditional, and `_open_window` returns `None` when `pinned` is absent, empty, or not a list — "nothing can decay", which argues only about the *new* event. The previous window's decay is never reported.

The contract the docstring states is that a later compaction *"replaces the pending set"*, which reads as replacing it with a new set of pins. A host that compacts without pinning anything (a truncation-style compaction with no constraint tracking) therefore silently clears an outstanding governance-decay finding.

## Current Behavior

```python
from snagline.detectors.compaction_tripwire import CompactionTripwireDetector

def ev(i, at, meta=None):
    return StepEvent(str(i), "ep", 1000.0 + i, at,
                     make_signature(at, "t", str(i)), "t", metadata=meta or {})

d = CompactionTripwireDetector(grace_steps=3)
d.observe(ev(0, "compaction", {"pinned": ["pinA", "pinB"]}))
d._episodes["ep"].pending.pins          # {'pinA', 'pinB'}, deadline 4

d.observe(ev(1, "compaction", {}))      # no pins
d._episodes["ep"].pending is None       # True -- the unconfirmed window is gone
```

The two pins were still unconfirmed and the deadline (ordinal 4) had not arrived; the risk that would have fired never does.

## Expected Behavior

Either keep the previous window when a new compaction carries no pins of its own, or — if clearing is intended — make the docstring say so explicitly and log the dropped pins so an outstanding governance-decay finding is visible rather than vanishing.

## Context

Verified against master at `b667169`. Severity is low: it requires a host that mixes pinning and non-pinning compactions.

Adhi1-2:
## Summary

Fixes #356.

`CompactionTripwireDetector.observe()` handled a `compaction` event with:

```python
st.pending = self._open_window(event.metadata, n)
```

`_open_window` returns `None` whenever the event carries no usable pins — a missing `pinned` key, an empty list, or a malformed non-collection value. So a truncation-style compaction that tracked no constraints unconditionally wiped the pending window of an *earlier* pin-bearing compaction, including pins that were still unconfirmed and whose grace deadline had **not yet arrived**. The eventual `governance_decay` risk vanished silently — nothing raised, nothing warned, and the state dump showed no pending window at all.

This is the worst failure mode for a governance-decay detector: the exact condition it exists to catch is destroyed by an unrelated event.

## Change

A pin-less compaction only argues about itself ("nothing decayed here") and carries no information about the previous window's pins, so it no longer replaces it. Only a compaction that pins constraints of its own restarts the grace countdown, which preserves issue #90's "new compaction resets pending set" rule as originally intended.

```python
if event.action_type == _COMPACTION:
    window = self._open_window(event.metadata, n)
    if window is not None:
        st.pending = window
    # ... otherwise the previous window stands
```

The retained window stays fully live: a later `constraint_present` still retracts its pins, its original deadline still fires exactly once, and the fired latch is untouched. The module docstring's contract statement was updated to say "a later *pin-bearing* compaction replaces the pending set", since that is the behavior the contract always described.

I chose option (a) from the issue (retain the window) over (b) (keep clearing, but log the dropped pins) because (b) still loses the finding — the alert is the entire point of the detector, and a host that logs but does not alert on governance decay is the bug, not the fix. Dropping the pins is also unrecoverable: once the hashes leave the pending set, no later event can re-establish them.

## Tests

Five new cases in `tests/detectors/test_compaction_tripwire.py`:

- `test_pinless_compaction_keeps_the_previous_window_alive` — the issue's exact repro, parametrized over all three pin-less forms (missing key, empty list, malformed string). Asserts the risk fires at the **original** deadline naming both pins' 16-hex prefixes.
- `test_confirmation_after_a_pinless_compaction_still_counts` — the retained window is live, not frozen; both pins confirmed after the pin-less event stay silent.
- `test_pinless_compaction_does_not_refire_a_spent_window` — a pin-less compaction cannot reset the fired latch; still exactly one risk.
- `test_pinless_compaction_is_transparent_to_a_later_replacement` — the #90 rule is not weakened: a later pin-bearing compaction still replaces and restarts the countdown.

Negative control verified: with the source change stashed, the three parametrized variants fail with `0 == 1` (the silent drop); the other two pass on master, which is correct — they guard behavior that already worked.

## Verification

- `pytest tests/detectors/test_compaction_tripwire.py` — 22 passed
- Full suite — 853 passed, 3 skipped, coverage 87.96% (gate 80%)
- `ruff check src tests`, `ruff format --check src tests` clean
- `mypy src tests` — no new errors (7 pre-existing environmental errors in `auto/openai.py` from a locally-installed `openai` SDK; CI installs only `langchain-core`)

The two suite failures (`test_cli_mentions_read_timeout_next_to_max_body`, `test_instrument_openai_without_sdk_is_safe_noop`) fail identically on pristine master in this environment.


Cyrax321 (approved):
## Review — PR #367: pin-less compaction no longer drops an unconfirmed grace window (#356)

Checked out `pull/367/head`. 3 files, +119/-8.

### Verification performed
- ✅ **Full suite: 857 passed, 2 skipped** — no regressions
- ✅ **Mutation-tested**: reverting `compaction_tripwire.py` to master fails the 3 parametrized `test_pinless_compaction_keeps_the_previous_window_alive` cases, restored → green.

### Assessment

This is the most consequential of the four small fixes I reviewed in this batch, because it's a governance-decay detector destroying its own finding.

`_open_window` returns `None` whenever a compaction carries no usable pins — a missing `pinned` key, an empty list, or a malformed non-collection. The old code assigned that `None` unconditionally, wiping the pending window of an *earlier* pin-bearing compaction — including pins still unconfirmed and whose grace deadline had **not yet arrived**. The eventual `governance_decay` risk vanished silently: nothing raised, nothing warned, state dump showed no pending window at all.

The semantic argument for the fix is exactly right: a pin-less compaction only argues about itself ("nothing decayed here") and carries *no information* about the previous window's pins, so it has no business replacing it. Only a compaction that pins constraints of its own restarts the countdown.

The PR's option choice is well-reasoned and worth quoting back: option (b) — keep clearing but log the dropped pins — was rejected because "it still loses the finding. The alert is the entire point of the detector, and a host that logs but does not alert on governance decay is the bug, not the fix." Dropping the pins is also unrecoverable: once the hashes leave the pending set, no later event can re-establish them. The docstring contract statement was updated to say "a later *pin-bearing* compaction replaces the pending set," matching what the contract always described.

### Verdict

✅ **Approve.** The worst failure mode for a governance detector — the exact condition it exists to catch, destroyed by an unrelated event — fixed with a correct semantic argument and a defensible design choice between two options, with the reasoning recorded.

Cyrax321 (commented):
**Code Review — PR #367**

The core fix is correct: a pin-less compaction (`_open_window` → None) now leaves `st.pending` untouched, preserving the original `deadline_ordinal` and the `n >= deadline_ordinal` check, so a confirmed or already-fired window is unaffected, no off-by-one is introduced, and the retained window still fires exactly once at its original deadline. It is neutral with respect to the #402/#406 restore work (`dump_state`/`load_state` untouched and round-tripping identically). The problems are ancillary: a stale base/CHANGELOG conflict, an un-updated contract doc, and an undocumented false-positive tradeoff.

Findings:

**1. `medium` — stale base; conflicts with master and the reviewed diff omits master's newer Unreleased bullet.**
`CHANGELOG.md:21`
Verified: `gh pr view 367` → `mergeable: CONFLICTING`; the branch is behind 15, based on `b667169`, and `git merge-tree` reports a content conflict in `CHANGELOG.md` (both this PR's #356 bullet and master's `f5824c3` `--list-versions` bullet insert adjacent to the `#317` paragraph). The diff reviewed here is against an out-of-date base that lacks the #293 bullet, so it cannot merge until rebased.

**2. `medium` — the authoritative user-facing contract still says a later compaction replaces the pending set.**
`docs/DETECTOR_GUIDE.md:391`
That file still reads "later compaction replaces the pending set and restarts the grace window" verbatim, while the module docstring now says "only a new pin-bearing compaction opens a new window" and claims "that is the behavior the contract always described". The PR touches only `CHANGELOG.md`, the detector module, and the test; an adapter author following DETECTOR_GUIDE will implement a truncation-style pin-less compaction expecting it to clear the pending set and instead get a `governance_decay` risk. Update the guide (or cite it as a follow-up).

**3. `low` — retention removes the only way to clear an unconfirmed window for legitimately revoked pins.**
`src/snagline/detectors/compaction_tripwire.py:135`
`if window is not None: st.pending = window` makes a pin-less compaction fully transparent; the only clearing paths left are per-pin `constraint_present` (which cannot honestly be emitted for a removed pin) and `reset`. Reproduced at head: `compaction[pinned=A], compaction[{}], x, x, x` now emits `('governance_decay', '3')`, whereas base suppressed it. Issue #356 explicitly chose this option, so it is intended — but the false-positive surface for a policy-retired pin is real and absent from the PR's rationale and tests. A sentence in the module docstring and a test naming the tradeoff would make it intentional rather than incidental.

PR #368, opened 2026-09-20
fix(cli): range-check the enforcement knobs in Config, exit 2 on a bad value

### LINKED ISSUE #353: bug: snagline serve prints its banner and then tracebacks on an out-of-range --min-severity-for-halt instead of exiting 2
## Description

`snagline serve` dies with an uncaught traceback instead of the clean exit-2 usage error every other bad CLI value produces.

`min_severity_for_halt` and `halt_timeout_s` are validated only inside `Monitor._configure_policy` (`monitor.py:312`, `:300`). In `_cmd_serve` the monitor is constructed inside `with suppress(KeyboardInterrupt)` — no exception handling — so the `ValueError` escapes to the top level.

Every other malformed flag the CLI rejects cleanly (`--webhook-url` missing, `--keyfile` without `--certfile`, `--client-ca` without `--certfile`, `--max-versions < 1`), so this one is an outlier rather than a design.

## Current Behavior

```
$ snagline serve --port 0 --min-severity-for-halt 5.0
snagline serve: listening on http://127.0.0.1:0 (POST /events, GET /health)
  -- no auth token set, every endpoint is open to anyone who can reach this port
ValueError: min_severity_for_halt must be within [0, 1]
```

The banner prints *before* the validation failure, so the operator is told the server is starting and then handed a traceback. The same happens with `SNAGLINE_MIN_SEVERITY_FOR_HALT=5` (the value reaches the monitor through `Config`) and with `--halt-timeout 0`.

In-process reproduction:

```python
from snagline.cli import main
main()   # sys.argv = ["snagline","serve","--port","0","--min-severity-for-halt","5.0"]
# raises ValueError instead of returning 2
```

## Expected Behavior

Exit 2 with a usage message on stderr, matching how `_cmd_serve` already handles its other invalid flags. `Config.__post_init__` / `Config.resolve()` validate every other closed-set or ranged knob; adding `min_severity_for_halt` and `halt_timeout_s` there would catch this at `_build_config` time, before the banner prints.

## Context

Two adjacent observations from the same code, happy to split if preferred:

- `halt_timeout_s <= 0` is validated only when the policy is *already* `halt_webhook` (`monitor.py:300`), unlike `min_severity_for_halt` two lines below which is checked unconditionally. A negative timeout is accepted under `observe` and surfaces only when the policy is later switched on.
- `Monitor.default()` reads `cfg.halt_timeout_s` straight into `_configure_policy`, so a `Config` carrying a negative timeout is rejected there but accepted by the direct constructor.

Adhi1-2:
Fixes #353.

## what was wrong

`min_severity_for_halt` and `halt_timeout_s` were validated only inside `Monitor._configure_policy`. `snagline serve` constructs the monitor *after* printing its "listening on" banner and inside `with suppress(KeyboardInterrupt)`, so a bad value escaped as an uncaught traceback with exit code 1 — after the operator had already been told the server was up:

```
$ snagline serve --port 0 --min-severity-for-halt 5.0
snagline serve: listening on http://127.0.0.1:0 (POST /events, GET /health)
  -- no auth token set, every endpoint is open to anyone who can reach this port
ValueError: min_severity_for_halt must be within [0, 1]
```

Every other malformed flag on that command (`--keyfile` without `--certfile`, `--client-ca` without `--certfile`) exits 2 with a usage message, so this was an outlier, not a design. The same bad value through `SNAGLINE_MIN_SEVERITY_FOR_HALT` also silently broke `snagline watch` and `snagline replay` (verified: both exit 1 with a traceback).

## the fix

Range-check both knobs in `Config` itself — at construction and again after env/file layering — following the existing `_validated_horizon` / `_validated_stagnation` pattern. `Config` already validates every other closed-set or ranged knob; these two were the only ones left to the consumer. The failure now surfaces at `_build_config` time.

The three commands that resolve a config (`serve`, `watch`, `replay`) convert the `ValueError` into a usage error: exit 2, message naming the knob on stderr, before anything else is printed.

Also addressed the issue's adjacent observation: `halt_timeout_s <= 0` was checked only when `policy` was *already* `halt_webhook`, unlike `min_severity_for_halt` two lines below it. A non-positive timeout has no valid reading under any policy, and the gate meant a negative timeout set while observing surfaced only once the policy was later armed — while `Monitor.default()` rejected the same config the direct constructor accepted. It is now checked unconditionally in both `Config` and `_configure_policy`.

## scope note

The issue's third observation — `Monitor.default()` reading `cfg.halt_timeout_s` straight into `_configure_policy` while the direct constructor ignored the config — is #352, fixed separately in #359.

## validation

- 8 new tests in `tests/test_enforcement_policy.py`: exit 2 + message for a bad `--min-severity-for-halt`, bad `--halt-timeout` (not gated on `--halt-forward`), bad value via `SNAGLINE_*` for all three commands, `Config` rejection at construction and after env layering, the documented edges still legal, stock config never raising, and the direct constructor rejecting a bad timeout under `observe`.
- Negative control: 6 of the 8 fail on clean `master` before the fix; the other 2 are regression guards (stock config / documented edges) that must pass both before and after.
- Full suite: **833 passed, 3 skipped** (2 failures are environmental — a missing `snagline` console script and the real openai SDK being installed — identical to master), coverage **87.88%** (gate 80%).
- `ruff check src tests`, `ruff format --check src tests` clean; `mypy src tests` reports only the 7 pre-existing `src/snagline/auto/openai.py` errors, unchanged on master and invisible to CI (no openai package installed there).

Rebased onto current `master` (single commit): the branch was cut before `_validated_token_runaway` moved out of `Config` into the detector, so the conflict in `config.py` dropped that hunk — it is unchanged on master and out of scope here.


Cyrax321 (approved):
## Review — PR #368: range-check the enforcement knobs in Config, exit 2 on a bad value (#353)

Checked out `pull/368/head`. 6 files, +231/-19.

### Verification performed
- ✅ **Full suite: 859 passed, 2 skipped** — no regressions; `test_enforcement_policy.py` (43 tests, 8 new) pass
- ✅ **Bug #353 reproduced on master verbatim**: `snagline serve --port 0 --min-severity-for-halt 5.0` prints the banner `snagline serve: listening on http://127.0.0.1:0 …`, *then* raises an uncaught `ValueError` — telling the operator the server is up before dying.
- ✅ **Branch exits 2 before any banner**: `snagline serve: min_severity_for_halt must be within [0, 1]; got 5.0`.
- ✅ Boundary values correct: `5.0` and `-0.1` rejected, `0.0` accepted; `halt_timeout_s` of `0.0`/`-1.0` rejected, `0.25` accepted; stock `Config()` never raises.

### Assessment

The failure mode is a good catch: `snagline serve` constructs the monitor *after* printing its banner and inside `with suppress(KeyboardInterrupt)`, so a bad value escaped as an uncaught traceback with exit code 1 — after the operator had already been told the server was live. Every other malformed flag on that command (`--keyfile` without `--certfile`, `--client-ca` without `--certfile`) exits 2 with a usage message, so this was an outlier, not a design. Moving validation into `Config` (following the existing `_validated_horizon` / `_validated_stagnation` pattern) means the error surfaces at `_build_config` time, before anything is printed.

The adjacent observation is the more interesting half: `halt_timeout_s <= 0` was previously checked *only when `policy` was already `halt_webhook`*, unlike `min_severity_for_halt` two lines below it. That gating meant a negative timeout set while *observing* surfaced only once the policy was later armed — and `Monitor.default()` rejected the same config the direct constructor accepted. Checking unconditionally in both `Config` and `_configure_policy` closes that asymmetry.

The three-command coverage (`serve`, `watch`, `replay` each converting `ValueError` to exit 2) is right — the description verified the same bad value via `SNAGLINE_MIN_SEVERITY_FOR_HALT` silently broke `watch` and `replay` too.

### Findings

**1. Overlaps PR #359 (approved) — shared files, needs a rebase** — *coordination*

Both PRs change `src/snagline/monitor.py`, `tests/test_enforcement_policy.py`, and `CHANGELOG.md`. #359 makes the constructor *apply* the config knobs; #368 makes `Config` *validate* their ranges. The intents are complementary and both are needed for a fully correct path (validation without application leaves the constructor ignoring the config; application without validation leaves the bad-value crash), but they restructure the same blocks in `_configure_policy` and add tests to the same file.

A mechanical merge-tree check shows overlap but no hard conflict, because #368 appears to have been written against a base that already includes part of #359's shape. **Whichever lands second needs a rebase and a re-run of `test_enforcement_policy.py`** to confirm the two directions compose: `Monitor.default()` should both *reject* a bad value (from #368) and *honour* a good one (from #359) in one code path.

**2. `halt_timeout_s` has no upper bound** — *observation, arguably correct*

Only the lower bound (positive) is checked. A huge timeout (e.g. `1e9`) is accepted and would mean a halt attempt effectively never times out. That's a legitimate configuration for a host that wants fire-and-forget halting, so no upper bound is defensible — flagging only so it's a visible choice rather than an omission.

### Verdict

✅ **Approve.** The banner-then-crash is a genuinely bad operator experience and this fixes it the right way (validate in `Config`, where every other closed-set knob already lives, rather than special-casing the CLI). The unconditional `halt_timeout_s` check is a real correctness improvement over the policy-gated version. Merge after rebasing on #359 (finding 1) so the two compose cleanly.

Cyrax321 (commented):
**Code Review — PR #368**

The two numeric enforcement knobs are correctly moved into `Config` validation and `serve`/`watch`/`replay` convert the resulting `ValueError` into exit 2 before the listening banner, with no documented "0 disables" convention broken (the halt timeout must be positive; `min_severity_for_halt`'s 0 edge stays legal). Two gaps: the non-numeric enforcement knob `halt_url` is still validated only at Monitor construction, so the banner-then-traceback-exit-1 path survives for it; and the new timeout check is not NaN/inf-safe.

Findings:

**1. `high` — `halt_url` is still validated only in `Monitor._configure_policy`, so the banner-then-traceback-exit-1 path persists for it.**
`src/snagline/monitor.py:307`, `src/snagline/config.py:154-182`
Verified: `_validated_enforcement` checks only `min_severity_for_halt` and `halt_timeout_s`; `Config.halt_url` is declared (line 479) but never validated, so `Config.resolve`/`dataclasses.replace` at `cli.py:1026-1044` succeed. The new `except ValueError` in `_cmd_serve` wraps those `replace` calls, not `Monitor.default`. Failure scenario: `snagline serve --port 0 --halt-forward ftp://x` (or `SNAGLINE_POLICY=halt_webhook` with no `SNAGLINE_HALT_URL`) prints the "listening" banner (and "halt forwarding enabled -> ftp://x"/`-> None`), then `Monitor.default` raises `ValueError` inside the `suppress(KeyboardInterrupt)` block, escaping as a traceback with exit 1 — the operator was already told the server is up. If the PR body/CHANGELOG claims this class is closed, either move the scheme and required-when-halt_webhook checks into `_validated_enforcement` or narrow the claim.

**2. `medium` — the new `float(cfg.halt_timeout_s) <= 0` check is not NaN/inf-safe.**
`src/snagline/config.py:181`
`nan <= 0` and `inf <= 0` are both `False`, and `_coerce`/argparse `type=float`/`json.loads` all admit `nan`/`inf`. Failure scenario: `SNAGLINE_HALT_TIMEOUT_S=nan`, `--halt-timeout nan`, or a JSON config with `NaN` passes validation, the server starts, and every halt `urlopen(timeout=nan)` raises inside the broad fail-open `except Exception` at `monitor.py:631-641` — counted as a `policy_error` and failing open, so halting is silently disabled with no startup error. `min_severity_for_halt` is already NaN-safe (`not 0.0 <= x <= 1.0`), so only this check needs `not math.isfinite(...)`.

**3. `low` — `_resolve_config_or_usage` catches only `ValueError`, but the validators call `float()` on raw field values.**
`src/snagline/cli.py:110`
A config file with `{"min_severity_for_halt": null}` or `{"halt_timeout_s": null}` makes `float(None)` raise `TypeError` in `_validated_enforcement`, which the helper does not catch — so `snagline watch --config bad.json` exits 1 with a traceback instead of the promised exit-2 message naming the knob. Not a regression (master also tracebacks, later in `_configure_policy`), but the helper's contract is narrower than the error types the validators can raise; catching `TypeError` too would match the intent.

PR #369, opened 2026-09-20
fix(latency): seed calibrated start on timed-sample count, not total (#348)

### LINKED ISSUE #348: bug: calibrated latency baseline is seeded from a profile that carries no latency samples, paging critical on step 0

Adhi1-2:
## What

`LatencyAnomalyDetector`'s calibrated start (issue #101) seeded its CUSUM baseline from a `BaselineProfile` entry, gating on `ToolBaseline.count` — the count of *every* tool call — while the values it seeds (`mean_latency` / `std_latency`) are computed from `latency_count` only.

A profile fitted from a stream whose adapter reports no `latency_ms` legitimately has `count=100, latency_count=0, mean 0, std 0` — `ToolBaseline` keeps the two counts separate deliberately so error rates can be measured on timing-less streams, which the auto-calibration contract supports. The detector froze onto `mean=0.0`, and since `seed()` floors `sigma0` to 1.0, the first live tool call was scored as a ~120-sigma deviation and **paged critical on step 0 of every episode**:

```
FailureRisk(episode_id='ep', step_id='0', score=1.0, trigger='latency_anomaly',
            detail='latency 120ms deviates from baseline (mean 0ms)',
            severity='critical')
```

## Fix

Gate the seeded start on `latency_count` instead of `count`. A profile that never measured latency has nothing to seed a latency baseline from, so that tool takes the ordinary learn-then-freeze path — which still detects a real anomaly once warm-up completes. Profiles that *did* measure latency are unaffected and still skip warm-up.

`ToolBaseline.from_dict` defaults `latency_count` to `count` for profiles predating the field, so legacy baselines still seed exactly as before.

## Scope

One changed condition plus its comment in `src/snagline/detectors/latency_anomaly.py`. Also tightens one sentence in `docs/DETECTOR_GUIDE.md` and adds a CHANGELOG entry.

## Verification

Reproduced against `b667169` before the fix (the exact snippet from the issue); the critical risk is now `None` and the tool still alarms on a real 400 ms spike after warm-up.

- `pytest`: **855 passed, 3 skipped**, coverage **87.97%** (80% gate)
- `ruff check src tests`, `ruff format --check src tests`, `mypy src tests` — clean
- Mutation check: reverted the gate to `count` and re-ran — **4 of the 6 new tests fail** (the two that still pass are the deliberate "do not over-correct" guards: a timed profile must still seed and skip warm-up, and a legacy profile must still seed)

Fixes #348.


Adhi1-2:
Closing as a duplicate of #361, which was opened earlier and fixes the same gate (`latency_count` vs `count` in the calibrated-start seeding) plus the adjacent non-finite-latency poisoning (#350). Mine covered only the #348 half, so #361 is the strictly broader fix and should be the one that lands.

I did not realize #361 was already open when I filed this — it is from the same fork, and my issue-claim map had not caught it. My branch and its tests are redundant; nothing here needs merging. The 6 tests I added are a subset of #361's coverage (it includes a direct pin on the seeding gate, `test_seeding_gate_uses_latency_count_not_count`, which mine lacked).

PR #371, opened 2026-09-20
fix(config): range-check the semantic goal-drift CUSUM knobs

### LINKED ISSUE #370: bug: semantic_drift_cusum_h of 0 deadens the goal-drift detector with a swallowed ZeroDivisionError; negative h or k storms false positives
## Description

The optional semantic goal-drift detector (`src/snagline/drift/goal_drift.py`) runs a CUSUM gate keyed on two `Config` floats, and neither is range-checked:

```python
st.debt = max(0.0, st.debt + signal - self._cfg.semantic_drift_cusum_k)
if st.debt >= self._cfg.semantic_drift_cusum_h:
    h = self._cfg.semantic_drift_cusum_h
    score = min(1.0, 0.5 + 0.5 * (st.debt - h) / h)   # h is the denominator
```

`semantic_drift_cusum_h` is the denominator of that score, so a zero is a `ZeroDivisionError` on every scored step. It is raised inside `_observe`, and `observe` wraps it in the standard fail-open handler:

```python
def observe(self, event: StepEvent) -> FailureRisk | None:
    try:
        return self._observe(event)
    except Exception:
        logger.exception("snagline: semantic_goal_drift raised; ignoring (fail-open)")
        return None
```

So the exception is swallowed and re-logged with a traceback **once per step for the rest of the run**, and no `goal_drift` risk ever fires. The detector is dead and the log says why, 100k times.

This is the exact bug class of #331 and #351 (a threshold knob whose value is used as a denominator), in the one detector family PR #337 does not reach: that PR validates `cusum_h` / `token_cusum_h` / the sigma floors / the window knobs, but no `semantic_drift_*` field appears in it.

## Reproduction (upstream `master`, zero deps)

A baseline with a 2-D `embedding_centroid`, an injected embedder returning an orthogonal vector (maximum drift), `semantic_drift_min_samples=2`, 20 steps:

| config | risks fired | log lines |
|---|---|---|
| `semantic_drift_cusum_h = 0` | **0** of 20 | 19 `ZeroDivisionError` tracebacks |
| `semantic_drift_cusum_h = -1` | **19** of 20 | 0 |
| `semantic_drift_cusum_k = -1` | **19** of 20 | 0 |

The negative cases are the mirror failure: `st.debt >= h` is true for every step once `h < 0` (debt is clamped to `>= 0`), and a negative slack `k` adds to the debt instead of subtracting from it, so the alarm trips on essentially every step. A false-positive storm is the other thing a drift detector must never do.

## Expected behavior

Range-check the two knobs at construction and after env/file layering, following the `_validated_horizon` / `_validated_stagnation` precedent (issues #92, #132, #331): a bad value is a startup configuration error naming the knob, not a dead or storming detector discovered later.

- `semantic_drift_cusum_h` must be `> 0` — it is a denominator.
- `semantic_drift_cusum_k` must be `>= 0` — negative slack inverts the CUSUM.

Both defaults (0.5, 0.05) are valid, so stock configurations never hit the checks.

Out of scope: `semantic_drift_tolerance` (a nonsense value only shifts sensitivity, it does not crash or storm) and `semantic_drift_min_samples` (0 or below just starts scoring on the first step).

## Context

Verified against `master` at `b667169`. This is the semantic-detector half of what #337 is doing for the deterministic detectors; a separate issue keeps it independent of that PR's review.

Adhi1-2:
## Summary

Fixes #370.

`SemanticGoalDriftDetector` runs a CUSUM gate on two `Config` floats, and `semantic_drift_cusum_h` is the denominator of its alarm score:

```python
st.debt = max(0.0, st.debt + signal - self._cfg.semantic_drift_cusum_k)
if st.debt >= self._cfg.semantic_drift_cusum_h:
    score = min(1.0, 0.5 + 0.5 * (st.debt - h) / h)
```

Neither knob was range-checked, so both fail loudly *in the wrong direction*, and the detector's own fail-open wrapper hid the worse one:

| config | risks / 20 steps | log |
|---|---|---|
| `semantic_drift_cusum_h = 0` | **0** | 19 `ZeroDivisionError` tracebacks |
| `semantic_drift_cusum_h = -1` | **19** | — |
| `semantic_drift_cusum_k = -1` | **19** | — |

The `h = 0` case is the severe one: `observe()` swallows the exception and re-logs a full traceback once per step for the rest of the run, and no `goal_drift` risk ever fires. A negative `h` makes `debt >= h` trivially true (debt is clamped to `>= 0`), and a negative slack `k` adds to the debt instead of subtracting from it — either storms false positives.

This is the same bug class as #331 and #351 in the one detector family PR #337 does not cover: that PR validates `cusum_h` / `token_cusum_h` / the sigma floors / the window knobs, but no `semantic_drift_*` field. Filed separately so it doesn't sit behind #337's review.

## Change

One new validator following the existing `_validated_horizon` / `_validated_stagnation` / `_validated_token_runaway` precedent, wired at both checkpoints:

- `__post_init__` — catches a literal in code or a config file.
- `Config.resolve` — catches the env layer, which bypasses `__post_init__` via `setattr`. `SNAGLINE_SEMANTIC_DRIFT_CUSUM_H=0` coerces cleanly from a string, so without this it would start up green and die later.

```python
if cfg.semantic_drift_cusum_h <= 0:
    raise ValueError("semantic_drift_cusum_h must be positive; it is the denominator ...")
if cfg.semantic_drift_cusum_k < 0:
    raise ValueError("semantic_drift_cusum_k must be >= 0; a negative slack adds to ...")
```

Scope is deliberately just these two: they are the only `semantic_drift_*` knobs with a demonstrated failure mode. `semantic_drift_tolerance` out of range only shifts sensitivity (no crash, no storm), and `semantic_drift_min_samples <= 0` just starts scoring on the first step. `cusum_k = 0.0` is accepted — it is a legitimate "no debt decay". Both shipped defaults (0.5, 0.05) are valid, so stock configurations never trip the checks, and the whole semantic path stays opt-in and off by default.

I did **not** add a guard inside the detector: the repo's established policy for this bug class (#92, #132, #317, #331) is that a bad knob is a startup configuration error, not something the hot path works around.

## Tests

`test_semantic_drift_cusum_knobs_out_of_range_aborts_startup` in `tests/test_config_loader.py`, mirroring `test_token_runaway_envelope_out_of_range_aborts_startup`:

- both knobs, both signs, rejected at construction;
- both rejected through `Config.resolve` (the env-layering path);
- the shipped defaults survive `resolve` with `SNAGLINE_SEMANTIC_DRIFT_ENABLED=true`;
- `cusum_k = 0.0` accepted as a legitimate boundary.

Negative control verified: with `config.py` stashed the test fails `DID NOT RAISE ValueError`.

## Verification

- Full suite — 848 passed, 3 skipped, coverage 87.98% (gate 80%)
- `ruff check src tests`, `ruff format --check src tests` clean
- `mypy src tests` — no new errors (7 pre-existing environmental errors in `auto/openai.py` from a locally-installed `openai` SDK; CI installs only `langchain-core`)

The two suite failures (`test_cli_mentions_read_timeout_next_to_max_body`, `test_instrument_openai_without_sdk_is_safe_noop`) fail identically on pristine master in this environment.


Cyrax321 (commented):
**Code Review — PR #371**

Correct and well-placed fix: `_validated_semantic_drift` is added to `Config` and wired into both `__post_init__` and `Config.resolve`, rejecting `semantic_drift_cusum_h <= 0` and `semantic_drift_cusum_k < 0` (the env-coercion bypass of `__post_init__` is genuinely re-checked). But the guards are not NaN/non-finite-safe, unlike the established `_validated_cusum`/`_validated_stagnation` precedent, and the scope note dismissing `semantic_drift_tolerance` is factually wrong.

Findings:

**1. `high` — non-finite `semantic_drift_cusum_h`/`_k` (NaN/inf) bypass both guards and silently deaden the detector.**
`src/snagline/config.py:246,252`
Verified head code uses `if cfg.semantic_drift_cusum_h <= 0:` and `if cfg.semantic_drift_cusum_k < 0:` — both `False` for `nan`, and `inf <= 0` is also `False`. `_coerce(float, 'nan')` succeeds, so `SNAGLINE_SEMANTIC_DRIFT_CUSUM_H=nan` (or JSON/TOML `NaN`/`inf`) starts green. Downstream, `goal_drift.py:228` `st.debt >= nan` is always False, so the detector never alarms for the run; `k=nan` makes `max(0.0, debt + signal - nan)` stay 0.0 so debt never accumulates. The in-repo precedent is deliberately NaN-safe (`_validated_stagnation` uses `not 0.0 < x <= 1.0`; #337's `_validated_cusum` uses `not value > 0`) — this PR regresses to `<= 0`/`< 0`. Use `if not cfg.semantic_drift_cusum_h > 0:` and `if not cfg.semantic_drift_cusum_k >= 0:`.

**2. `medium` — `semantic_drift_tolerance` is left unchecked, but the stated rationale ("only shifts sensitivity") is false.**
`src/snagline/config.py:252`
`goal_drift.py:225-226` computes `tol = semantic_drift_tolerance; signal = 0.0 if dev <= tol else min(1.0, (dev - tol) / (2.0 - tol))` with `dev = max(0.0, 1.0 - sim) ∈ [0, 2]`. Evaluated: `tol=-0.3` gives `signal=0.13` even for an identical centroid (`dev=0.0`), so with default `k=0.05`/`h=0.5` debt grows ~0.08/step and pages every ~7 steps on healthy traffic; `tol >= 2.0` makes `dev <= tol` always true, `signal` always 0, and the detector silently never fires. The field has no validator at any layer. A `[0, 2)`-style bound (or at least `>= 0` and `< 2`) belongs in this function.

**3. `low` — no upper bound on `semantic_drift_cusum_k`: any value ≥ 1.0 silently disables the detector.**
`src/snagline/config.py:252`
The semantic signal is capped at 1.0 (`goal_drift.py:226`), so for `k >= 1.0`, `signal - k <= 0` and `st.debt` stays 0 — the fire condition `debt >= h` (h > 0) is unreachable with no startup error. The project rejects silently-unreachable fire conditions elsewhere (`stagnation_min_novelty=0.0`, `token_budget_warn_fraction>1.0`); a `k < 1.0` bound would match that precedent.

PR #373, opened 2026-09-20
fix(config): warn when an env key names an object-typed field (#355)

### LINKED ISSUE #340: bug: a recognized-but-non-scalar env var (SNAGLINE_CALIBRATION_BASELINE / SNAGLINE_GOAL_DRIFT_BASELINE) is silently dropped with no warning

### LINKED ISSUE #355: bug: Config.from_env_overrides silently drops non-scalar fields like SNAGLINE_GOAL_DRIFT_BASELINE, contradicting its docstring
## Description

`Config.from_env_overrides` logs a warning only for values that fail coercion. Fields whose type is not one of `(bool, int, float, str)` are skipped by construction, with no log line at all:

```python
# src/snagline/config.py:548
hint = _coercible_hint(hints[name])
if hint in (bool, int, float, str):
    try:
        overrides[name] = _coerce(hint, value)
    except ValueError:
        logger.warning("snagline: ignoring bad env %s=%r", key, value)
```

`_coercible_hint` deliberately keeps object-typed fields out of reach — `goal_drift_baseline` and `calibration_baseline` are `BaselineProfile | None`, which cannot be built from a string — so setting either env form is a silent no-op:

```python
cfg = Config.from_env({"SNAGLINE_GOAL_DRIFT_BASELINE": "/tmp/x.json",
                       "SNAGLINE_CALIBRATION_BASELINE": "/tmp/y.json"})
cfg.goal_drift_baseline      # None
cfg.calibration_baseline     # None
# (no warning logged)
```

The module docstring promises that ignored keys are *"ignored (logged at warning)"*, so the contract and the behavior disagree. An operator who sets the env form of either field gets zero feedback that it can never apply, and discovers it only by noticing the detector staying inert.

For these two fields the *path* forms (`goal_drift_baseline_path`, `calibration_baseline_path`) are the ones that load from a file; the object-valued env names simply cannot work.

## Expected Behavior

Log the same `ignoring bad env` warning (or a more specific "cannot be set from the environment") when a present env key names a non-scalar field, so a mistyped or impossible env var is as visible as a malformed one.

## Context

Adhi1-2:
## What

`Config.from_env_overrides` warned only for values that *failed coercion*. A present key naming an object-typed field — `goal_drift_baseline` or `calibration_baseline`, both `BaselineProfile | None` — was skipped by construction with **no log line at all**:

```python
hint = _coercible_hint(hints[name])
if hint in (bool, int, float, str):
    try:
        overrides[name] = _coerce(hint, value)
    except ValueError:
        logger.warning("snagline: ignoring bad env %s=%r", key, value)
# else: nothing — silently dropped
```

`_coercible_hint` keeps those fields out of reach of string coercion on purpose (a profile is a fitted artifact, not a path), so `SNAGLINE_GOAL_DRIFT_BASELINE` / `SNAGLINE_CALIBRATION_BASELINE` were silent no-ops. An operator setting either got zero feedback that it could never apply, and found out only by noticing the detector staying inert — while the module docstring promised that ignored keys are *"ignored (logged at warning)"*.

## Fix

Log the same class of warning when a present key names a non-scalar field. Because `from_env_overrides` is also the loader `Config.resolve()` uses, `serve` / `watch` / `replay` all surface it at startup.

The advice is per-field and accurate rather than generic:

- `SNAGLINE_CALIBRATION_BASELINE` → `use SNAGLINE_CALIBRATION_BASELINE_PATH for a file path` (that path form exists)
- `SNAGLINE_GOAL_DRIFT_BASELINE` → `pass the object in code` (there is **no** `goal_drift_baseline_path` field — a hardcoded `_path` suffix would have invented one and sent the operator chasing a nonexistent knob)

The existence check is generic (`f"{name}_path" in hints`), so a future object-typed field with a path variant gets the right advice automatically.

The method docstring's list of ignored cases now names this third one explicitly.

## Scope

`src/snagline/config.py`: one added `else` branch plus a docstring line. `tests/test_config_loader.py`: 4 tests. CHANGELOG entry. No README change needed — it already documents these two fields correctly as "cannot come from files or the environment; pass those objects in code, or use the path variant below".

## Verification

Reproduced against `b667169` before the fix: both env forms set nothing and logged nothing.

- `pytest`: **853 passed, 3 skipped**, coverage **87.98%** (80% gate)
- `ruff check src tests`, `ruff format --check src tests`, `mypy src tests` — clean
- Mutation check: removed the new `else` branch and re-ran — **3 of the 4 new tests fail**. The fourth (`test_unknown_env_keys_stay_silent`) is a scope guard that must pass both before and after: a genuinely unknown key like `SNAGLINE_BOGUS_FIELD` still stays silent, since a host's environment may carry keys belonging to sibling tools.

Fixes #355.

## Stacked on #371 → #374

This branch is now **rebased on top of #374**, which is itself on top of #371. All three PRs append independent logic to the same three files (`config.py`, `tests/test_config_loader.py`, `CHANGELOG.md`), so unstacked they would hit textual conflicts in `from_env_overrides` / `__post_init__` / the test module's import block and stall any automerge sequence.

The diff against #374's head is purely additive: it adds the non-scalar warning, its four tests, and its CHANGELOG entry, and touches nothing #371 or #374 own.

Suggested merge order is **#371, then #374, then this one**. The three changes are logically independent, so any order is correct; if the maintainer prefers a different sequence, say so and I will re-stack.

---

Fixes #340 — issue #340 (filed by me earlier) and issue #355 describe the same defect: a recognized-but-non-scalar env var like SNAGLINE_CALIBRATION_BASELINE / SNAGLINE_GOAL_DRIFT_BASELINE is silently dropped with no warning. This PR covers #340's reported scenario verbatim, including the `overrides: {}` no-op and the absent log line next to the "unknown key" and "bad value" cases, so it closes both. Linking them here so #340 does not stay open and attract a duplicate PR after merge.


Cyrax321 (commented):
**Code Review — PR #373**

Correct, well-scoped change: `Config.from_env_overrides` now logs a warning when an env key names an object-typed field, with a per-field actionable alternative. The detection is right — the only non-scalar `Config` fields are `goal_drift_baseline` and `calibration_baseline`, both warned, and a sweep of scalar fields produced no false positives. Three low-severity notes.

Findings:

**1. `low` — the warning can be emitted multiple times per process; there is no dedup/latch.**
`src/snagline/config.py:589`
`from_env_overrides` is called from several independent entry points: `Config.resolve` (`config.py:670`), `_resolve_episode_ttl` (`http_server.py:212-221`, reached when `episode_ttl_seconds` is None), and again via `make_server`. On `snagline serve` with `SNAGLINE_GOAL_DRIFT_BASELINE` set, the identical warning is logged at least twice at startup, and up to three more times for a host calling `make_server(monitor)` with default None args. Startup-noise only (never per-step) and it mirrors the pre-existing "ignoring bad env" behaviour, but a one-shot latch (as `goal_drift.py:209` uses) would keep it to one line.

**2. `low` — the CHANGELOG code span for the path variant is split across a line break, so it renders wrong.**
`CHANGELOG.md:22`
The source has `` `SNAGLINE_CALIBRATION_ `` + newline + `` BASELINE_PATH` ``, which CommonMark renders as `SNAGLINE_CALIBRATION_ BASELINE_PATH` (with an embedded space). A reader copy-pasting the suggested fix sets a variable that does not exist — the exact failure mode this PR prevents. Keep the span on one line.

**3. `low` — the diff carries stacked-PR changes beyond its stated scope.**
`src/snagline/config.py:589`
The base is stale (`b667169`; branch is ahead 2 / behind 15), so `gh pr diff 373` also shows `_validated_semantic_drift`/`_validated_divisor_thresholds` and their CHANGELOG rows from #371/#374. The genuinely-owned change is only the new `else` branch. Reviewers reading the diff as this PR's scope may mis-attribute a regression; the body should say so or the branch should rebase.

PR #374, opened 2026-09-20
fix(config): range-check the loop and error-cascade threshold knobs

### LINKED ISSUE #372: bug: the loop and error-cascade threshold knobs are unvalidated divisors: 0 deadens the detector via a swallowed ZeroDivisionError, negatives emit risks with negative scores
## Description

The four remaining count-threshold knobs in `LoopDetector` and `ErrorCascadeDetector` are used as **divisors** in their score computation, and none of them is range-checked:

```python
# src/snagline/detectors/loop.py:212 and :344
score = min(1.0, count / self.repeat_threshold * 0.5)
score = min(1.0, count / self.loop_stall_steps * 0.5)

# src/snagline/detectors/error_cascade.py:120 and :123
score = min(1.0, consecutive / self.consecutive_threshold)
score = min(1.0, sum(w) / self.error_threshold)
```

The `min(1.0, ...)` clamp bounds the top of the score but not the bottom, so a negative threshold does not clamp away — it produces a **negative score** that is emitted as a real `FailureRisk`.

None of `loop_repeat_threshold`, `loop_stall_steps`, `cascade_error_threshold`, `cascade_consecutive_threshold` appears in `_validated_horizon` / `_validated_stagnation` / `_validated_token_runaway`, and PR #337 covers only the window-size and min-samples knobs (`loop_window_size`, `cascade_window_size`, `loop_cycle_window_size`, `meltdown_window_size`, `cusum_min_samples`, `token_min_samples`) plus the CUSUM bars. These four are the threshold half of the same problem, in the same two detectors.

## Reproduction (upstream `master`, zero deps, 6 steps)

Constructing the detector directly and calling `observe`, so the exception is visible rather than swallowed:

| knob | result | monitor-visible result |
|---|---|---|
| `loop_repeat_threshold = 0` | `ZeroDivisionError` on the first repeat | 0 risks, one logged traceback per step |
| `loop_stall_steps = 0` | `ZeroDivisionError` once stall scoring starts | dead detector, per-step traceback |
| `cascade_consecutive_threshold = 0` | `ZeroDivisionError` on the first error | dead detector, per-step traceback |
| `cascade_error_threshold = 0` | `ZeroDivisionError` on the first error | dead detector, per-step traceback |
| `loop_repeat_threshold = -1` | fires on step 0 with **score = -0.5** | fabricated alert |
| `cascade_error_threshold = -1` | fires on the first error with **score = -1.0** | fabricated alert |

The zero cases are the same swallowed-exception deadening as #331 / #351 / #370: `Monitor.ingest` wraps detector `observe` in its fail-open handler, so the exception is logged once per step for the rest of the run and no risk ever fires. The negative cases are worse than #333's "fabricates alerts" — the fabricated risk carries a negative score, which reaches every sink as a real `FailureRisk`.

## Expected behavior

Require all four to be `>= 1`, following the `_validated_horizon` / `_validated_stagnation` precedent (issues #92, #132) and what #337 does for the window knobs: a bad value is a startup configuration error naming the knob, at construction and after env/file layering. `1` stays valid — it is a legitimate "alert on the first occurrence" — so the check is `>= 1`, not `> 1`.

All four defaults (3, 25, 3, 3) are valid, so stock configurations never trip the checks.

## Context

Verified against `master` at `b667169`. This is the threshold-knob counterpart of #333 (window knobs) and #370 (semantic CUSUM knobs); filed separately so it does not sit behind #337's review.

Adhi1-2:
Fixes #372.

## The bug

`loop_repeat_threshold`, `loop_stall_steps`, `cascade_error_threshold` and `cascade_consecutive_threshold` each **divide** their detector alert score:

- `detectors/loop.py`: `count / self.repeat_threshold * 0.5` (repeat + stall modes)
- `detectors/error_cascade.py`: `consecutive / self.consecutive_threshold`, `sum(w) / self.error_threshold`

They are settable via `SNAGLINE_LOOP_REPEAT_THRESHOLD` etc. and coerce cleanly from a string, but were never range-checked — unlike the window knobs (#333) and the CUSUM bars (#331). Both failure directions are bad:

- **`0`** raises `ZeroDivisionError` on the first candidate step. `Monitor.ingest` wraps every detector `observe` fail-open, so the exception is swallowed and re-logged **once per step** while no `loop` / `error_cascade` risk ever fires. The detector is dead for the entire run, and the operator sees a traceback storm instead of an alert.
- **Negative** values satisfy the `count < threshold` guard trivially and emit a `FailureRisk` with a **negative score** (measured: `-0.5` and `-1.0`), because `min(1.0, ...)` clamps only the top. That fabricated risk reaches every sink.

Reproduction on master, one detector over 8 steps:

```
loop_repeat_threshold = 0  risks= 0 raised=ZeroDivisionError: division by zero
loop_stall_steps      = 0  risks= 0 raised=ZeroDivisionError: division by zero
cascade_error_threshold       = 0  risks= 0 raised=ZeroDivisionError
cascade_consecutive_threshold = 0  risks= 0 raised=ZeroDivisionError
loop_repeat_threshold = -1  risks= 1  (score -0.5)
cascade_error_threshold = -1  risks= 1  (score -1.0)
```

## The fix

`_validated_loop_and_cascade` rejects values `< 1` for all four knobs, wired at both layers like the existing validators:

- `__post_init__` — catches literals (`Config(loop_repeat_threshold=0)`) and config files.
- `Config.resolve` — catches env layering, which bypasses `__post_init__` via `setattr`. Without this, `SNAGLINE_LOOP_REPEAT_THRESHOLD=0` starts up green and deadens the detector once steps flow.

The error names the knob and both failure modes, matching the style of the existing validators. `1` stays legitimate ("alert on the first occurrence"), and the shipped defaults are untouched.

## Verification

- New `test_loop_and_cascade_thresholds_out_of_range_abort_startup` in `tests/test_config_loader.py`; negative control confirmed (fails on master without the fix).
- `ruff check` / `ruff format --check` clean; `mypy src tests` unchanged at the pre-existing 7 errors in `src/snagline/auto/openai.py` (local `openai` SDK; CI does not install it).
- Full suite: **850 passed, 1 skipped, 87.99% coverage**. The 3 failures (`test_cli_mentions_read_timeout_next_to_max_body`, `test_real_actionledger_flags_review`, `test_instrument_openai_without_sdk_is_safe_noop`) reproduce identically on pristine `upstream/master` — all environmental, none related to this change.
- CHANGELOG entry added under `[Unreleased] / Fixed`.

## Stacked on #371

This branch is now **rebased on top of #371** (`fix/semantic-drift-knob-validation-370`), because both PRs append an independent validator call to the same two blocks in `config.py` (`__post_init__` and `resolve`) and would otherwise hit a trivial textual conflict that stalls any automerge sequence. The diff against #371's head is purely additive: it adds `_validated_loop_and_cascade`, its wiring at both call sites, its test, and its CHANGELOG entry, and touches nothing #371 owns.

Suggested merge order is **#371 first, then this one**; once #371 lands this PR should rebase clean. If the maintainer prefers the reverse order, say so and I will re-stack them the other way. The two validators are logically independent, so either order is correct.


Adhi1-2:
Closing this as a duplicate of #329, which I missed when I filed #372.

#322 (filed 2026-09-19, claimed by #329 the same day) is the same defect: `_validated_divisor_thresholds` in that PR range-checks all four of these knobs (`loop_repeat_threshold`, `loop_stall_steps`, `cascade_error_threshold`, `cascade_consecutive_threshold`) with the same `>= 1` bound and the same wiring at both `__post_init__` and `resolve`. My #372 landed a day later and restated it; this branch restated the fix. My apologies for the extra review surface.

I unstacked #373 off this branch, so nothing open depends on it. If you would rather land the focused one-function validator than review it inside #329's 28-file PR, the branch is still here and green -- say the word and I'll reopen. Otherwise #329 is the one to merge and this can stay closed.

Adhi1-2:
Duplicate of #322 / #329; see the comment above. Closing mine to keep one fix per defect.

PR #377, opened 2026-09-20
fix(server): compare the auth token in constant time

### LINKED ISSUE #375: security: the sidecar auth token is compared with ==, leaking it through a timing side-channel; the token guards the halt endpoint
## What

`SnaglineHTTPRequestHandler._authorized()` compares the caller-supplied token against the configured `auth_token` with plain `==`:

```python
# src/snagline/server/http_server.py
def _authorized(self) -> bool:
    token = self.snagline_auth
    if not token:
        return True
    authz = self.headers.get("Authorization", "")
    if authz.startswith("Bearer "):
        return authz[len("Bearer ") :].strip() == token   # <- data-dependent
    return self.headers.get("X-Snagline-Token") == token   # <- data-dependent
```

`str.__eq__` compares left to right and returns at the first differing character, so its runtime grows with the length of the shared prefix. That is a classic timing side-channel (CWE-208): an attacker who can reach the sidecar and average response times over many requests recovers the token character by character, prefix-first.

## Why it matters here

The token is the only thing standing between a network caller and every non-`/health` endpoint, including:

- `POST /events` — inject fabricated steps into the monitor (poison detections, burn an operator's alert budget),
- `GET /metrics` and `GET /risks` — read live telemetry and every raised risk,
- `GET /episodes` — enumerate active episodes,
- and, when `--halt-forward` is configured, the enforcement path: an attacker who obtains the token can deliver a `HaltDirective` and **stop a production agent run**.

`docs/ATTACH_ANY_SYSTEM.md` documents putting the sidecar behind a reverse proxy with TLS termination, so this is a network-facing gate in the intended deployment, not loopback-only.

## Measured leak

Local microbenchmark, 2M comparisons each, median of 5 runs, against a 20-character token:

| guess vs. real token | time / 2M calls |
|---|---|
| no shared prefix | 0.0350 s |
| shares 4 chars | 0.0353 s |
| shares 12 chars | 0.0358 s |
| full match | 0.0382 s |

About **1.6 ns per comparison** separates a total mismatch from a full match, ~9% relative. That is nanosecond-scale, so exploiting it over a real network needs a quiet sidecar and heavy averaging — I am not claiming a one-shot critical vulnerability. But the fix costs nothing and the comparison gate is exactly where constant-time comparison is expected.

## The fix

`hmac.compare_digest` (stdlib, so the zero-dependency core is preserved), on UTF-8 **bytes** rather than `str`:

```python
return hmac.compare_digest(
    authz[len("Bearer ") :].strip().encode(), token.encode()
)
```

The bytes form is required, not a style choice: `compare_digest` raises `TypeError: comparing strings with non-ASCII characters is not supported` when handed two non-ASCII `str`. `Authorization` is attacker-controlled, so the naive `compare_digest(header, token)` would turn a probe into a handler crash (a 500 or a dead thread from unauthenticated input). UTF-8-encoding both sides is a bijection for valid strings, so equality semantics are unchanged.

I have this fix ready with tests, including a regression guard for the non-ASCII header case.

## Scope deliberately left out

`startswith("Bearer ")` is case-sensitive, while RFC 7235 makes auth schemes case-insensitive — a client sending `bearer <token>` gets 401 and falls through to the header check. Separate concern, separate change if the maintainer wants it.

Adhi1-2:
Fixes #375.

## The bug

`SnaglineHTTPRequestHandler._authorized` compared the caller-supplied token with plain `==`:

```python
authz = self.headers.get("Authorization", "")
if authz.startswith("Bearer "):
    return authz[len("Bearer ") :].strip() == token   # data-dependent
return self.headers.get("X-Snagline-Token") == token   # data-dependent
```

`str.__eq__` compares left to right and returns at the first differing character, so its runtime grows with the shared prefix — a timing side-channel (CWE-208) that leaks the token prefix-first to anyone who can average response times.

That gate is the only protection on every non-`/health` endpoint: `POST /events` (telemetry injection, which can burn an operator's alert budget), `GET /metrics` and `GET /risks` (readout), `GET /episodes`, and — when `--halt-forward` is set — the directive path that can **stop a production agent run**. `docs/ATTACH_ANY_SYSTEM.md` documents deploying the sidecar behind a reverse proxy with TLS, so this is a network-facing gate in the intended setup.

## Measured

Local microbenchmark, 2M comparisons, median of 5 runs, 20-character token:

| guess | time / 2M calls |
|---|---|
| no shared prefix | 0.0350 s |
| shares 4 chars | 0.0353 s |
| shares 12 chars | 0.0358 s |
| full match | 0.0382 s |

~1.6 ns per comparison, ~9% relative. Nanosecond-scale, so real exploitation needs a quiet sidecar and heavy averaging — I am not claiming a one-shot critical vulnerability. The remediation is free, and constant-time comparison is the expected treatment at exactly this gate.

## The fix

`hmac.compare_digest` (stdlib — the zero-dependency core is preserved), on UTF-8 **bytes**:

```python
token_b = token.encode()
...
return hmac.compare_digest(authz[len("Bearer "):].strip().encode(), token_b)
```

The bytes form is required, not a style choice. `compare_digest` raises `TypeError: comparing strings with non-ASCII characters is not supported` when handed two non-ASCII `str`, and `Authorization` is attacker-controlled — the naive string form would turn an unauthenticated probe (`Authorization: Bearer café`) into a handler crash. UTF-8-encoding both sides is a bijection for valid strings, so authorization decisions are byte-for-byte unchanged. The `X-Snagline-Token` branch also short-circuits an absent header to `False` before the comparison, since `compare_digest` does not accept `None`.

## Tests

Five new tests in `tests/test_http_server.py`:

- `test_partial_prefix_token_is_still_rejected` — a guess sharing 22 of 24 characters never authorizes.
- `test_non_ascii_token_header_is_rejected_without_crashing[Authorization, X-Snagline-Token]` — sent through a raw socket (urllib refuses non-ASCII header values; a real attacker has no such scruple). Asserts a clean 401 and that the handler thread survives to serve a subsequent legitimate request 202. This is the guard against the naive-fix trap above.
- `test_absent_token_header_is_rejected_without_error` — the `None` short-circuit.
- `test_token_comparison_goes_through_hmac_compare_digest` — pins the invariant itself.

About that last one: **every behavioural test here also passes with plain `==`**, because string equality rejects a wrong token just as well — the vulnerability is purely in the timing, which is not separable from noise in a non-flaky unit test. So the invariant is pinned by observing the call it routes through; a silent revert to `==` would leave every other test green and this is the only line of defence. I verified that: stashing the source change, exactly this test fails.

## Verification

- `ruff check` / `ruff format --check` clean; `mypy src tests` unchanged at 7 pre-existing errors in `src/snagline/auto/openai.py` (local `openai` SDK installed; CI does not install it).
- Full suite: **854 passed, 1 skipped, 87.98% coverage**. The 3 failures (`test_cli_mentions_read_timeout_next_to_max_body`, `test_real_actionledger_flags_review`, `test_instrument_openai_without_sdk_is_safe_noop`) reproduce identically on pristine `upstream/master` — all environmental, unrelated to this change.
- CHANGELOG entry added under a new `[Unreleased] / Security` section.

## Deliberately out of scope

`startswith("Bearer ")` is case-sensitive while RFC 7235 makes auth schemes case-insensitive — a `bearer <token>` client gets 401 and falls through. Noted in the issue; separate change if the maintainer wants it.


Cyrax321 (commented):
**Code Review — PR #377**

Correct fix: `_authorized` now compares the token with `hmac.compare_digest` over UTF-8 bytes, with a `None` short-circuit for the `X-Snagline-Token` branch. It is behaviour-preserving for string pairs (`encode` is injective), closes the prefix-timing leak, and the byte form avoids the `TypeError` a naive `compare_digest(str, str)` would raise on non-ASCII header values. Two caveats.

Findings:

**1. `medium` — the new "partial prefix" test provides no real regression coverage.**
`tests/test_http_server.py:190`
The three guesses are all non-equal strings, so both `guess == token` and `compare_digest(guess.encode(), token.encode())` return False and yield the identical 401. The test passes byte-for-byte with the vulnerable `==` (the PR body itself concedes "every behavioural test here also passes with plain `==`"). A future revert to `==` keeps it green, so it guards nothing — only `test_token_comparison_goes_through_hmac_compare_digest` catches a regression. Either drop the misleading test or make it assert the property it claims (e.g. that `_authorized` routes through `hmac.compare_digest`, via mock).

**2. `low` — `compare_digest` is constant-time only for equal-length inputs, so token *length* remains a timing oracle; the "constant-time" claim is unqualified.**
`src/snagline/server/http_server.py:567`
CPython's `hmac.compare_digest` dispatches to `_tscmp`, which is constant-time only when `len(a) == len(b)`; here the attacker-controlled-length guess is compared against the fixed-length `token_b`. An attacker varying guess length can learn the secret's length even though the character-by-character leak is closed. This is inherent to all string token comparison (not a regression), so low — but the docstring, comment and CHANGELOG should say "constant-time for a fixed-length input" rather than the unqualified claim, so a future maintainer does not treat the gate as fully leak-free.

PR #378, opened 2026-09-20
docs(changelog): backfill [Unreleased] with the merged post-0.1.0 work

Adhi1-2:
## Summary

The `[Unreleased]` section documented only **#317** of the ~40 user-visible commits that landed between the `v0.1.0` tag and master. It still read *\"Nothing yet\"* under Added while the restored time axis, the global-mode instrumentation fix, the strict-restore hardening, and both perf PRs had all shipped.

This backfills it. Coverage by area:

| area | entries |
|---|---|
| core & state (snapshot/restore, locks, fail-open) | 7 |
| detectors (stagnation, latency, goal-drift, ESN, signatures) | 7 |
| monitor & time axis (clock rewind, breach warning, retention cap) | 4 |
| adapters & auto-instrumentation (global mode, streams, clock) | 4 |
| server, CLI & sinks (batch ingest, watch finalize, min-severity) | 6 |
| Changed (perf: signature LRU, episode-lock) | 2 |
| Added (hook bridge fields, site) | 2 |

## Verification

Every entry was written against the merge diff and the commit body, **not the PR title**. That mattered: several titles invert or blur what the code does.

- #283's title reads \"separate the fill gate from the novelty gate\", which sounds like a refactor of an existing gate. It is a missing *second* condition — a window that had not yet filled its **scaled** target reported a novelty rate over the target size and fired a spurious risk.
- #286 restores windows at the **effective** size for the restored scaler position, not \"windows\" generically; the base-size rebuild silently truncated history when auto-scaling had grown the window live.
- #261 is not \"add a warning\" but the reverse — a score-0.7 pre-breach warning fired *after* the score-1.0 breach, severity running backwards.
- #256 does not \"handle negative deltas\"; it stops `last_ts` from being **rewound**, which is what caused the double-counting.

Where a commit body was empty I read the diff directly (the boolean env validation, the batching-sink rate-limit state, and the CUSUM baseline-adoption fix).

## Scope

Deliberately excluded:
- Dependabot and pure-formatting commits — no user-visible content.
- Test-only commits that pin existing behaviour (#285, the mypy annotations) — no behaviour changed, so no changelog line. Worth noting they exist, but Keep a Changelog's own guidance is that an entry describes a change a user would observe.
- The open PRs. This covers only merged work, so it does not preempt any in-flight branch.

The release workflow (`release.yml`) extracts release notes by regex over this file, so the section structure was re-validated against that exact pattern — it parses cleanly and the `0.1.0` section is still found.

Test suite: 847 passed, 3 skipped, 87.96% coverage. Two tests fail on this checkout, both pre-existing environment artifacts and both verified to fail identically with this change stashed on clean master: `test_cli_mentions_read_timeout_next_to_max_body` spawns an installed `snagline` binary this venv path lacks, and `test_instrument_openai_without_sdk_is_safe_noop` asserts a no-op with the SDK absent while this venv has `openai` installed (CI installs only `langchain-core`). Neither reads the CHANGELOG.

Ruff check and format clean.

## Summary by CodeRabbit

- **New Features**
  - Added a landing-page animation.
  - Improved `snagline hook` output and URL modes to serialize all available event fields.

- **Bug Fixes**
  - Documented fixes across core state handling, detectors, monitoring, time-axis behavior, adapters, auto-instrumentation, server functionality, CLI commands, and output sinks.

- **Documentation**
  - Expanded the unreleased changelog with a structured summary of additions, fixes, and changes.
  - Updated README guidance to avoid maintaining a static test count.


Cyrax321 (commented):
**Code Review — PR #378**

The bulk of the line-by-line claims verify against the corresponding merged commits, and the structure is reasonable. But the branch was last updated 2026-09-20 and its base `b667169` predates 15 further master commits (PRs #307, #328, #329, #292, #293), so it is `CONFLICTING`, omits that later work, and — because its base never contained the #293 entry — would delete a shipped fix from the changelog if merged as-is.

Findings:

**1. `high` — the backfill stops at `b667169` and omits all merged work after it.**
`CHANGELOG.md:8`
`git log 8edf3a9..master` lists 15 commits absent from the PR head, including #307 (O(1) window scaling / #298), #328 (12-issue docs/core/packaging batch, #295-#310), #329 (seven ingest-path contract fixes, #321-#327), and #292 (interactive site). If this section is cut into v0.1.1 by `release.yml`, none of that work appears in any released changelog. Rebase and extend the backfill to current master.

**2. `high` — merging as-is deletes the already-shipped `--list-versions` fix (#293) entry.**
`CHANGELOG.md:25`
`b667169` predates #293; master lines 14-25 carry the entry (added by `f5824c3`). The PR rewrites the whole Fixed block from the stale base, so `8edf3a9:CHANGELOG.md` contains zero occurrences of "list-versions" — the shipped fix silently loses its record. (`gh pr view 378` → `mergeable: CONFLICTING`.)

**3. `medium` — the malformed trailing fragment in master's #317 entry is dropped rather than corrected.**
`CHANGELOG.md:25`
Master's #317 line ends with a stray paste: "…knob (#317). 57305ac (fix(cli): make --list-versions read-only on fit and retrain paths)". The PR rewrites the entry to end cleanly at "(#317)." and drops the fragment, but does not reattach the `--list-versions` content it belongs to. A manual conflict resolution can preserve the run-on; split it out properly instead.

**4. `medium` — the boolean-env entry overstates: a bad value warns and is ignored, it does not abort.**
`CHANGELOG.md:53`
`_coerce` raises for a bad bool, but `from_env_overrides` catches it and logs `ignoring bad env %s=%r` (`config.py:582-584`); `resolve` does not re-validate booleans. The entry says "a configuration error naming the value" and gives `SNAGLINE_FAIL_OPEN`/a `DETECTORS_ENABLED` variable that does not exist (`Config` has `fail_open`). The silent-skip class the entry claims to fix is exactly what happens.

**5. `medium` — the #217 entry conflates issue #236 and attributes the whole fix to #217.**
`CHANGELOG.md:22`
Issue #236 ("restore_dict(strict_names=True) applies snapshot state before the composition check rejects it") is distinct and cited nowhere; the entry also names `Monitor.snapshot` for logic that lives in `restore_dict` (#217's fix), so a grep of the named API finds nothing.

**6. `low` — the #314/#315 entry describes only the #314 half and cites #315 for it.**
`CHANGELOG.md:104` — commit `e526cdc` also fixes clock-merge orphanage and persists the deferred latency baseline-shift fields that issue #315 actually names; neither is mentioned.

**7. `low` — the retention-cap entry misstates the ordering and the pre-fix behavior.**
`CHANGELOG.md:101` — the merged fix loads detector state first and enforces/evicts after (not "before loading detector state"), and the prior bug was unbounded retained detector state, not a "transient" cap overshoot.

PR #379, opened 2026-09-20
fix(ml): gate the ensemble's latency term on latency_count, not count

Adhi1-2:
## What

The same defect issue #348 fixed for `LatencyAnomalyDetector` is still live in the `ml` extra. `EsnCusumDetector._mahalanobis_score` gated on `ToolBaseline.count` while the moments it reads are computed from `latency_count`:

```python
# src/snagline/ml/esn_ensemble.py:326
if tb.count < 2:
    return 0.0
...
sigma_l = max(tb.std_latency, _SIGMA_FLOOR_ABS_MS, _SIGMA_FLOOR_REL * tb.mean_latency)
```

`ToolBaseline` keeps the two counts separate deliberately — `baseline.py` documents `count` as every observed tool call and `latency_count` as the subset that carried `latency_ms`, so error rates can be measured on streams whose steps report no timing (the auto-calibration contract, #101). A healthy profile from a stream whose adapter never timed anything legitimately reads `count=100, latency_count=0`, so `mean_latency` is `0.0` and `std_latency` is `0.0`.

The latency term then scores a live `100 ms` call as `(100 - 0) / max(0, 1.0, 0.0)` — a 100-sigma deviation, saturating the term to ~1.0. That value feeds `max(anomaly, mahalanobis)`, so it overrode the ESN's healthy residual and armed the CUSUM on **every** step:

```
latency_ms=100.0 on every step, baseline has no latency samples
fires in 20 steps: 4
  step 4: score=0.625 trigger=ml_ensemble
  step 9: score=0.625 trigger=ml_ensemble
  ...
```

A healthy episode paged a false `ml_ensemble` risk every few steps, on any monitor with `ml_ensemble_enabled` and a baseline the latency detectors could never have used. #361 fixed the zero-dependency side of this and did not touch this module.

## Fix

The latency term now gates on `tb.latency_count >= 2` — the same threshold `ToolBaseline.std_latency` itself uses, and the same fix #361 applied. With no timed samples the tool contributes its error rate and nothing else, which is exactly what the count split exists to allow.

The error term is deliberately untouched. Removing the latency term must not remove the term #101's split is for: a tool that never errored in the healthy baseline is genuinely anomalous when it errors live, and that still scores ~1.0.

`snagline baseline` also printed the misleading line that hid this. `n=100 mean=0.0ms std=0.0ms` claims a hundred samples averaged zero milliseconds; it now reads `n=10 timed=0 mean=n/a std=n/a`. Tools whose every step was timed print byte-identically to before.

## Scope

`src/snagline/ml/esn_ensemble.py` (one condition plus a docstring), `src/snagline/cli.py` (a `_tool_line` helper used by both `baseline fit` output paths), `tests/test_ml_esn_ensemble.py` (4 tests), `tests/test_baseline.py` (2 tests), CHANGELOG. No behavior change beyond the two sites above.

## Verification

Reproduced against `b667169` before the fix (output above).

- `pytest`: **877 passed, 2 skipped**, coverage **88.00%** (80% gate)
- `ruff check src tests`, `ruff format --check src tests CHANGELOG.md` — clean
- `mypy src tests` — clean (run with `--python-version 3.12`: the repo pins `python_version = "3.10"`, which numpy 2.5.3's bundled stubs reject at their own `numpy/__init__.pyi:737` before any project file is checked — an environment artifact, not a project error)
- Mutation check: reverted the one-line gate and re-ran — **2 of the 4 new esn tests fail**, the two that pin the defect. The other two are scope guards that pass both before and after: the untimed baseline still scores a genuine error > 0.9, and a timed baseline still flags a 5000 ms latency > 0.9.


Cyrax321 (commented):
**Code Review — PR #379**

Small, well-targeted fix: the ensemble's Mahalanobis latency term now gates on `tb.latency_count >= 2` instead of the ambient `tb.count`, matching the threshold `ToolBaseline.std_latency` already uses and eliminating the 100-sigma false alarms on untimed baselines. The one-line core fix is correct (the `lat > 0.0` and `max(std, 1.0, ...)` guards prevent falsy-zero and div-by-zero). Remaining items are consistency nits around the CLI helper, docs, and duplicated threshold logic.

Findings:

**1. `medium` — `_tool_line` prints `mean=n/a` for a tool with exactly one timed sample, hiding a well-defined mean.**
`src/snagline/cli.py:740-743`
Verified: the helper gates both fields on `if tb.latency_count >= 2:` else `lat = "mean=n/a std=n/a"`. But `mean_latency` is defined at `latency_count == 1` (`latency_sum / latency_count`); only `std_latency` is undefined below 2 (`baseline.py:62`). A tool called 51 times with one timed call prints `n=51 timed=1 mean=n/a std=n/a`, discarding the real measurement and misreporting it as untimed — the same misleading-output class this PR set out to fix. Render the mean separately from the std.

**2. `low` — the changelog/body claim the zero-dependency latency detector "already had" this fix, but at the merge base it still gates on `count`.**
`CHANGELOG.md:24`
Verified: on the base, `LatencyAnomalyDetector.observe` still runs `if candidate is not None and candidate.count >= self.min_samples` (`latency_anomaly.py:218`), and the fix that changes it lives in the still-open #361. So the two detectors diverge until #361 lands, contradicting the changelog. Reword to reference #361 as pending.

**3. `low` — the `>= 2` gate also suppresses the latency term for one timed sample, an unannounced scope widening.**
`src/snagline/ml/esn_ensemble.py:341`
The previous `count >= 2` gate did include the single-timed-sample case; `>= 2` now excludes it, so a profile with one measured 100 ms call plus many untimed calls silently ignores a 9000 ms regression for that tool even though the healthy mean could anchor a floored sigma. The body only claims to exclude "no timed samples" — either accept this or note it.

**4. `low` — the `latency_count >= 2` threshold is duplicated across three modules.**
`src/snagline/baseline.py:62`, `src/snagline/cli.py:740`, `src/snagline/ml/esn_ensemble.py:341`
A `ToolBaseline.has_latency_moments` property (mirroring `std_latency`) would be the single source of truth; this PR introduces the CLI copy rather than reusing one, so a future threshold change must be edited in three places.

**5. `low` — `_tool_line`'s docstring references a `baseline fit` subcommand that does not exist.**
`src/snagline/cli.py:729` — the CLI registers only `baseline` and `baseline retrain`; the docstring (and the PR body) repeat the inaccurate name.

**6. `low` — the new test's docstring cites a "four times in 20 steps" repro but the body drives `_healthy(40)`.**
`tests/test_ml_esn_ensemble.py:265` — minor, but the evidentiary narrative and the executed stream should match.

PR #380, opened 2026-09-20
fix(tests): unstale MiniStorage double for CONTINUUM's append_event kwargs (#296)

### LINKED ISSUE #296: [tests] test_real_actionledger_flags_review crashes against current CONTINUUM: MiniStorage.append_event rejects the source= kwarg

Parthipashok04:
## What

`tests/sinks/test_continuum_sink.py::test_real_actionledger_flags_review` — the **only** test that exercises `ContinuumSink` against the real `continuum` package — crashed on import of the real API:

```
E  TypeError: test_real_actionledger_flags_review.<locals>.MiniStorage.append_event()
            got an unexpected keyword argument 'source'
../Continuum/src/continuum/actions/ledger.py:833: TypeError
```

The in-test `MiniStorage` double was written against an older `Storage.append_event` contract. The real signature has since grown keyword-only parameters, and `ActionLedger._record` passes one of them:

```python
self.storage.append_event(self.run_id, event_type, payload, source=self._source)
```

This is a **stale test double, not a product bug** — shipped sink code is untouched. `ContinuumSink` calls the ledger, never `Storage` directly.

## Fix

Spell out the current keyword-only parameters on the double and add a `**kwargs` catch-all so it stays a superset of the real signature:

```python
def append_event(
    self, run_id, type=None, payload=None, *,
    causer_event_id=None, expected_sequence=None, source=None,
    **_unused,   # future kwonly additions decay silently, not hard-fail
) -> Any:
```

This is the second time an API addition has silently broken the only live test this sink has — and CI cannot see it, because the test skips whenever `continuum` is not importable (CI never installs `CONTINUUM`). The catch-all is the point of the change, not just the three named parameters.

## Verification

Reproduced the original failure with a current CONTINUUM checkout on `PYTHONPATH`, then confirmed the fix:

- ✅ `tests/sinks/test_continuum_sink.py` — **8 passed** with real `continuum` importable (including the live test)
- ✅ Same module — **7 passed, 1 skipped** without `continuum` (CI path; clean `importorskip`)
- ✅ Full suite — coverage gate passes at **87.61%** (required ≥80%)
- ✅ `ruff check` / `ruff format --check` / `mypy` all clean
- ⚠️ 13 pre-existing failures in `tests/test_server_tls.py` + `tests/test_server_mtls.py` — confirmed identical on the pristine tree (unrelated TLS/environment issue in this dev environment), not caused by this change

Closes #296.

## Summary by CodeRabbit

* **Tests**
  * Updated the CONTINUUM integration test double to accept additional event parameters, improving compatibility with evolving event data.


Cyrax321 (commented):
**Code Review — PR #380**

Redundant with a fix already on master and currently unmergeable — close it. The change is directionally right (accept CONTINUUM's newer keyword-only `append_event` params), but master already carries an equivalent fix, so merging adds nothing and risks reverting master's version during conflict resolution.

Findings:

**1. `high` — master already contains this fix, and the PR is `CONFLICTING`.**
`tests/sinks/test_continuum_sink.py:148-156`
Verified: master's double already has the `**kwargs: Any` catch-all with the issue #296 comment (line 153, from an earlier merged commit). This PR is based on the older `b667169`, so its only hunk edits lines master has already changed; `gh pr view 380` reports `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`. Resolving toward the PR side replaces master's `**kwargs` with the named-params-plus-`**_unused` variant (functionally identical) and gains nothing; resolving the other way drops the PR. The body says "Closes #296", but #296 is already closed on master.

**2. `medium` — the stated root cause (CI-invisible test) is not fixed.**
`tests/sinks/test_continuum_sink.py:153-156`
The PR's own comment says "the live test is CI-invisible, so it can only decay silently". `.github/workflows/ci.yml` never installs `continuum-agent`, so `pytest.importorskip("continuum")` keeps the test skipped, and the `**_unused` catch-all now guarantees *future* signature drift passes silently rather than failing. Adding a `continuum-agent` CI leg (or asserting the double's accepted kwargs are a superset of the real `Storage.append_event` signature) would close the actual issue.

**3. `low` — the three spelled-out keyword-only params are dead surface.**
`tests/sinks/test_continuum_sink.py:149`
The body reads only `run_id`/`type`/`payload`; `causer_event_id`, `expected_sequence`, `source` are accepted and ignored, and `**_unused` would already absorb them — so the named params change no behavior and are strictly larger than master's single `**kwargs` line.

**4. `low` — silently discarding `expected_sequence`/`source` lets the double diverge from real Storage semantics.**
`tests/sinks/test_continuum_sink.py:149`
If CONTINUUM later uses `expected_sequence` for optimistic-concurrency enforcement or `source` in the fold, `MiniStorage` ignores them and the only live test stays green against a double that no longer models the backend. Inherent to hand-rolled doubles, but it is the specific weakening a signature-parity guard would catch.

Adhi1-2:
Heads-up: the #296 half of this PR landed on master independently in `04dc984` ("close 12 unattended docs/core/packaging/test issues"). The CONTINUUM sink's `MiniStorage` double now accepts the keyword params `ActionLedger` passes, so `test_real_actionledger_flags_review` no longer crashes:

```
tests/sinks/test_continuum_sink.py  7 passed, 1 skipped
```

**What this PR still offers that master does not have:** the dedicated `test-continuum` CI leg. Master's workflow has no continuum leg at all (`grep -c test-continuum .github/workflows/ci.yml` = 0), and the CONTINUUM sink's only *live* test skips unless the package is importable, while the zero-dep legs never install it. So the parity guard this PR adds only actually runs on a machine that happens to have continuum installed -- the exact silence that let the `source=` kwarg drift through as an in-the-wild `TypeError` in the first place.

That half is unchanged in value. The signature-parity fix itself is redundant now.

Two options:

1. Reduce this PR to just the CI leg, rebased on current master (it also branches from `b667169`, 15 commits behind, and would revert `04dc984`'s ci.yml ruff pin and the continuum double fix if merged as-is).
2. Close it, if you would rather not add a leg that tracks CONTINUUM's default branch.

Leaving it open for your call.

Parthipashok04:
Rebased onto current master (`2d61f86`) and reduced to the CI leg only, per @Adhi1-2's option 1 — the `append_event` kwargs fix landed independently in `04dc984`, so the signature-parity half is redundant and the branch no longer conflicts.

One thing worth flagging: **turning the leg on immediately caught live drift.** Against CONTINUUM `main` today, master's double fails with

```
AttributeError: 'MiniStorage' object has no attribute 'read_all_events'
```

`claim()` now scans full history for spent authorities on every claim whose arguments are a mapping (`continuum/actions/ledger.py:1018`), so the double needs the method. This is the case the `**kwargs` catch-all cannot make: it absorbs new *keywords* on one method, but cannot notice the *absence of another*. That is exactly the silence issue #296 was about, and it is now caught by CI instead of by a user hitting a `TypeError`.

I added `read_all_events` to the double so the leg lands green rather than red — if #441 lands first I'll rebase onto it and drop the hunk.

Response to the review findings:
- **CodeRabbit, persist-credentials:** applied, `false` on both checkouts in the new job.
- **CodeRabbit, pin actions to SHAs:** kept at `@v7` — the other 9 invocations in this file use the same mutable tags, and pinning only the two new ones leaves the file half-pinned. Happy to pin all of them in a separate hardening PR if you want it, but it should be repo-wide, not per-PR.
- **@Cyrax321, "close it":** the double fix itself is on master, but the CI-invisibility root cause is not — master has no continuum leg, so the live test still skips on every PR. This is that leg.

Parthipashok04:
CI note: the checks on this push show **"Approve and run workflows"** rather than a queue, because the PR branch lives in a fork and it modifies `.github/workflows/ci.yml` — GitHub gates workflow-modifying fork PRs behind a maintainer approval. Nothing is failing; the jobs have not started yet. Approving once runs all legs, including the new `Test (py3.12, continuum extra)` job.

I verified that leg end-to-end locally against CONTINUUM `main` (`e150b47`) with the exact command from the job: **8 passed**, including the previously-skipping `test_real_actionledger_flags_review`.

PR #381, opened 2026-09-20
fix(examples): honour --help in baseline_to_monitor.py and replay_offline_trajectory.py (#297)

### LINKED ISSUE #297: [examples] baseline_to_monitor.py and replay_offline_trajectory.py ignore --help and run their demo instead
## Description

Most runnable examples under `examples/` take flags through `argparse`, so `--help` prints usage and exits. Two of them build no parser at all, so every argument — including `--help` — is silently ignored and the script runs its full demo:

- `examples/baseline_to_monitor.py` (`def main()` at line 63, no `argparse`)
- `examples/replay_offline_trajectory.py` (`def main()` at line 56, no `argparse`)

The other scripts in the directory (`raw_loop_example.py`, `real_time_llm_demo.py`, `real_agent_executor_demo.py`, …) all define a parser, so these two are the outliers.

`--help` is the first thing a newcomer reaching for an example tries, and instead of usage they get a live run: `replay_offline_trajectory.py` writes a synthetic trajectory to a temp file and replays it, and `baseline_to_monitor.py` fits a baseline and prints detector output. Nothing is damaged, but the scripts' own docstrings document a plain `python3 examples/<name>.py` invocation with no flags, so the silent ignore is undocumented rather than intended.

## Current Behavior

```shell
$ PYTHONPATH=src python3 examples/replay_offline_trajectory.py --help
[demo] replaying 38 steps from /var/folders/.../snagline_demo_trajectory.jsonl
{"episode_id": "demo-replay", "step_id": "27", "score": 0.5, "trigger": "loop", ...}
...
[demo] replayed 38 steps; see stderr above for any FailureRisk lines.
$ echo $?
0
```

```shell
$ PYTHONPATH=src python3 examples/baseline_to_monitor.py --help
Fitted baseline for tools: ['search']
Healthy episode risks: 0 (expect 0)
Drifting episode risks: 2 (expect >= 1)
  - ml_ensemble score=1.00: ensemble of detector signals
  - ml_ensemble score=1.00: ensemble of detector signals
```

## Expected Behavior

Both scripts should accept `-h`/`--help` and print usage, the way `raw_loop_example.py` already does:

```shell
$ PYTHONPATH=src python3 examples/raw_loop_example.py --help
usage: raw_loop_example.py [-h] [--healthy]
...
```

The minimal fix is a one-argument `argparse.ArgumentParser` in each `main()` whose `parse_args()` result is ignored (both demos are currently parameterless). If a real flag falls out naturally — e.g. `--healthy` for `baseline_to_monitor.py`, mirroring `raw_loop_example.py`'s false-positive check — that is a nice addition, but not required: the point is that `--help` stops silently running a demo.

Parthipashok04:
## Summary

Closes #297.

`examples/baseline_to_monitor.py` and `examples/replay_offline_trajectory.py` built no `argparse` parser at all, so **every** argument — including `--help` — was silently ignored and the script ran its full demo instead of printing usage. Every other runnable script in `examples/` defines a parser, so these two were the outliers.

## Changes

**`examples/baseline_to_monitor.py`**
- Added an `ArgumentParser` in `main()` so `-h`/`--help` prints usage and exits.
- Added a real flag that fell out naturally: `--healthy` runs only the healthy episode (expect zero risks) — a false-positive check mirroring `raw_loop_example.py --healthy`, as suggested in the issue. Default behaviour is unchanged (healthy + drifting episodes).

**`examples/replay_offline_trajectory.py`**
- Added an `ArgumentParser` in `main()`; the demo is still parameterless, so `parse_args()`' result is intentionally unused.
- Replaced two `__import__("sys").stderr` calls with a top-level `import sys`, and dropped a pre-existing unused `EpisodeMeta` import (ruff flagged both while linting the file).

## Testing

- New `tests/test_examples_help.py` runs each previously-broken script (plus `raw_loop_example.py` as the reference) with `--help` and asserts:
  - exit code 0 and `usage:` in stdout,
  - no demo output markers (`[demo] replaying`, `Fitted baseline for tools:`, `Healthy episode risks:`) on stdout **or** stderr — i.e. the demo really did not run.
- It also has a guard test that runs **every** pure-stdlib example with `--help`, so future outliers are caught. Examples requiring an optional extra (`langchain`/`langgraph`) are skipped, since they cannot be imported without the extra installed.
- Verified manually: both scripts still run their demos as before, and `baseline_to_monitor.py --healthy` stops after the healthy episode.
- `ruff check` clean on all three changed files.
- Full suite: `840 passed`, 3 skipped. The 13 failures are all in `tests/test_server_tls.py` / `tests/test_server_mtls.py` (SSL handshake/cert tests) and are unrelated to this change — no `examples/` or test file here touches the TLS server path.


Cyrax321 (commented):
**Code Review — PR #381**

Both target scripts now parse args and exit 0 on `--help`, and the demos still run (verified against head). The fix is incomplete, though: a third example still ignores `--help`, and the new guard test silently skips exactly that file.

Findings:

**1. `medium` — `examples/langchain_example.py` still ignores `--help` and runs the demo.**
`examples/langchain_example.py:19`
Verified: the head file has `def main()` with no `argparse` anywhere, so `python3 examples/langchain_example.py --help` runs the live demo instead of printing usage. The PR body asserts "every other runnable script in examples/ defines a parser, so these two were the outliers" — that is false for this third script. Either fix it too or scope the body/issue honestly.

**2. `medium` — the guard test skips the one remaining broken example.**
`tests/test_examples_help.py:83,92`
`optional_markers = ("snagline[", "langchain", "langgraph")` and `if any(marker in source for marker in optional_markers): continue`. Since `langchain_example.py`'s docstring contains "langchain", the test skips it, never appends it to `checked`, and passes green while the bug stands. Any future example whose docstring merely mentions `snagline[drift]` is likewise skipped. The marker should be tied to an actual optional import (e.g. parse the AST for the import and skip only when it is absent), not a substring of prose.

**3. `low` — module-level `CASES` is dead code.**
`tests/test_examples_help.py:21`
`CASES` is defined but never referenced (each test hardcodes its script name), with a comment implying it drives the reference check. A maintainer adding a newly-broken example to `CASES` gets no coverage, and lint does not flag unused module-level names.

Parthipashok04:
Thanks for the review — all three findings addressed in `adf3d85`.

**1. langchain_example.py (medium)** — It was indeed the one remaining example with a `main()` and no parser. Added an `ArgumentParser` at the top of `main()`, *before* the optional `langchain_core` import, so `--help` prints usage and exits 0 even without the extra installed (the demo itself still needs it). Behaviour without the extra is unchanged: it still prints the install notice.

**2. The guard skipped exactly that file (medium)** — The `optional_markers` substring check matched the word "langchain" in the docstring, so `langchain_example.py` was never appended to `checked` and the test passed green while the bug stood. Replaced with AST + actual importability, never prose:
- every example with a `main()` must now *statically* define a parser — that check applies to the optional-extra examples too, since it needs no import;
- `--help` is then actually run, and a failure is attributed to a missing extra only when the example imports that extra **at module scope** and the interpreter genuinely lacks it (`find_spec`-guarded — it raises when a parent package is missing). Every skip now names the missing module, so it stays auditable.

Side effect is wider coverage, not just a fixed skip: CI installs `langchain-core`, so `real_agent_demo.py` moves from skipped to runtime-checked there; `real_agent_executor_demo.py` is skipped only because it needs the full `langchain` package (`langchain.agents`).

**3. Dead `CASES` (low)** — Removed. The per-script tests now share a `_assert_help_not_demo` helper, which also carries the langchain demo markers.

Testing: `tests/test_examples_help.py` 5/5 green both locally (no langchain) and in a simulated CI environment (langchain-core only); neutering the new parser makes the suite fail as intended. Full suite 841 passed, 3 skipped — the only failures are the 13 pre-existing TLS handshake tests, untouched by this change.

Adhi1-2:
Superseded by master for #297's stated scope: both examples the issue names now honour `--help`, landed in `04dc984` ("close 12 unattended docs/core/packaging/test issues").

Verified on master -- both exit 0 on `--help` and print usage rather than running the demo:

```
baseline_to_monitor.py --help: exit 0
replay_offline_trajectory.py --help: exit 0
```

`examples/baseline_to_monitor.py:66` builds an `argparse.ArgumentParser` with the documented flags, so the README's invocation examples now match.

**One residual, out of #297's scope:** this PR also adds the same `--help` handling to `examples/langchain_example.py`, which master did not touch. There, `--help` still prints the "This example needs langchain-core" install notice and runs the demo instead of usage. It is a genuine one-line improvement but a different file from what #297 asked for.

Closing this PR since its issue is resolved. Happy to ship the `langchain_example.py` one-liner as its own small PR if you want it -- say the word and I'll open it.

Adhi1-2:
Follow-up: I do not have permission to close this PR (it is not mine), so the "Closing this PR" in my previous comment did not take effect and it is still open. State it clearly rather than leaving a misleading instruction -- the call to close or reduce it is yours (or the maintainer's). Everything above about #297 being fixed on master still stands, and the `langchain_example.py` offer is genuine if you want it.

Parthipashok04:
Rebased onto current `upstream/master` and force-pushed (`adf3d85` -> `8bb7855`), which also clears the merge conflict.

Acting on @Adhi1-2's note that master's `04dc984` already resolved #297's stated scope for `baseline_to_monitor.py` and `replay_offline_trajectory.py`, the PR is now reduced to only what master still lacks:

1. `examples/langchain_example.py` — the residual @Cyrax321 flagged in review. `--help` now prints usage and exits 0 before the optional `langchain_core` import, so it works without the extra installed; the demo path is unchanged (still prints the install notice). This was the last example with a `main()` and no parser.
2. `tests/test_examples_help.py` — new on master. Runs each previously-broken script (plus `raw_loop_example.py` as the reference) with `--help`, asserting exit 0, `usage:` in stdout, and no demo markers on stdout/stderr. The guard walks all of `examples/` and fails on any future script with a `main()` but no parser; optional-extra skips are tied to a real `find_spec` check, never to prose.

The rebase resolved the two conflicting files in favour of master's versions — master already implements `--help`/`--healthy` there, so the previous cosmetic differences were dropped rather than re-litigated.

Testing: `tests/test_examples_help.py` 5/5 green; `ruff check` clean on both changed files; full suite 901 passed, 3 skipped, 13 failed — the 13 are all the pre-existing `test_server_tls.py` / `test_server_mtls.py` handshake failures on this openssl-less Windows host, unchanged by this work (no `examples/` or TLS path is touched).

PR #382, opened 2026-09-20
fix(sinks): catch closed-stream ValueError in ConsoleSink.emit (#327)

### LINKED ISSUE #327: bug: ConsoleSink.emit's except OSError misses a closed stream's ValueError, so alerts escape into the host ingest path

Parthipashok04:
## Summary

Fixes #327.

`ConsoleSink.emit`'s raw-stream path caught only `OSError`, but a **closed** stream raises `ValueError: I/O operation on closed file`. That exception escaped `emit()`, breaking the `AlertSink` fire-and-forget contract (`sinks/base.py`) for the **default** sink — the path every `Monitor.default()` user lands on.

The inline comment named "a closed pipe" as the motivating case, and closing a stream is exactly what raises `ValueError` instead of `OSError`.

## Changes

**`src/snagline/sinks/console.py`**
- Widen `except OSError` → `except (OSError, ValueError)` so both stream-failure shapes are swallowed.
- Port `HeartbeatSink`'s `_fault_logged` latch: a permanently dead stream now warns **once**, not once per alert — the existing comment already claimed "logged once" but the warning fired on every `emit`. The latch resets on a successful write, so a stream that recovers re-arms the warning.

**`tests/test_console_sink.py`** — 3 new tests:
- `test_console_sink_is_fire_and_forget_on_closed_stream` — the exact repro from the issue; a closed `StringIO` must not raise across repeated emits.
- `test_console_sink_logs_stream_fault_only_once` — 5 emits on a dead stream produce exactly 1 warning.
- `test_console_sink_fault_latch_resets_on_recovery` — a flaky stream that recovers then fails again re-arms the latch, and the recovered alert really lands.

## Verification

- Reproduced the `ValueError` traceback from the issue against `master` before the fix; gone after.
- `pytest tests/test_console_sink.py` — **6 passed**, and `console.py` is at **100% coverage**.
- Full suite: **839 passed, 3 skipped**. 13 failures in `test_server_tls.py`/`test_server_mtls.py` — these are **pre-existing and environmental**, not caused by this change. They shell out to `openssl req ...` to generate certs, which exits non-zero on this Windows host. Verified by running the same two files on a clean worktree of `master`: **identical 13 failures**.
- `ruff check` + `ruff format --check` clean; `mypy` clean.

## Note

This may overlap with #329 ("seven silent contract violations in the ingest path"), which references #327. Happy to rebase or close if that PR lands first.


Adhi1-2:
Confirmed this is a correct fix — I reproduced the same `ValueError: I/O operation on closed file` on `b667169` before seeing this PR.

Heads up that my batch PR #329 ("seven silent contract violations", covering #321–#327) already contains this exact change: the same `except (OSError, ValueError)` widening and the same once-only `_fault_logged` latch with reset-on-recovery. So the two are duplicates at that site.

This one is the better shape for reviewing and bisecting — single issue, single file, self-contained tests — so if the maintainer would rather take this than unpack it from the batch PR, I'll strip the `sinks/console.py` and `sinks/heartbeat.py` hunks out of #329 and rebase. Either order works; just don't merge both.

Cyrax321 (commented):
**Code Review — PR #382**

Superseded — please close or rebase rather than merge. The mechanism this PR ports (log-once latch + `(OSError, ValueError)` in `ConsoleSink.emit`) is correct and its tests are sound (flush inside the try, latch reset on recovery), but master already landed the identical fix via #329, and GitHub reports this PR `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY`.

Findings, most severe first:

**1. `high` — the PR's entire source change duplicates what master already merged in #329.**
`src/snagline/sinks/console.py:71-79`
Master's `console.py` already contains `self._fault_logged = False` init, the latch inside the except branch, `except (OSError, ValueError):` with log-once, and a *better* warning message ("...; further failures are silent"). Verified at master `2d61f86`. Merging as-is forces conflict resolution in both hunks; a resolution keeping the PR side of the warning hunk silently reverts master's improved message, and the three new tests land as near-duplicates of `tests/test_issues_321_327.py:335-365` (closed-stream no-raise, warn-once, recovery are all already covered). The right action is close — or rebase to only what master lacks, which is nothing functionally.

**2. `low` — the widened tuple also swallows `UnicodeEncodeError`, conflating content-encoding faults with stream death.**
`src/snagline/sinks/console.py:75`
`UnicodeEncodeError` subclasses `ValueError` and the `write()` inside the try encodes, so an alert whose detail is unencodable on an ASCII stderr arms the latch, and every subsequent drop — encoding-related or a genuinely different failure — is silent for the process lifetime. Delivery behaviour matches base either way; this is diagnostic conflation only. (Same tradeoff exists on master's merged version, so this applies to the landed code as much as this PR.)

**3. `low` — `TypeError` from binary streams still escapes `emit()`.**
`src/snagline/sinks/console.py:72`
`ConsoleSink(stream=open(path, 'wb'))` raises `TypeError` out of `emit()` into the host ingest path — the escape class this PR fixes for `ValueError`, left outside the tuple. Pre-existing on base and already tracked upstream: issue #391, with open PR #410 as the proper construction-time guard. Scoped follow-up, not a regression here.

**4. `low` — the `_fault_logged` latch is an unsynchronized plain bool under concurrent dispatch.**
`src/snagline/sinks/console.py:76`
Sinks dispatch outside Monitor's per-episode lock (`monitor.py:349-353`), so two threads hitting a dead stream can log duplicate warnings or miss a re-arm; `Monitor._log_fault_once` guards its set with `_fault_lock` while this latch does not. Worst case is duplicated (never missing) warnings — benign, and `HeartbeatSink` ships the identical pattern, so it matches repo convention.

Adhi1-2:
Superseded by master: this PR's fix landed independently in #329 (commit `b817d83`, "seven silent contract violations in the ingest path"), and it is the same fix, not a different one.

Master's `src/snagline/sinks/console.py:73` now has the identical shape this PR proposes:

```python
except (OSError, ValueError):
    if not self._fault_logged:
        self._fault_logged = True
        logger.warning(
            "snagline ConsoleSink: write to stream failed; dropping "
            "alert (fire-and-forget)"
        )
```

Both the `ValueError` widening and the once-per-outage `_fault_logged` latch, with the same "mirrors HeartbeatSink" reasoning. Verified the behaviour on master: a closed stream's alert is dropped with one latched warning and `emit` returns cleanly instead of escaping into the host path.

Closing rather than holding a second implementation of the same 5-line change. No residual scope that I could find.

Adhi1-2:
Follow-up: I do not have permission to close this PR (it is not mine), so the "Closing" in my previous comment did not take effect and the PR is still open. Posting this so the state is unambiguous -- please close it yourself, or a maintainer can. Everything above about the fix being on master still stands.

Parthipashok04:
## Closing: the fix landed on master, and every review finding is now tracked elsewhere

Following @Cyrax321's review, I verified the state of this PR against current master (`2d61f86`) before deciding whether to rebase. There is nothing non-duplicative left to carry, so I'm closing it rather than pushing a second implementation of an already-merged change.

### The issue this PR targets is fixed on master

`b817d83` ("seven silent contract violations in the ingest path") landed the identical shape this PR proposes. Reproduced both sides:

- **Pre-fix base `b667169`** — `ConsoleSink(stream=closed_stream).emit(...)` raises `ValueError: I/O operation on closed file` out into the caller. The bug is real.
- **Master `2d61f86`** — 5 emits on the same closed stream return cleanly, exactly **one** latched warning is logged, and `_fault_logged` is armed. Master's version also carries the clearer message (`...further failures are silent`).

Master's `src/snagline/sinks/console.py:71-79` now has the `except (OSError, ValueError)` widening, the `_fault_logged` latch, and the reset-on-recovery — all three of this PR's source changes. Its `tests/test_issues_321_327.py:335-365` cover all three of this PR's tests (closed-stream no-raise, warn-once, recovery). `pytest tests/test_console_sink.py tests/test_issues_321_327.py` → 27 passed, `console.py` at 100% coverage. (The 13 `test_server_tls`/`test_server_mtls` failures are the usual Windows/`openssl` environmental baseline here, unrelated to sinks.)

### Where each review finding now lives

| Finding | Covered by |
| --- | --- |
| #1 Source change + tests duplicate master | **master** — `b817d83` |
| #2 `UnicodeEncodeError` conflated with stream death | **#444** (open, mergeable) |
| #3 Binary-stream `TypeError` escapes `emit()` | **#410** (open, mergeable) — construction guard *and* the runtime `except (OSError, ValueError, TypeError)` net |
| #4 `_fault_logged` latch unsynchronized under concurrent dispatch | **#425** (open, mergeable) |

No residual scope. #4 in particular is better placed in #425, which is already rebased and mergeable.

### One loose end worth a maintainer's attention

**Issue #327 is still open** despite being fixed on master. `b817d83`'s message references "(#321-#327)" in parentheses, which GitHub does not parse as a closing keyword, so it never auto-closed. Citing `b817d83` should be enough to close it.

Thanks for the review, @Cyrax321 — and to @Adhi1-2 for flagging the duplication early, which made this an easy call. Reopening is still possible if master's version ever turns out to be insufficient.

PR #387, opened 2026-09-20
fix(config): range-check the goal-drift and ml-ensemble score knobs (Fixes #385)

### LINKED ISSUE #385: bug: the goal-drift and ml-ensemble score/tolerance knobs are unvalidated -- out-of-range values either storm false positives on healthy traffic or silently deaden the detector
`goal_drift_score_threshold`, `ml_ensemble_score_threshold`, `goal_drift_error_tolerance` and `goal_drift_latency_k` are compared against but never range-checked, at construction or after env/file layering. Every other comparison-style knob in the project is now guarded (`_validated_stagnation`, `_validated_divisor_thresholds` in #329, `_validated_cusum` / `_validated_counts_and_windows` in #337, `_validated_semantic_drift` in #371); these four were missed because they are not divisors and not counts, so they sit outside every existing validator's scope.

All four are reached through `Config.resolve`, which sets fields via `setattr` and therefore bypasses `__post_init__` entirely, so `SNAGLINE_GOAL_DRIFT_SCORE_THRESHOLD=-1.0` is accepted by every layer with no error.

## The two failure shapes

Both detectors gate their own output on a score threshold:

```python
# detectors/goal_drift.py:54
if score < self._cfg.goal_drift_score_threshold:
    ...  # re-arm and return None
# detectors/ml_ensemble.py:83
if combined < self._cfg.ml_ensemble_score_threshold:
    return None
```

Scores are clamped to `[0, 1]`, so a threshold **above 1.0 makes the comparison always true and the detector can never emit anything**, on any input. A **negative threshold makes it always false, so a score of exactly 0.0 -- an episode identical to the healthy baseline -- still pages.** The tolerance knobs are the same comparison run backwards: `goal_drift_error_tolerance` is *subtracted*, so a negative value *adds* to the drift, and a negative `goal_drift_latency_k` makes `z > k` true for every `z >= 0`.

## Reproduction

`PYTHONPATH=src python3` against `b667169`:

```
healthy 80-step episode, 4 tools, baseline fitted from the SAME traffic:
  goal_drift_score_threshold=0.5 (default)     0 risks
  goal_drift_score_threshold=-1.0              1 risks
  goal_drift_error_tolerance=-0.5              1 risks
  goal_drift_latency_k=-3.0                    1 risks

genuinely drifting episode (90x slower, every step errors):
  goal_drift_score_threshold=0.5             1 risks
  goal_drift_score_threshold=2.0             0 risks

ml_ensemble over a tier-1 detector scoring 0.9 on every failing step:
  ml_ensemble_score_threshold=0.5             40 risks
  ml_ensemble_score_threshold=2.0             0 risks
```

Four wrong outcomes from four knobs:

- `goal_drift_score_threshold=-1.0` pages `goal_drift` on an episode byte-identical to the baseline. The latch re-arms only when the score drops below the threshold, which a negative threshold makes impossible, so it stays latched for the episode.
- `goal_drift_error_tolerance=-0.5` turns a zero error-rate difference into a 0.5 contribution on every evaluation.
- `goal_drift_latency_k=-3.0` makes every tool's latency term fire (four tools x 0.3 = 1.0).
- `goal_drift_score_threshold=2.0` and `ml_ensemble_score_threshold=2.0` **deadens the detector**: the second block is an episode 90x slower than baseline with every step erroring, and the ensemble block is a tier-1 detector reporting 0.9 on all 40 failing steps. Neither emits anything.

The deadening side is the worse half. `SNAGLINE_ML_ENSEMBLE_SCORE_THRESHOLD=2.0` is a plausible operator typo for an env override, and the result is an opt-in detector that passes every health check, reports `detector_errors == 0`, and watches nothing -- the project's own class-1 failure mode.

## Proposed bound

A score threshold is only meaningful in `(0, 1]` -- `0` itself would fire on a zero score, and anything above `1` is unreachable given the clamp. The tolerance knobs are `>= 0`, and `goal_drift_latency_k` is `>= 0` for the same reason (`k` is a slack subtracted from a non-negative z). One `_validated_score_thresholds(cfg)` validator wired at `__post_init__` and `resolve`, alongside the existing ones, covers all four.

This is deliberately scoped to the four comparison knobs. `goal_drift_min_samples` is a count and belongs with the `_validated_counts_and_windows` family in #337, which does not currently list it.

Adhi1-2:
Fixes #385.

## what was wrong

`goal_drift_score_threshold`, `ml_ensemble_score_threshold`, `goal_drift_error_tolerance` and `goal_drift_latency_k` were compared against but never range-checked. Every other comparison-style knob is guarded by a validator now; these four were missed because they are **neither divisors nor counts** -- a bad value never raises downstream, it only mis-routes alerts.

Both detectors gate their own output on a score threshold:

```python
# detectors/goal_drift.py:54          # detectors/ml_ensemble.py:83
if score < self._cfg.goal_drift_score_threshold:     if combined < self._cfg.ml_ensemble_score_threshold:
    ...  # re-arm, return None                            return None
```

Scores are clamped to `[0, 1]`, so:

- **above 1.0** the comparison is always true and the detector can never fire on *any* input;
- **0.0 or below** it is always false, so a score of exactly `0.0` -- an episode indistinguishable from the healthy baseline -- still pages;
- a **negative** tolerance or `k` *adds* to the drift, because both are slack subtracted from a non-negative measure.

## the fix

One `_validated_score_thresholds(cfg)` validator, wired at `__post_init__` and `resolve` exactly like `_validated_stagnation` / `_validated_token_runaway`:

| knob | bound |
|---|---|
| `goal_drift_score_threshold`, `ml_ensemble_score_threshold` | `(0.0, 1.0]` |
| `goal_drift_error_tolerance`, `goal_drift_latency_k` | `>= 0.0` |

An out-of-range value is a startup configuration error naming the knob, at construction, from a config file, and from `SNAGLINE_*` env layering (which reaches the fields via `setattr` and therefore bypasses `__post_init__` entirely). Defaults are all in range, so stock configurations never hit the checks. The `*_enabled` flags remain the way to disable a detector.

## verification

Reproduced all four against `b667169` before fixing (issue #385 carries the script):

```
healthy 80-step episode, baseline fitted from the SAME traffic:
  goal_drift_score_threshold=0.5 (default)     0 risks
  goal_drift_score_threshold=-1.0              1 risks
  goal_drift_error_tolerance=-0.5              1 risks
  goal_drift_latency_k=-3.0                    1 risks

genuinely drifting episode (90x slower, every step errors):
  goal_drift_score_threshold=0.5               1 risks
  goal_drift_score_threshold=2.0               0 risks

ml_ensemble over a tier-1 detector scoring 0.9 on every failing step:
  ml_ensemble_score_threshold=0.5             40 risks
  ml_ensemble_score_threshold=2.0              0 risks
```

Negative control: neutering the two validator call sites fails the new test (`DID NOT RAISE ValueError`), so it pins the fix rather than the validator's mere presence.

Local gate: `849 passed, 1 skipped, 4 failed`; coverage 87.96% against the 80% gate; `ruff check` / `ruff format --check` clean. All four failures are pre-existing and environmental, not regressions -- three reproduce identically on pristine `master` (`test_cli_mentions_read_timeout_next_to_max_body`, `test_real_actionledger_flags_review` which is #296, and `test_auto_openai_without_sdk_is_safe_noop`), and the fourth is its anthropic twin, which fires here only because this dev venv has `anthropic` installed. That pair is exactly what #295 describes: the "SDK absent" tests never simulate the SDK's absence. CI installs neither SDK and is unaffected.

`mypy src tests` reports 11 errors in `auto/openai.py` and `auto/anthropic.py` on this checkout -- all pre-existing, all in the optional-SDK modules, and all invisible to CI, which installs no `openai`/`anthropic` package. Confirmed identical on pristine `master`.

## scope

Deliberately limited to the four comparison knobs the issue names. `goal_drift_min_samples` is a warm-up count and belongs in #337's `COUNT_AND_WINDOW_KNOBS` list, which does not currently include it -- I left a note there rather than duplicating that PR's family.


Cyrax321 (commented):
**Code Review — PR #387**

Clean extension of the established startup-validation pattern (#92/#132/#317): `_validated_score_thresholds` is wired at both `__post_init__` and `resolve`, the bounds are NaN-safe for the score thresholds (`not 0.0 < value <= 1.0`) with correct inclusive/exclusive choices, and no legitimate disabled state is rejected (the `*_enabled` flags remain the off switch). Tests cover construction, env layering, and boundaries. The residual gaps are an NaN/±inf hole in the *tolerance* check itself, the missing detector-level defense-in-depth guards the sibling validators ship with, and same-class sibling knobs left unvalidated.

Findings, most severe first:

**1. `medium` — the tolerance/slack check `if value < 0.0:` admits NaN and +inf, silently deadening the detector.**
`src/snagline/config.py:256-258`
Verified: `nan < 0.0` is False and `inf < 0.0` is False, so `goal_drift_error_tolerance=nan` / `goal_drift_latency_k=inf` pass — while the score-threshold loop in the same function (lines 246-248) uses the NaN-safe form. Reachable via env: `_coerce(float, 'nan')` succeeds and `resolve`'s re-validation (line 675) lets it through. Downstream, `GoalDriftDetector._drift_score` computes `max(0.0, tb.error_rate - (ref.error_rate + nan)) == 0.0` and `z > nan`/`z > inf` is always False (`goal_drift.py:87,97`), so error and latency drift can never contribute and the detector stays installed with `detector_errors == 0` — the silent failure mode this PR exists to make a startup error. `if not value >= 0.0:` (matching the sibling loop's form) closes it.

**2. `medium` — no detector-level defense-in-depth guards, unlike the sibling precedent this PR cites.**
`src/snagline/detectors/goal_drift.py:32-43`, `src/snagline/ml_ensemble.py:42-53`
`StagnationDetector` (#132) and `TokenRunawayDetector` (#317) each re-check their knobs in `__init__` "as defense in depth for direct use"; `GoalDriftDetector` and `MLOrchestrator` accept any Config unchecked. Verified: a host that builds a `Config`, then mutates `cfg.goal_drift_score_threshold = 2.0` post-construction (attribute assignment bypasses `__post_init__` — the exact path the sibling tests exercise for stagnation) reaches `Monitor.default()`, whose gate `score < 2.0` never fires and reports `detector_errors == 0` for the run. The Config-layer checks are real; the repo's documented two-layer contract for post-construction mutation is left unimplemented here.

**3. `low` — same-class sibling knobs in the goal-drift family remain unvalidated.**
`src/snagline/config.py:324-326`
`semantic_drift_tolerance` (and `semantic_drift_cusum_k`/`h`) have no range check at any layer on this branch, and tolerance has the identical mis-route failure shape: `=3` makes `signal` always 0.0 (`drift/goal_drift.py:226`), a negative one makes signal > 0 on every step. #371's semantic-cusum checks are not on this base, so this is a follow-up scope gap rather than a regression — but it is the same knob family the PR's title claims to cover.

**4. `low` — `goal_drift_min_samples` remains unvalidated; ≤0 scores from the first step, a huge value disables scoring permanently.**
`src/snagline/config.py:303`, `src/snagline/detectors/goal_drift.py:51`
`SNAGLINE_GOAL_DRIFT_MIN_SAMPLES=-3` resolves cleanly and `live.total_steps < min_samples` is False immediately, so a half-warmed profile pages on warm-up noise. The PR body's deferral note targets #337's `COUNT_AND_WINDOW_KNOBS`, which is not merged on this branch — so the invariant the note claims is deferred is not actually re-established anywhere yet.

**5. `low` — the validator docstring references functions that don't exist on this branch or master.**
`src/snagline/config.py:241-242`
`_validated_counts_and_windows` (and per the wording, `_validated_divisor_thresholds`' pairing) is cited as the family these knobs "sit outside", but grep finds neither name in the head file, the base, or current master — a dangling reference that misdocuments the validation surface until the #333/#337 family lands.

PR #396, opened 2026-09-20
fix(state): keep Redis episode locks alive past their TTL (#326)

Parthipashok04:
## What

`RedisStateBackend.episode_lock` held a 30s-TTL lock for a whole episode's ingest work and never extended it. When the section ran longer than the TTL, the key expired under the holder, `release()` failed the token check, and `LockNotOwnedError` escaped the `finally` into `Monitor.ingest` / `end_episode`.

Because it arrived *after* the `yield`, callers could not attribute it to locking — it surfaced as if the episode's own work had raised. Worse, the worker had already been mutating detector state on an expired lock, so mutual exclusion was already broken by the time the error appeared.

## The three knobs from the issue, all applied

1. **`lock_timeout` is a constructor parameter** — default raised from the fixed 30s to 300s, and reachable from the documented 12-factor env path via `SNAGLINE_STATE_REDIS_LOCK_TIMEOUT`. Non-positive values raise `ValueError` rather than becoming a silent no-op.
2. **Background renewal** — one daemon thread per backend calls `lock.extend()` on a cadence while sections are live, so the TTL bounds a *stuck* section rather than merely a slow one. Renewal resets to the full TTL, not the interval, so the deadline does not silently shrink to a third once renewal starts.
3. **A lost lock is logged, not raised** — the key is already gone, so there is nothing to release; the warning names the episode so the operator can see mutual exclusion may have been violated.

## Notes on the design

- **One renewer thread per backend, not one per `episode_lock` call.** `episode_lock` runs on *every ingest*; thread startup alone (~50 µs) would dwarf the ~1.9 µs per-step budget the memory backend is tuned for (#299). The renewer only ever *extends* locks and never acquires or releases them, so it cannot steal a lock from its holder.
- **`thread_local=False` is required** for renewal — redis-py stores the lock token in thread-local storage by default, so the renewer thread would see no token and refuse to extend. The renewer only extends, so sharing the token across threads is safe.
- **Scope kept to #326.** The acquisition-timeout contract (#325) is filed separately and is deliberately untouched here — that is why `episode_lock` still yields on a failed acquire rather than changing that behaviour mid-fix.
- **`LockNotOwnedError` is caught specifically**, not bare `Exception`, so a genuine redis connectivity error still surfaces instead of being swallowed as a lost lock.

## Testing

CI has no `redis` installed, so the tests inject a time-aware fake `redis` module into `sys.modules` (the same shape as the reproducer in the issue) that genuinely models TTL expiry and `extend()` pushing the deadline out — the tests are not mocked into vacuousness. 9 new tests in `tests/test_state_backend.py`:

- the exact issue reproducer (section outlives TTL → no raise, warning logged)
- renewal keeps a live section's lock past the nominal TTL, and extends at least once
- `thread_local=False` is actually requested when renewal is on
- the renewer drops a lost lock instead of churning against a dead key
- the renewer table drains on section exit (no per-episode leak)
- configurable timeout, env-var wiring, unparsable env value rejected
- the swallowed loss verified end-to-end through `Monitor.ingest` + `end_episode` with a slow detector

Verified that without renewal the same hold *does* lose the lock and warn, confirming the renewal test is not vacuous.

**Results:** `851 passed, 3 skipped`. The 13 `test_server_tls.py` / `test_server_mtls.py` failures are pre-existing on this machine — `openssl` is not installed locally — confirmed identical on a clean checkout of master with my changes stashed. `ruff check`, `ruff format --check`, and `mypy src` all clean; `state.py` coverage 95%.

## Summary by CodeRabbit

* **New Features**
  * Redis-backed episode locks now support configurable timeouts, with a 300-second default.
  * Active locks can renew automatically during long-running operations.
  * Lock timeout configuration can be set through the `SNAGLINE_STATE_REDIS_LOCK_TIMEOUT` environment variable.

* **Bug Fixes**
  * Expired Redis locks are now logged with the affected episode instead of raising an error during ingestion.

* **Documentation**
  * Added documentation covering lock coordination, timeout settings, renewal behavior, and lock-expiry handling.


Cyrax321 (commented):
**Code Review — PR #396**

Well-scoped, well-tested fix for the `LockNotOwnedError` leak (#326): the TTL is tunable, a single daemon renewer extends live locks at `ttl/3` cadence, unregister happens before release, `thread_local=False` keeps the token shared, and a lost lock is logged instead of raised. The core fix is correct. The renewer's broad `except Exception`, though, conflates transient Redis failures with a truly lost lock — which can silently re-defeat the mutual-exclusion guarantee this PR adds.

Findings, most severe first:

**1. `high` — `except Exception` in `renew_all` unregisters a lock on transient Redis errors, not just on true loss.**
`src/snagline/state.py:214-226`
redis-py's `extend` raises `LockNotOwnedError` only on token mismatch; connection failures raise `ConnectionError`/`TimeoutError`, which the bare `except Exception` also swallows into `self.unregister(_episode_id)`. Failure scenario: one renewal pass hits a transient `ConnectionError` while the section is still live → the lock is unregistered, no later pass extends it → the key expires at the next TTL boundary → another worker acquires it and both mutate detector state concurrently, the exact violation #326 reports — with only the (misleading) "TTL expired" release warning hinting at it. Catching `LockNotOwnedError`/`LockError` specifically and leaving other errors for the next pass would preserve the guarantee.

**2. `medium` — an in-flight renew pass can unregister a *new* holder's registration for the same episode.**
`src/snagline/state.py:199-226`
The pass operates on a snapshot taken under `self._lock` while exit/release proceeds: holder A exits (`unregister("ep")` → `release()`, which with `thread_local=False` clears the shared token), thread B immediately acquires and registers `("ep", L2)`, then the pass calls `L1.extend()` → redis `LockError("Cannot extend an unlocked lock")` → caught → `unregister("ep")` removes B's entry. B's lock is then never extended and expires mid-section. Small timing window and uncovered by tests, so PLAUSIBLE rather than CONFIRMED — but the table is keyed only by episode_id, so the pass cannot tell whose registration it is dropping.

**3. `medium` — with renewal on, a hung-but-alive holder is renewed forever; the TTL no longer bounds anything.**
`src/snagline/state.py:235-237`
`run()` is an unconditional `while True: sleep; renew_all()` with no per-section lifetime cap, and `renew_all` extends any registered lock without checking holder liveness. A detector that hangs inside `observe`/`finalize` holds the distributed lock for the process lifetime with no warning; old behavior at least broke the deadlock after one TTL. The code comment and PR body assert the TTL "bounds a stuck section", which is now inverted for the stuck-but-alive case. A renewal-count/total-hold cap (or a warning past N renewals) would restore a bound.

**4. `low` — `lock_renew_interval` accepts any positive value, so `1e-9` turns the renewer into a busy-loop that hammers Redis.**
`src/snagline/state.py:272-275,236`
Only `interval < 0` is rejected; `time.sleep(1e-9)` is effectively a tight loop — a spinning daemon thread even with an empty table, plus one extend script per live lock per pass. A sane floor (0.05–0.1 s) would close it.

**5. `low` — `SNAGLINE_STATE_REDIS_LOCK_TIMEOUT=inf`/`nan` pass the `<= 0` check and fail later, in `acquire()`.**
`src/snagline/state.py:360,260`
`float('inf')` reaches `int(inf*1000)` in redis-py → `OverflowError` escaping `episode_lock` into ingest; `nan` makes `next_try_at > stop_trying_at` always False → the worker blocks on acquire forever. Both should be rejected at parse time alongside the existing `ValueError` for unparsable strings (same NaN/inf class the sibling PRs #412/#387 are closing elsewhere).

Parthipashok04:
Pushed `eb21b48`, addressing every finding from the review and from CodeRabbit. Each change below has a test that I verified fails against the pre-fix `state.py` and passes against the new one.

**1. high — `except Exception` conflated transient redis errors with lock loss.** `renew_all` now distinguishes them: only redis-py's `LockError` (ownership loss — the key expired, was released elsewhere, failed the token check) unregisters the lock. A connection error or timeout is logged once per section and left registered, so the next pass retries. `test_redis_lock_renewal_survives_a_transient_error` drives this through the renewer with a `ConnectionError` and asserts the entry survives, renewal resumes, and the loss warning does *not* fire.

**2. medium — a renew pass could unregister a later holder's registration for the same episode.** `register` now returns the entry and `unregister` pops only when `self._renewals.get(episode_id) is expected`. The table is keyed by episode id alone, so this is what lets a pass tell its own stale entry from a live holder's. `test_renewer_keeps_a_later_holders_registration` reproduces the sequence (A expires and exits, B registers, the pass extends A and fails) and asserts B is still renewed.

**3. medium — a hung-but-alive holder was renewed forever, silently.** I kept renewing rather than hard-stopping: stopping would let the lock expire under a section that is still mutating detector state, which re-creates #326. The bound that survives is visibility — once a section has been extended past one full `lock_timeout`, it is named once with its episode id. `test_renewer_warns_when_a_section_outlives_the_ttl` covers it.

**4. low — interval accepted any positive value.** A positive `lock_renew_interval` now has a 0.1s floor (`1e-9` was a busy loop firing one extend script per live lock per pass) and must be less than `lock_timeout`, since at or above the TTL the first renewal lands after the lock has already expired.

**5. low — inf/nan slipped through.** `math.isfinite` guards `lock_timeout`, `lock_renew_interval`, and the `SNAGLINE_STATE_REDIS_LOCK_TIMEOUT` parse — `float("inf")` was reaching `int(inf * 1000)` in redis-py, and `nan` made the acquisition loop unbreakable.

**CodeRabbit's termination path:** `run()` waits on an `Event` rather than `time.sleep`, and `RedisStateBackend.close()` sets it and joins the thread, so a backend replaced by `Monitor.default` can be put down instead of living until process exit.

**One suggestion I did not take:** decoupling the acquisition wait from the TTL (`blocking_timeout=30` with a 300s TTL). A distributed lock has to be able to wait out a lawful holder's full TTL; a shorter wait would time out and run the detector section with no lock held, silently disabling the guarantee the backend exists for — worse than a slow ingest. The wait stays coupled to `lock_timeout`.

Verification: 857 passed, 3 skipped in the worktree; the pristine-PR baseline on this machine is 840 passed, 3 skipped, so the delta is exactly the 17 new cases with no regressions. `ruff check`, `ruff format --check`, and `mypy src` are clean; `state.py` coverage is 96%. The `test_server_tls`/`test_server_mtls` failures are the pre-existing local openssl issue and were not run.

Adhi1-2:
Heads-up: both issues this PR targets landed on master independently in #329 (commit `b817d83`), via a simpler approach than this PR's. I checked master's `RedisStateBackend.episode_lock` against each:

- **#325 (no lock held on timeout)** -- fixed. `acquire()`'s return is now honoured: a `False` from `blocking_timeout` raises `RuntimeError` *before* the yield, so the critical section can no longer run unlocked.
- **#326 (`LockNotOwnedError` out of the `finally`)** -- fixed differently. The TTL is now 300s and configurable (`SNAGLINE_STATE_REDIS_LOCK_TIMEOUT`), and a lost lock is logged rather than raised out of the caller's `ingest`/`end_episode`.

So the defect class is closed on master.

**What this PR still offers that master does not have:** background lock *renewal* while a section is still running. Master's answer is a longer fixed TTL plus a warning after the fact, so an episode whose ingest genuinely exceeds 300s still loses the lock mid-section -- logged, but mutual exclusion is already violated by then. This PR's watchdog keeps holding it for an unbounded section.

That is a real difference, but it is a design choice rather than a bug fix now, and it costs a thread per held lock. Two options:

1. Reduce this PR to *just* the renewal (drop the `RuntimeError` and TTL/logging hunks master already has), rebased on current master so it does not revert the landed work.
2. Close it, accepting master's longer-TTL-and-log as the contract.

Leaving it open for your call -- but note it currently branches from `b667169`, 15 commits behind master, so merging it as-is would revert `b817d83`'s state.py work. A rebase is required either way.

Adhi1-2:
Reviewing the renewer for #326 — the `unregister`-before-`release` ordering closes the *future*-pass race, but an in-flight pass that already snapshotted the entry still fires a false loss warning on a perfectly normal exit.

`renew_all` snapshots the table under the lock, then extends **sequentially** with the lock dropped:

```python
with self._lock:
    live = list(self._renewals.items())
for _episode_id, entry in live:
    try:
        entry.lock.extend(self._ttl, replace_ttl=True)
    except Exception as exc:
        if _is_lock_loss(exc):
            logger.warning("...was lost mid-section...mutual exclusion may be violated...")
            self.unregister(_episode_id, entry)
```

Sequence on a healthy episode:

1. Pass snapshots `live` while `ep` is still registered — `entry` is in the list.
2. The holder finishes normally: `unregister(ep, entry)` then `lock.release()` (which **deletes the key**).
3. The pass reaches `entry.lock.extend(...)`. The key is gone, so redis-py fails the token check and raises `LockNotOwnedError` → `_is_lock_loss` is True → it logs *"Redis episode lock for 'ep' was lost mid-section … mutual exclusion may be violated while the section keeps running"*.

The section is **not** still running — it just exited cleanly — so the warning is both false and self-contradictory. `unregister`-before-`release` doesn't help here: the concurrent pass is holding its own reference to `entry` from the pre-unregister snapshot.

This isn't microsecond-rare under load. The passes' `extend` calls are blocking network round-trips done in series, so an entry at position *N* is extended roughly *N × RTT* after the snapshot. On a backend with many concurrent episodes, that window is hundreds of ms to seconds, and any episode that completes normally inside it trips the warning — the exact signal operators are told to treat as a mutual-exclusion break.

The existing tests don't catch it: `test_..._loss_is_reported_while_the_section_is_still_running` forces expiry with the entry still registered (a real loss), and `test_renewer_keeps_a_later_holders_registration` uses a lock whose key genuinely expired, so its warning is arguably legitimate and the test never asserts the benign case stays quiet.

Suggested fix — re-check identity under the lock before reporting loss, so a holder that already unregistered is treated as a normal exit, not a violation:

```python
except Exception as exc:
    if _is_lock_loss(exc):
        with self._lock:
            still_live = self._renewals.get(_episode_id) is entry
        if not still_live:
            continue  # holder released between snapshot and extend; benign
        logger.warning("...was lost mid-section...")
        self.unregister(_episode_id, entry)
        continue
    ...
```

A genuine mid-section expiry leaves the entry registered (the holder hasn't exited), so `still_live` is True and the real warning still fires; a normal exit removes it first, so the false one is suppressed. A regression test would register an entry, `unregister` it, then run a pass whose `extend` raises `LockNotOwnedError` and assert no "mutual exclusion may be violated" line. The rest of the PR — transient-vs-loss split, identity-checked `unregister`, ceil-based stuck warning, inf/nan rejection — looks right to me.

Issue #400, opened 2026-09-20
bug: a colon-less detector key in strict restore raises IndexError, not the documented ValueError


abyyxhek:
A colon-less detector key in strict restore raises `IndexError`, not the documented `ValueError`.

`Monitor.restore_dict(..., strict_names=True)` builds its expected-composition list at monitor.py with:

```python
expected = [k.split(":", 1)[1] for k in sorted(dumped_detectors, key=_detector_slot_order)]
```

A caller-supplied snapshot whose detector key has no `":"` makes `split` return a one-element list, so `[1]` raises `IndexError: list index out of range`. The method's own docstring coaches callers to catch `ValueError` ("a caller that catches the ValueError never has to wonder which half of the snapshot landed"), and every other malformed-input path in the method raises `ValueError` — so this one escapes unhandled from a caller that did exactly what the docs say.

Reproduced against pristine master:

```python
monitor.restore_dict({"format_version": 1, "detectors": {"loop": {...}}}, strict_names=True)
# IndexError: list index out of range
```

The fix is to validate the key shape and raise `ValueError` naming the malformed key.

Issue #399, opened 2026-09-20
bug: Monitor._dispatch iterates the live sink list, so a concurrent remove_sink skips or duplicates alerts


abyyxhek:
`Monitor._dispatch` iterates the live sink list, so a concurrent `remove_sink` skips or duplicates alerts.

`_dispatch` (monitor.py) does `for sink in self._sinks:` with no snapshot, and `remove_sink` does a shifting `list.remove`. `list.remove` moves the tail left underneath the iterator, so the element after the removed one is either skipped (missed alert) or revisited after an append re-adds it (duplicate). `add_sink`'s docstring promises "Safe to call while ingests run from other threads; dispatch tolerates a concurrent append" — plain append is indeed safe in CPython, but remove is not, and only the append half is documented.

Reproduced against pristine master with two sinks `[a, b]`, one thread dispatching 300 risks and another doing `remove_sink(a); add_sink(a)` in a loop, `b.emit` yielding to widen the window: **54 of 60 rounds diverged**, with `a` receiving 0, 1, 2, 16, 288 of the 300 alerts.

The fix is a one-line `list(self._sinks)` snapshot in `_dispatch` (a sink added mid-dispatch simply waits for the next risk), plus correcting the `add_sink` docstring to state the guarantee precisely.

Issue #398, opened 2026-09-20
bug: the live-episode LRU cap evicts still-running episodes, silently destroying in-flight detection state


abyyxhek:
The live-episode LRU cap evicts episodes that are still running, silently destroying their in-flight detection state.

`Monitor.ingest` (monitor.py) touches the LRU and evicts over cap:

```python
# LRU by last-seen is safe: an active episode is by definition recently
# seen and cannot be evicted out from under itself.
```

That assertion is false. With N+1 concurrently-live episodes under a cap of N, the least-recently-seen id is by construction still active — its thread is simply between events. `_evict_episode` then acquires that episode's lock, drops its `_clocks` entry, and calls `reset()` on every detector, wiping the loop window, cascade counters and CUSUM baselines. Eviction is silent by design (no finalize risk, no log), so the lost detection is invisible.

Reproduced against pristine master with `Config(max_live_episodes=2)` and a `LoopDetector(repeat_threshold=3)`:

```
live before C: ['A', 'B']
A loop window before C: [sig, sig]      # 2 of 3 repeats seen
live after C:  ['B', 'C']
A loop window after C:  []              # destroyed while A is still running
A clock present after C: False
risks from A's completed loop: []       # A's loop completes and is NEVER detected
```

The monitor cannot distinguish "between events" from "gone forever", so eviction over cap is unavoidable — but the consequence (silent, total loss of in-flight detection for a running episode) should be observable, not invisible. This bites at exactly the scale the library targets: the default cap is 10_000 concurrent episodes, and any operator who sets a smaller cap hits it immediately.

Also worth fixing: the code comment quoted above is provably wrong and should be corrected.

Issue #397, opened 2026-09-20
bug: semantic_drift_tolerance is unvalidated -- 2.0+ silently deadens the goal-drift detector and NaN storms false alarms


abyyxhek:
`semantic_drift_tolerance` is unvalidated: `2.0` or higher permanently deadens the goal-drift detector with no warning, and `NaN` storms false alarms.

`Config.semantic_drift_tolerance` (config.py) is the noise threshold for the semantic goal-drift detector. It is consumed at `src/snagline/drift/goal_drift.py:222`:

```python
tol = self._cfg.semantic_drift_tolerance
signal = 0.0 if dev <= tol else min(1.0, (dev - tol) / (2.0 - tol))
```

`dev = max(0.0, 1.0 - cosine)` is bounded to `[0, 2]`, so the knob's valid domain is `[0, 2)`. Nothing checks it. PR #371 range-checks the sibling knobs `semantic_drift_cusum_h` / `semantic_drift_cusum_k` but not this one.

Reproduced against pristine master `b667169` with a fixed embedder returning an orthogonal vector (`dev = 1.0`), `semantic_drift_min_samples=2`, 40 steps:

| `semantic_drift_tolerance` | alarms |
|---|---|
| `0.3` (default) | 19 — correct |
| `2.0` / `2.5` | **0 — the detector is permanently inert, no log line, no error** |
| `nan` | **39 at score ~0.95 — a false alarm on essentially every step** |

The `2.0` case is worth noting: `2.0 - tol` is the divisor of `signal`, but `dev <= tol` is always true at `tol = 2.0`, so the division is never reached and there is no crash — the detector simply goes silent forever. A `NaN` tolerance makes `min(1.0, NaN)` return `1.0`, so every step scores a full-strength alarm.

An operator who sets `SNAGLINE_SEMANTIC_DRIFT_TOLERANCE=2` (or a config file with a wrong unit) gets a silently disabled tier-2 detector, which is the same fail-quiet shape as #331/#350. The fix is to range-check the knob at construction and in `resolve`: finite and `0.0 <= tol < 2.0`.

PR #401, opened 2026-09-20
fix: four unattended contract violations (#397, #398, #399, #400)

abyyxhek:
#397 semantic_drift_tolerance is the denominator-adjacent knob of the
    goal-drift signal and was never range-checked. Its valid domain is
    [0, 2) -- dev is bounded to [0, 2], so at or above 2.0 the comparison
    dev <= tol is always true, the signal is pinned at 0.0 and the detector
    is permanently inert with no log line; a NaN tolerance makes
    min(1.0, NaN) return 1.0, scoring a full-strength alarm on every step.
    Range-checked at construction and re-checked after env/file layering,
    matching _validated_stagnation's contract.

#398 The live-episode LRU cap evicted episodes that were still running.
    The comment claimed "an active episode is by definition recently seen
    and cannot be evicted out from under itself", which is false: with N+1
    concurrently-live episodes under a cap of N the least-recently-seen id
    is by construction still active -- its thread is between events -- and
    _evict_episode wipes its clock and detector windows. The monitor cannot
    distinguish that from a genuinely finished episode, so eviction stays
    fail-open, but it now logs the fault (deduplicated, naming the cap) so
    a too-small cap surfaces as a warning instead of unexplained missed
    detections. The false comment is corrected.

#399 _dispatch iterated the live sink list, so a concurrent remove_sink
    shifted the tail left underneath the iterator and skipped a sink that
    never stopped being registered, or handed an appended sink the risk
    already in flight. Dispatch now iterates a snapshot; the add_sink
    docstring, which over-claimed thread-safety for append only, states
    the guarantee precisely.

#400 A colon-less detector key under strict_names raised IndexError out of
    a method whose contract says a caller catching ValueError "never has to
    wonder which half of the snapshot landed". Malformed keys now raise
    ValueError naming the key.

18 regression tests; 12 verified failing against pre-fix master. Stock configurations never hit the new checks (all defaults are in range).

Summary of Changes

Please provide a clear and concise summary of the changes you've made.

Justification

Explain why this change is important and which use case benefits from this
change.

Kind of Contribution

Please remove what does not apply: Bug Fix, Performance Improvement,
New Feature, Cleanup, Documentation, Tests

Benchmark (for Performance Improvements)

If this change affects detection latency or overhead, include numbers from
`benchmarks/overhead_benchmark.py` before and after the change.

Unit Tests:

What unit tests were added or updated under `tests/`? Every new detector,
adapter, or sink behaviour should be covered by a focused test case.

Execution Tests:

How was this change verified end to end? For example, attach the monitor to
one of the examples in `examples/` or a real agent run and describe what
SNAGLINE reported.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for semantic drift tolerance values, including environment- and file-based configuration.
  * Improved episode-capacity handling with a deduplicated warning when active episodes are evicted.
  * Improved event delivery reliability when sinks are added or removed during dispatch.
  * Snapshot restoration now reports malformed detector keys with a clear validation error instead of an unexpected indexing error.


Cyrax321 (commented):
**Code Review — PR #401**

All four fixes land at the right layers (Config validation at construction + resolve + file/env layering, monitor-level observability/robustness) with correct inclusive/exclusive bounds ([0,2) verified: 2.0 rejected, 1.9999999999999998 accepted, NaN/inf rejected on every path) and no regression for previously-valid configs — the full 841-test suite passes at PR head `81fb35c`, with 12/18 new tests verified failing against pristine master semantics. The remaining findings are consistency gaps in the documentation and messaging the PR itself corrects elsewhere.

Findings, most severe first:

**1. `low` — the false LRU comment #398 disproves still stands verbatim in `Config.max_live_episodes` and the README.**
`src/snagline/config.py:513-516`, `README.md:307`
The PR corrects the "a genuinely active episode is by definition recently seen, so it cannot be evicted out from under itself. Eviction is silent" claim in `monitor.py`, but the same sentences remain in the Config field comment and README ("evicted silently (no finalize)"). A contributor reading either is told the exact misconception #398 disproved, with the corrected comment now saying the opposite one file over. One-line fixes.

**2. `low` — the eviction fault logs at ERROR and fires for the cap's designed use case, while its message advises raising the cap.**
`src/snagline/monitor.py:730-736`
Verified at PR head: `_evict_episode` logs via `_log_fault_once` → `logger.error`, and it fires for a genuinely dead leaked episode (the host never called `end_episode` — the scenario the cap exists for, `test_bounded_memory_repro`) as well as for the harmful live-eviction case. The message "raise max_live_episodes if episodes are concurrent" is the wrong advice for the leak case (raising it worsens memory retention), and the PR text says "surfaces as a warning" while the code emits ERROR. Deduped once per monitor, so noise is bounded and fail-open teardown is unchanged — a `logger.warning` plus wording that distinguishes the two cases would fit both.

**3. `low` — the #400 shape check enforces only the name half of `<slot>:<name>`.**
`src/snagline/monitor.py:919`
Keys with an empty or non-integer slot (`':loop'`, `'x:loop'`) pass the new validation and, via the name-suffix fallback, still apply state under `strict_names=True` despite the message claiming the key "must be '<slot>:<name>'" (verified: restoring a real loop dump under key `'x:loop'` with `strict_names=True` loads state and raises nothing). Pre-existing laxity on master, so this is an adjacent incompleteness of #400's contract rather than a new regression.

PR #402, opened 2026-09-21
fix(monitor): contain a malformed detector entry in restore_dict (#384)

### LINKED ISSUE #384: bug: one malformed detector entry aborts Monitor.restore_dict mid-loop, leaving detectors holding mutually inconsistent state
## Summary

`Monitor.restore_dict` applies each detector's snapshot with an unguarded `load(entry)` (monitor.py:918), and the `load_state` implementations it calls hard-subscript the snapshot payload. One malformed entry therefore raises `KeyError` / `ValueError` mid-loop, and the restore stops there — leaving the monitor half-restored: detectors that already loaded have the snapshot's state, the detector that raised keeps its live state, and the sink and time-axis restoration that follow the loop never run at all.

`snapshot()` writes atomically (tmp + `os.replace`), so a truncated file is not the trigger; the realistic ones are version skew across releases and a hand-edited snapshot. `detectors/base.py` documents that "restore-time errors are deliberately loud" — but loudness is only useful if the state is left untouched, which is what `strict_names=True` carefully guarantees *before* applying anything (monitor.py:882-892) and what the default tolerant path silently throws away.

## Reproduction

Against a fresh clone of master (`b667169`). The payload below is what an older release would have written before the re-fit fields existed, minus one key:

```python
mon2.restore_dict({"format_version": ..., "detectors": {
    "2:latency_anomaly": {"states": [[["ep", "tool"], {"n": 5, "mean": 10.0}]]},
}})
# -> KeyError: 'm2'
```

End-to-end, with 20 live events ingested first and a snapshot from an empty monitor:

```
restore_dict RAISED -> KeyError 'm2'
AFTER: loop detector state        -> {}               (loaded the empty snapshot: 20 episodes of live state gone)
AFTER: latency_anomaly detector   -> 20 live states   (raised, so it kept everything it had accumulated)
AFTER: _live_episodes             -> 20 entries       (cleared at monitor.py:1035, which the raise never reached)
```

The monitor now believes 20 episodes are live, has no loop-detector state for any of them, and does have latency state. `end_episode` on one of them resets nothing in the loop detector and something in the latency one. The `_live_episodes` rebuild at monitor.py:1033-1045 and the sink loop at monitor.py:931-936 also never ran.

For every detector I checked, the restore path is the fragile one — the same hard-subscript pattern, and the only place in the codebase that deliberately declines to fail open:

- `latency_anomaly._state_from_dict`: `raw["n"]`, `raw["mean"]`, `raw["m2"]`, `raw["cusum"]`, `raw["mu0"]`, `raw["sigma0"]`, `raw["frozen"]` (the re-fit fields beside them already use tolerant `.get()` precisely for this reason)
- `token_runaway.load_state`: `raw["n"]`, `raw["mean"]`, ... → `KeyError`
- `silent_abort.load_state`: `raw["step_id"]`, `raw["timestamp"]`, `raw["action_type"]`, `raw["error"]` → `KeyError`
- `side_effect_guard.load_state`: `for key, count in raw_counts` with `key[0]`, `key[1]` → `ValueError: not enough values to unpack`

## Suggested fix

Two layers, both small:

1. Per-entry tolerance in each `load_state` — skip or default a malformed entry instead of raising, mirroring the `.get()` treatment the re-fit fields already get. A bad entry should cost one detector's restored state for one episode, not the whole restore.
2. Per-detector containment in `restore_dict`, so one detector that still cannot parse its payload logs a warning and leaves itself untouched while the rest of the restore completes:

```python
try:
    load(entry)
except Exception as exc:
    logger.warning("snagline: snapshot for %s rejected: %s", matched_key, exc)
    continue
consumed.add(matched_key)
```

Option 2 alone is the important one: it bounds the damage of any future `load_state` that forgets the tolerant pattern, and it is what makes the "loud" error path honest — the caller still learns which slot failed, via the warning, without being handed a monitor whose components disagree with each other.

## Severity

Moderate. The trigger needs a mismatched or edited snapshot, but the outcome — a monitor whose detectors hold mutually inconsistent state, with no indication of it — can silently corrupt detection for the episodes involved.

Adhi1-2:
## What

`Monitor.restore_dict` applied each detector's snapshot with an unguarded `load(entry)` (`monitor.py:918`). Every `load_state` it calls hard-subscripts fields an older release did not write, so one malformed entry raised `KeyError` / `AttributeError` **mid-loop** — and the restore stopped right there:

- detectors that had already loaded held the snapshot's state;
- the detector that raised kept everything it had accumulated live;
- the sink restoration and the time-axis / live-episode rebuild that follow the loop **never ran at all**.

The monitor was left with components that disagreed about which episodes existed, with no indication of it. `snapshot()` writes atomically (tmp + `os.replace`), so a truncated file is not the trigger — version skew across releases and hand-edited snapshots are.

Reproduced against `b667169`:

```
restore RAISED KeyError: 'm2'
AFTER: loop detector state        -> {}               (loaded the snapshot: 20 episodes of live state gone)
AFTER: latency_anomaly detector   -> 20 live states   (raised, so it kept everything)
AFTER: _live_episodes             -> 20 entries       (cleared later in the method, which the raise never reached)
```

The monitor now believes 20 episodes are live, has no loop-detector state for any of them, and does have latency state. `end_episode` on one of them resets nothing in the loop detector and something in the latency one.

`detectors/base.py` documents that restore-time errors are "deliberately loud" — but loudness is only honest if the state is left untouched, which is exactly what `strict_names=True` guarantees *before* applying anything (`monitor.py:882-892`). The default tolerant path silently threw that guarantee away on the first bad payload.

## Fix

Per-detector containment in `restore_dict`: one entry that cannot parse logs its slot and the exception, keeps that detector on its live state, and lets the rest of the restore complete. The caller still learns which slot failed — via the warning — without being handed a monitor whose components disagree.

A rejected slot is now marked consumed, so it is not *also* reported below as an unknown-slot orphan. One bad entry deserves one message, not two, and the orphan warning describes a composition problem this is not.

I deliberately did **not** add per-entry tolerance to every `load_state`. The issue suggests it as a second layer, but containment alone is what bounds the damage of any future `load_state` that forgets the tolerant pattern — and it keeps each `load_state` honest about what it requires. With containment in place, a detector that hard-subscripts is loud once per restore rather than silently half-applied.

The sink restore loop was already safe: `DedupSink.load_state` is the only implementation and it is fully defensive (`dedup.py:173`), documenting "fail open to delivery, never raise into setup" and dropping malformed input with a warning.

## Scope

`src/snagline/monitor.py` — the `try`/`except` around `load(entry)` plus the `consumed.add` move. `tests/test_monitor_snapshot.py` — 4 tests. CHANGELOG. No behavior change beyond the loop body.

## Verification

Reproduced the `KeyError` and the inconsistent-state aftermath against `b667169` before the fix (output above).

- `pytest`: **875 passed, 2 skipped**, coverage **88.07%** (80% gate)
- `ruff check src tests`, `ruff format --check src tests CHANGELOG.md` — clean
- `mypy src tests --python-version 3.12` — clean. The repo pins `python_version = "3.10"`, which numpy 2.5.3's bundled stubs reject at their own `numpy/__init__.pyi:737` before any project file is checked — an environment artifact, not a project error; CI's own legs are unaffected.
- Mutation check: removed the containment and re-ran — **3 of the 4 new tests fail**, the three that pin the defect. The fourth (`test_strict_restore_rejects_before_any_load_runs`) is a regression guard that passes both before and after: containment must not soften the strict path, where a composition mismatch is a setup error that raises before any `load_state` runs.

Fixes #384.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved snapshot restoration resilience when a detector entry contains malformed or outdated state data.
  * Restoration now continues for valid detectors, sinks, and time-axis settings instead of stopping at the invalid entry.
  * The affected detector retains its current live state, and the rejected entry is reported with a single warning rather than being misclassified as an unknown snapshot entry.

* **Documentation**
  * Added an unreleased changelog entry describing the snapshot restoration fix.


Adhi1-2:
Acted on the Major finding on `load_state`: `ErrorCascadeDetector`, `LoopDetector` and `MeltdownDetector` now build every restored mapping — windows, scaler positions, streaks, fired flags, and the loop detector's near/cycle/stall families — into locals and publish them only once every field has parsed, so a snapshot rejected partway leaves the detector on its complete live state rather than a mix.

Regression added (`tests/test_monitor_snapshot.py`): builds live error-cascade state, snapshots it, loads a payload whose `windows` parses but whose `counts` does not, and asserts the post-failure `dump_state()` still equals the live one — comparing window **values**, not just keys. A second test confirms a well-formed snapshot still applies in full. Verified the atomicity test fails against the pre-fix source with exactly the half-applied mix.

Note the same shape exists in the loop and meltdown detectors, so the transactional fix is applied to all three rather than just the one this finding named.

Cyrax321 (commented):
**Code Review — PR #402**

The per-entry containment is well-engineered, with honest mutation-testing and a genuine atomicity improvement for `error_cascade`. But it was cut from a stale base (`b667169`) and conflicts with the containment fix already merged on master (09e5fd4, PR #329 for issue #324) — the PR body never mentions it — and the new warning's "detector keeps its live state" guarantee holds for only two of ~10 shipped detectors.

Findings, most severe first:

**1. `high` — stale base: master already contains #324's containment, and this PR conflicts with it.**
`src/snagline/monitor.py:918-932`
Master already has `try: load(entry) / except Exception:` (comment cites issue #324), the sink try/except with `consumed_sinks`/`orphaned_sinks` bookkeeping, its own tests in `tests/test_issues_321_327.py` (`test_restore_dict_skips_malformed_detector_entry`), and different warning text ("malformed snapshot entry for detector %r; ignored"). `gh pr view 402` → `mergeable: CONFLICTING`. The PR's tests assert "rejected" in the message while master's assert "malformed ... ignored", and the PR's improvement (consuming a rejected slot so it is not double-reported as an orphan) is real — but the author is unknowingly re-implementing a merged change and must reconcile, otherwise whichever shape wins breaks one side's tests.

**2. `high` — the "detector %r keeps its live state" claim is false for 6+ shipped detectors.**
`src/snagline/monitor.py:939`
Verified at the PR head: `token_runaway.load_state` starts `self._states = {}` then `s.n = raw["n"]`; `meltdown.load_state` starts `self._eps = {}` then `int(counts...)` inside the loop; `stagnation.load_state` starts `self._windows = {}` then `int(n)`; `loop.load_state` assigns `_windows` before `_counts`. A malformed entry in any of these still destroys (or half-applies) live state, yet the log tells the operator the safe outcome happened — for the exact inconsistency this PR exists to eliminate. Either make the transactional fix uniform (as #419 does for token_runaway/stagnation) or soften the message to "may keep its live state".

**3. `medium` — sink loads remain unguarded, so a raising sink aborts restore between detector loading and the time-axis/LRU rebuild.**
`src/snagline/monitor.py:958-963`
The PR justifies this with "DedupSink.load_state is the only implementation and it is fully defensive", but sinks are caller-supplied and `sinks/base.py` documents `load_state` as part of the sink snapshot protocol. A custom sink raising on a version-skewed entry aborts `restore_dict` after detectors applied but before `_clocks` and `_live_episodes` are rebuilt — the component-disagreement state the PR fixes for detectors, reintroduced one loop later. Master's #324 fix already guards this loop; the conflict resolution should adopt that shape.

**4. `medium` — `strict_names=True` callers lose the "reject loudly before anything is applied" contract.**
`src/snagline/monitor.py:882-892,924`
`strict_names` gates only the composition check; the try/except then wraps strict restores too, so a strict-mode caller whose slot 3 payload is hand-mangled now gets slots 0-2 restored, slot 3 live, sinks/clocks rebuilt — and no exception, only a log line. The docstring (`monitor.py:865-867`) still promises "a rejected restore leaves the monitor untouched", and the new strict test covers only composition mismatch, not payload malformation. Either keep `except` semantics for strict mode or update the docstring.

**5. `low` — `except Exception` swallows programming errors identically to data malformation, with no metric.**
`src/snagline/monitor.py:926`
A refactor-renamed attribute or a refactored `load_state` bug (TypeError/AttributeError) is contained the same as version skew, one log line per detector per restore, and no `Metrics` counter is bumped — dashboards keyed on self-observability show nothing while all detectors silently keep stale state for the process lifetime. The class-level contract comment at `monitor.py:800-808` ("malformed payloads raise rather than fail open... a monitor that silently starts blank is precisely the quiet misbehavior this feature exists to prevent") is left unchanged and now contradicted. The message does include `type(exc).__name__`, which mitigates; a rejected-restore counter and a `raise` on strict mode would fully restore the contract.

Adhi1-2:
Heads-up on scope, checked against current master: the containment half of this PR is no longer needed.

`restore_dict` on master now guards every `load(entry)` -- `src/snagline/monitor.py:928` (detectors) and `:956` (sinks) log and skip a malformed entry, and a rejected slot is marked consumed. That landed in b817d83, and I confirmed the repros from #384 and #324 both return cleanly instead of raising:

```
snagline: malformed snapshot entry for detector '2:latency_anomaly'; ignored
RESTORE: no exception raised
```

What master still does **not** have, and what this PR is now the sole carrier of, is the transactional `load_state`: building the windows/counts/streaks/fired flags into locals and publishing once, so a payload rejected partway cannot leave a detector half-restored. That is real remaining work and overlaps with #417, whose PR #419 does the same for `token_runaway` and `stagnation` -- worth a look before this rebase lands, so the two agree on the pattern instead of each inventing one.

Suggest dropping the `monitor.py` containment hunks and the CHANGELOG entry describing them (master's wording for the same behaviour differs), and retitling around the atomicity work so the diff review covers only what master is missing.

PR #404, opened 2026-09-21
fix(sinks): stop logging the destination URL, which is the credential (#390)

### LINKED ISSUE #390: bug: SlackSink and WebhookSink log the destination URL -- which is the credential -- on every failed POST, breaking the invariant PagerDutySink keeps
## Summary

On a failed POST, `SlackSink` and `WebhookSink` log the full destination URL:

```python
logger.exception(
    "snagline Slack sink POST to %s failed; ignoring (fail-open)",
    self._url,                       # slack.py:74-77, webhook.py:77-80
)
```

That URL *is* the credential. A Slack incoming-webhook URL embeds its secret as the final path segment (`https://hooks.slack.com/services/T…/B…/<secret>`), and webhook URLs frequently carry basic-auth credentials (`https://user:pass@host/`). A sink that cannot reach its destination is exactly the moment an operator goes looking in the logs — so this writes the secret into whatever log pipeline, ELK stack or stdout capture is attached, where it is then retained and indexed long after the alert is forgotten.

The codebase already knows the rule and follows it one file away: `PagerDutySink` logs `"snagline PagerDuty sink POST failed; ignoring (fail-open)"` with **no** routing key (pagerduty.py:91-93). `PagerDutySink` and `SlackSink` disagree about the same invariant.

## Reproduction

No setup beyond a destination that refuses the connection:

```python
from snagline.sinks.slack import SlackSink
SlackSink("https://hooks.slack.com/services/T000/B000/SECRET_TOKEN_HERE", timeout=1).emit(
    FailureRisk("ep", "s1", 0.9, "loop", "detail", 0.0)
)
```

```
ERROR:snagline:snagline Slack sink POST to https://hooks.slack.com/services/T000/B000/SECRET_TOKEN_HERE failed; ignoring (fail-open)
...
```

`logger.exception` additionally attaches the full traceback, which repeats the URL inside the stack frames.

## Suggested fix

Drop the URL from the message, matching the PagerDuty precedent. What an operator actually needs to identify the sink is already available without the secret — the sink type and, if useful, a redacted form:

```python
redacted = re.sub(r"//([^/@]+@)?", "//", self._url)   # strip userinfo
# or simply omit, as PagerDuty does
logger.exception("snagline Slack sink POST failed; ignoring (fail-open)")
```

Same for `WebhookSink`. Also worth noting: the sinks accept the URL at construction and keep it on `self._url`, so a `repr()` of the sink in any other diagnostic path leaks it too — a `__repr__` that shows the host but not the path would be the thorough version.

## Severity

Moderate, and it compounds with the redirect issue above: a misconfigured destination produces both a silently dropped alert and a credential written to the logs on the same step.

Adhi1-2:
## What

`WebhookSink` and `SlackSink` logged the full destination URL when a POST failed. That URL **is** the credential:

- **Slack** incoming webhooks embed the secret as the final path segment: `https://hooks.slack.com/services/T.../B.../<secret>`
- **Arbitrary webhooks** routinely carry basic auth: `https://user:pass@host/`

A sink that cannot reach its destination is exactly the moment an operator goes reading the logs — so every dead-endpoint error was handing the credential to whoever was already looking at a broken integration.

## Reproduction (pre-fix)

```python
sink = WebhookSink("https://alice:hunter2@hooks.example/alerts")
with patch(urlopen, side_effect=OSError):
    sink.emit(risk)   # -> "snagline webhook sink POST to https://alice:hunter2@hooks.example/alerts failed"
```

`logger.exception`'s traceback does **not** re-leak it — the traceback renders source lines, not argument values — so dropping the URL from the format string is a complete fix, not a partial one. (Verified, not assumed.)

## Change

New shared helper `redacted_destination(url)` in `sinks/base.py` — the existing home for shared sink plumbing. It returns `scheme://host:port/`, dropping path, query, fragment and userinfo, or `<invalid destination url>` for an unparseable destination. Both sinks use it for the failure log, and both gain a `__repr__` that uses it, since the default attribute dump would leak the same URL into any diagnostic capture.

This matches `PagerDutySink`, which already logs no routing key — the three network sinks now agree.

| File | Change |
|---|---|
| `src/snagline/sinks/base.py` | + `redacted_destination` |
| `src/snagline/sinks/webhook.py` | redacted failure log + `__repr__` |
| `src/snagline/sinks/slack.py` | redacted failure log + `__repr__` |

## Tests

New `tests/sinks/test_sink_base.py` covers the helper (basic auth, Slack path secret, query tokens, port retention, IPv6, degenerate inputs). New cases in both sink test files assert a failed POST logs the host but never the secret, and that `repr` is clean.

All 4 new sink tests fail on the pre-fix tree (verified in a scratch worktree) and pass after. Full gate on this branch:

- `pytest`: **860 passed, 1 skipped, 4 failed** — the 4 are the pre-existing environmental baseline on this dev machine (a sibling Continuum checkout on disk, and `openai`/`anthropic` installed in the venv so the without-SDK no-op tests cannot be exercised), none related to sinks.
- `mypy --no-incremental src tests`: **11 errors in 2 files** (`auto/openai.py`, `auto/anthropic.py`) — the known local baseline, all in optional-SDK modules invisible to CI.

Fixes #390.


Cyrax321 (commented):
**Code Review — PR #404**

The fix correctly removes the credential from both sinks' failure logs and `repr`s; the new `redacted_destination`/`describe_failure` helpers are robust on tricky inputs (unbalanced IPv6 brackets, userinfo with `@`, port-only netloc, `HTTPError`/`URLError` naming), and the 26 new/changed sink tests pass on the PR head. However, the cross-file trace shows the same leak class survives in sibling paths the PR does not touch — including the Monitor's halt-webhook fail-open log and the CLI hook forward, both of which print the raw URL on the exact failure lines an operator reads.

Findings, most severe first:

**1. `high` — the Monitor's halt-webhook fail-open log still embeds the raw `halt_url`, the identical credential leak this PR fixes for sinks, one file over.**
`src/snagline/monitor.py:618-622`
`halt_url` is validated for scheme only (`monitor.py:305-308`), so it may carry basic-auth credentials or a secret path segment. When the endpoint dies, the except block calls `self._log_fault_once(f"halt webhook {self._halt_url} failed; continuing (fail-open)")`, which `logger.error`s the full URL — exactly when the operator reads it. The new `redacted_destination` helper in `sinks/base.py` is never applied here (monitor imports only `AlertSink` from sinks.base), and no test in `tests/test_enforcement_policy.py` asserts the fault log omits the URL. Route the message through the same helper.

**2. `high` — `snagline hook --url` prints the raw destination URL *and* `str(exc)` (which embeds the URL again) to stderr.**
`src/snagline/cli.py:650-653`
Verified: the handler prints `f"snagline hook: forward to {args.url} failed: {exc}"`. With a credential-bearing URL (`https://alice:hunter2@hooks.example/alerts`) and a dead endpoint, a `URLError("no host given: https://alice:hunter2@...")` writes the credential twice in one line to stderr, which CI and hook logs capture. Same issue class as #390; the PR's sink-only scope misses this path.

**3. `low` — the `snagline serve` startup banner prints the full `halt_url`, including userinfo credentials.**
`src/snagline/cli.py:1048-1051`
`halt_url` is a `Config` field resolved from env/file layering (not just argv, which `ps` would expose anyway), so `snagline serve: halt forwarding enabled -> https://user:pass@host/halt ...` writes captured credentials at every startup when policy is `halt_webhook`. Deliberate one-time banner, hence low — but the same redaction applies.

**4. `low` — `describe_failure` discards all exception detail, including safe, actionable reasons.**
`src/snagline/sinks/base.py:64`
`URLError` wrapping `socket.gaierror("[Errno -2] Name or service not known")` now logs as `failed (URLError)`, so a dead endpoint's DNS failure, refusal and timeout are indistinguishable. Verified that socket-level URLError reasons do *not* embed the URL, so including `getattr(exc, "reason", None)` would be safe and diagnosable — the all-or-nothing choice is a documented tradeoff, not a bug, but worth revisiting.

**5. `low` — the stated rationale for adding `__repr__` is factually wrong: plain classes have no "default attribute-dump repr".**
`src/snagline/sinks/webhook.py:54`
`WebhookSink`/`SlackSink`/`PagerDutySink` are plain classes with explicit `__init__` (no `@dataclass`), so the pre-PR repr was `<WebhookSink object at 0x...>` with no attributes — no leak existed, and `Monitor.snapshot_dict` only persists `dump_state()` when a sink defines it (`monitor.py:818`). The added `__repr__` is harmless and genuinely more diagnostic; the code comment and CHANGELOG claim are what's wrong.

**6. `low` — no sink-level test pins the `HTTPError` path end-to-end.**
`tests/test_webhook_sink.py:171`
The sink failure tests use `OSError` and a hand-built `URLError`; only the helper-level test covers `HTTPError`. Verified that `urlopen` raises `HTTPError(req.full_url, ...)`, so `e.filename` *is* the credential-bearing URL — a regression logging `str(exc)` or `exc.filename` on the sink path would pass the current tests. One pinning test per sink would guard the actual credential field.

PR #405, opened 2026-09-21
fix(server): reject a non-positive --max-body-bytes instead of 413-ing everything (#394)

### LINKED ISSUE #394: bug: snagline serve --max-body-bytes 0 is accepted and rejects every POST with 413, silently dropping all telemetry
## Summary

`--max-body-bytes` is accepted with no validation or clamping (cli.py:305-310), and `http_server` stores it unmodified:

```python
_Handler.snagline_max_body = max_body_bytes        # http_server.py:915
_Handler.snagline_risks = deque(maxlen=max(1, max_risks))   # line 913, clamped
```

The sibling knob on the next line is clamped; this one is not. `do_POST` compares `length > self.snagline_max_body` (http_server.py:636), so `--max-body-bytes 0` makes **every** POST answer 413 and all telemetry is silently rejected — the sidecar starts cleanly, reports healthy on `/health`, and drops 100% of inbound events.

## Reproduction

```
$ snagline serve --max-body-bytes 0
$ curl -s -X POST -H 'Content-Type: application/json' -d '{"episode_id":"e","step_id":"s",...}' \
    http://127.0.0.1:8787/events
{"error": "payload too large"}     # for every event, forever
```

A negative value behaves identically, since the comparison is a strict `>`.

## Why 0 is the natural mistake

The neighbouring flag documents the opposite convention: `--episode-ttl-seconds` says "0 disables" (cli.py:324-333), and `--read-timeout`/`--max-versions` elsewhere use 0 or `None` as the "unlimited/off" sentinel. An operator who wants "no body limit" is most likely to write `--max-body-bytes 0` — the one value that rejects everything. There is no `None` escape hatch and no startup warning.

## Suggested fix

Clamp at the same place `max_risks` is, so the two knobs agree:

```python
_Handler.snagline_max_body = max(1, max_body_bytes) if max_body_bytes else 1_000_000
```

Better, follow the `--read-timeout` precedent and let `None` mean "no limit", validated in the CLI's arg parsing so `0` and negatives are rejected at startup with a clear message instead of silently inverting the knob's meaning at runtime.

## Severity

Low, but a clean instance of the silent-misconfiguration class the codebase has been closing elsewhere (#333, #353): nothing errors, nothing logs, and the monitoring product quietly observes nothing.

Adhi1-2:
## What

`snagline serve --max-body-bytes 0` was accepted with no validation. `do_POST` compares the declared length against the cap with a strict `>` (`http_server.py:636`), so a cap of 0 rejects **every** POST:

```
$ snagline serve --max-body-bytes 0
$ curl -s -X POST -H "Content-Type: application/json" -d {episode_id:...} http://127.0.0.1:8787/events
{"error": "payload too large"}     # for every event, forever
```

The sidecar starts cleanly and answers `GET /health` 200, so nothing errors and the monitoring product quietly observes nothing. Reproduced pre-fix: every POST → 413, `/health` → 200.

## Why 0 is the natural mistake

The neighbouring knob, `max_risks`, is clamped with `max(1, ...)` at the same site — this one was not. And `--episode-ttl-seconds` on the very next line documents **0 disables**, so 0 is exactly what an operator reaching for "no body limit" would type. There is no `None`/unlimited value on the flag, and a negative behaves identically.

## Change

Two layers, both loud:

- **CLI** (`cli.py`): `--max-body-bytes <= 0` exits 2 with a message naming the flag, before the banner claims a working server — same shape as the existing `--keyfile requires --certfile` check above it. Help text now states there is no unlimited value and that 0 does not mean off.
- **Server** (`http_server.py`): `make_handler` raises `ValueError` on a non-positive cap, next to the `max_risks` clamp. Library callers who never touch the CLI cannot build a green-but-deaf server either.

I chose **reject** over the issue's suggested clamp: silently substituting the default is the same silent-misconfiguration class the issue is about, and the codebase has been moving the other way (#317, #353 — out-of-range knobs are a startup error naming the knob).

| File | Change |
|---|---|
| `src/snagline/cli.py` | reject before banner + help text |
| `src/snagline/server/http_server.py` | raise in `make_handler` |

## Tests

New `tests/server/test_max_body_bytes_394.py` covers the server layer (0, -1, -1_000_000 across both `make_handler` and `make_server`) plus a **positive control**: a small-but-positive cap still serves the 413 path, so the guard did not break legitimate over-cap rejection. New parametrized cases in `tests/test_cli.py` assert the CLI exits 2, never starts `serve()`, and mentions both the flag and the no-unlimited-value trap.

All 7 new tests fail on the pre-fix tree (verified in a scratch worktree); the positive control passes both ways. Full gate on this branch:

- `pytest`: **868 passed, 1 skipped, 4 failed** — the 4 are the pre-existing environmental baseline on this dev machine, none related to this change:
  - `test_real_actionledger_flags_review` — a sibling Continuum checkout on disk with an incompatible `MiniStorage`.
  - `test_instrument_{openai,anthropic}_without_sdk_is_safe_noop` — the SDKs are installed in this venv, so the without-SDK no-op path cannot be exercised.
  - `test_cli_mentions_read_timeout_next_to_max_body` — shells out to a `snagline` binary that is not installed next to this interpreter; I ran `python -m snagline.cli serve --help` directly and confirmed its three assertions still hold with the new help text.
- `mypy --no-incremental src tests`: **11 errors in 2 files** (`auto/openai.py`, `auto/anthropic.py`) — the known local baseline, in optional-SDK modules invisible to CI.

Fixes #394.


Cyrax321 (commented):
**Code Review — PR #405**

Careful, well-tested fix: `--max-body-bytes <= 0` is now rejected loudly at both entry points (CLI exits 2 naming the flag; `make_handler` raises `ValueError` propagated through `make_server`/`serve`), with a positive 413 control test. The PR also bundles a coherent credential-redaction change (#390). Two residual gaps: a NaN hole that silently disables the 413 limit, and an ordering nit where `make_server` loads TLS files before the cap check.

Findings, most severe first:

**1. `medium` — the guard is not NaN-safe: a NaN cap passes validation and silently disables the 413 limit instead of raising.**
`src/snagline/server/http_server.py:915`
`max_body_bytes <= 0` is `False` for `float('nan')`, so `make_handler(monitor, max_body_bytes=float('nan'))` succeeds (verified) and `do_POST`'s `length > self.snagline_max_body` (http_server.py:636) is False for every length — the server runs green with *no body cap at all*, a worse silent failure than the 413-everything bug this PR fixes. A library caller computing the cap (env-derived or math-derived) can land here. The neighbouring `_resolve_episode_ttl` already does `if not math.isfinite(value)` (http_server.py:234); the same shape on this knob closes it. (`inf` also slips past "no unlimited value", though it at least keeps rejecting large bodies.)

**2. `low` — `make_server` loads TLS cert/key files before running the cap check, so a misconfigured cap is masked by a TLS error.**
`src/snagline/server/http_server.py:1045-1046`
`tls_context = _resolve_ssl_context(...)` executes before `make_handler(...)`, so `make_server(monitor, certfile='/nonexistent.pem', max_body_bytes=0)` raises `FileNotFoundError` instead of the intended `ValueError("max_body_bytes must be positive...")` (verified). A caller logging the startup error gets a misleading "missing cert" diagnosis for what is the #394 misconfiguration. Swapping the two statements puts the cap error first; the CLI path is unaffected (its check at `cli.py:1043` runs before `serve()`).

**3. `low` — a positive fractional cap is accepted and creates a server that rejects every non-empty POST.**
`src/snagline/server/http_server.py:915`
The annotation is `int` but is not enforced: `make_handler(..., max_body_bytes=0.5)` is accepted (verified), and `1 > 0.5` makes a 1-byte body 413 — the same deaf-server failure this PR exists to prevent, reached by a float-typed caller. The neighbouring `_resolve_episode_ttl` validates `isinstance` and finiteness, showing the codebase's established pattern for float-prone knobs; an `isinstance(value, int)` (or integral) check would match it.

**4. `low` — `--help` text hard-codes an issue number into user-facing output.**
`src/snagline/cli.py:311`
"…(issue #394) — unlike --episode-ttl-seconds, 0 does not mean off" appears in `snagline serve --help`; adjacent flags keep issue refs in comments/changelog. Cosmetic and consistent-ish with `--cooldown-seconds`, which also cites an issue — nit, not defect.

PR #406, opened 2026-09-21
fix(detectors): seed the inferred scaler position in scaled-window restore (#403)

### LINKED ISSUE #403: bug: restore sizes scaled windows from the inferred position but forgets to seed _counts, so the first post-restore observe re-truncates (or leaves meltdown stuck wide)
## Summary

PR #286 fixed issue #268 by sizing restored windows at the *effective* scale rather than the base. But the fix computes the scaler position and uses it only to size the deque — it rebuilds `_counts` from the snapshot's `counts` dict alone. An episode present in `windows` but absent from `counts` therefore restores at the right width, and then loses it on the very next `observe`, because the live code reads `_counts.get(ep, 0)` to compute the target and sees `0`.

The `load_state` sites claim tolerant `.get()` handling for exactly this shape ("pre-#92 snapshots carry no scaler positions") — the tolerance is half-applied. It sizes the window as if the position were known and then forgets it.

Two detectors, opposite symptoms, one cause:

| Detector | Restored window | After the restore completes | Effect |
|---|---|---|---|
| `MeltdownDetector` | too wide | **stuck** — `push` pops at most one item per call, so an already-oversized window never shrinks back to target | entropy over the wrong sample size; bit-tuned thresholds misfire |
| `ErrorCascadeDetector` / `LoopDetector` | correct width | **truncated on the first observe** — `n = 0 + 1` → `target = base`, refits the deque down to base | restored history silently discarded |

## Reproduction (meltdown — window stuck wide)

Against `b667169`:

```python
cfg = Config(meltdown_window_size=6, window_scale_steps=10, max_window=24)
d = MeltdownDetector(config=cfg)
for i in range(40):
    d.observe(...)          # window grows to the effective 24

snap = d.dump_state()
snap["counts"] = {k: v for k, v in snap.get("counts", {}).items() if k != "ep"}

d2 = MeltdownDetector(config=cfg)
d2.load_state(snap)
len(d2._eps["ep"].window)   # 18  -- sized from the inferred position
"ep" in d2._counts          # False
for i in range(20):
    d2.observe(...)
len(d2._eps["ep"].window)   # 18  -- still 18, target is base 6, never recovers
```

`push` (`meltdown.py:56`) appends one and pops at most one, so an oversized window shrinks by zero net items per step and stays oversized for the rest of the episode.

## Reproduction (error_cascade — history discarded)

```python
cfg = Config(cascade_window_size=4, window_scale_steps=10, max_window=16)
# 40 steps -> live width 16, counts["ep"] == 40

d2.load_state(snap_without_counts)
len(d2._windows["ep"])      # 8   -- correctly wider than base
d2.observe(one_more_event)
len(d2._windows["ep"])      # 4   -- restored history gone in one step
```

## Trigger

Not exotic — any snapshot whose `counts` lacks an episode present in `windows`:

- a pre-#92 snapshot (the case the `.get()` tolerance exists for);
- a snapshot that crossed a version boundary where `counts` was added;
- a partial or hand-edited payload.

`Monitor.restore` now contains a malformed entry (#384, PR #402), so these do not crash the restore — they silently apply the wrong window.

## Suggested fix

Seed `_counts[ep]` with the position `load_state` just inferred, instead of rebuilding it from the snapshot alone:

```python
self._counts = {}
for ep, keys in state.get("windows", {}).items():
    n = int(state.get("counts", {}).get(ep, len(keys)))
    ...
    self._counts[ep] = n
```

That closes both directions at once: the restored detector picks up where its window says it is, and the first post-restore observe targets the effective size instead of the base. `LoopDetector` has the same pattern and needs the same treatment.

Separately, `MeltdownDetector.push` should pop in a loop (`while len(...) > maxlen`) rather than once — a window that is somehow oversized for any other reason is currently stuck for life.

Adhi1-2:
## What

The follow-on to #268. PR #286 fixed restored windows being sized at the base by computing the auto-scaler position from the payload and sizing the deque from it — but it used that position only to size the deque. The scaler's own `counts` dict is rebuilt from the payload alone, so an episode present in `windows` but absent from `counts` restores at the correct width and then immediately loses it, because `observe` reads the counts dict to pick its next target and finds nothing.

The `load_state` sites claim tolerant `.get()` handling for exactly this shape ("pre-#92 snapshots carry no scaler positions"). The tolerance is half-applied: it sizes the window as if the position were known and then forgets it.

Three detectors, two symptoms, one cause:

| Detector | Restored width | After the restore completes |
|---|---|---|
| `MeltdownDetector` | too wide | **stuck** — `push` popped at most one item per call, so an oversized window never shrank back, and entropy was computed over a sample the bit-tuned thresholds were not written for |
| `ErrorCascadeDetector` / `LoopDetector` | correct | **truncated on the first observe** — `n = 0 + 1` targets the base, refitting the deque down and discarding the restored history |

Reproduced against `b667169` before the fix:

```
meltdown     live 24 -> restored 18 -> still 18 after 20 observes (target 6, never recovers)
error_cascade live 16 -> restored  8 -> 4 after ONE observe
```

## Fix

`load_state` seeds the position it inferred into the scaler dict alongside sizing the window — one line per detector, plus the same treatment for the near/cycle families in `LoopDetector`, which carry their own positions. A position that ships without a window still survives, and a position that ships with one stays authoritative.

`MeltdownDetector.push` now pops `while` over rather than once. A call that appends one and pops at most one can only ever hold a too-wide window at its current width, so any caller that hands it an oversized window pins the wrong width for the life of the episode.

## Scope

`src/snagline/detectors/meltdown.py` (seeding + the `while` loop), `src/snagline/detectors/error_cascade.py` (seeding), `src/snagline/detectors/loop.py` (`_restore_scaled_windows` extracting the shared shape — the three families were byte-identical, and a helper keeps the fix in one place), `tests/test_scaled_window_restore.py` (6 tests), CHANGELOG.

`test_restore_without_counts_stays_backward_compatible` already existed and asserted the restored state; it never observed afterward, which is why the truncation went unnoticed. The new tests observe.

## Verification

Reproduced both symptoms against `b667169` before the fix (output above); both recover after.

- `pytest`: **881 passed, 2 skipped**, coverage **88.13%** (80% gate)
- `ruff check src tests`, `ruff format --check src tests CHANGELOG.md`, `mypy src tests --python-version 3.12` — clean. The repo pins `python_version = "3.10"`, which numpy 2.5.3's bundled stubs reject at their own `numpy/__init__.pyi:737` before any project file is checked — an environment artifact, not a project error; CI's own legs are unaffected.
- Mutation check:
  - removed the seeding in all three detectors — the error_cascade and loop tests fail, pinning the truncation. The meltdown seeding mutation is not directly pinned by these tests: once the scaler is seeded its `load_state` no longer produces an oversized window, so that detector's restored width is already correct.
  - reverted the `while` to an `if` — `test_oversized_window_can_shrink_back_to_target` fails. That test drives `push` directly because `load_state` now sizes the window correctly by construction; the loop is the guard that keeps the invariant honest for any other caller.
- Backward compatibility unchanged: `test_restore_without_counts_stays_backward_compatible` and the three #268 round-trip tests still pass.

Fixes #403.


Cyrax321 (commented):
**Code Review — PR #406**

Well-executed fix: the inferred scaler position is now seeded for all four scaled window families in `loop`/`error_cascade`/`meltdown` (near/cycle via the shared helper), the meltdown `while` trim is correct and count-consistent, and the #402 containment interplay is handled transactionally in loop and meltdown. Quality is high; the two substantive findings are that the atomicity claim is not applied uniformly and one scaling detector is left with the pre-fix restore shape.

Findings, most severe first:

**1. `medium` — `error_cascade.load_state` is not transactional despite the PR's blanket claim.**
`src/snagline/detectors/error_cascade.py:160-176`
It assigns `self._windows = {}` and `self._counts = {}` *before* the parse finishes; `int(counts.get(ep, len(flags)))` (164) and the `_consecutive`/`_fired` comprehensions (173-176) can still raise. `loop.py:433-475` and `meltdown.py:184-219` build locals and publish at the end; `error_cascade` publishes incrementally, and no test covers its atomicity (the PR added atomicity tests only for loop and meltdown). Failure scenario: `Monitor.restore_dict` catches the raise and logs "detector keeps its live state" (`monitor.py:938`), but the detector actually holds snapshot windows/counts with live `_consecutive`/`_fired` — a live `_fired[ep]=True` over restored history suppresses the cascade alert the snapshot was kept for. The CHANGELOG entry "Detector load_state is now transactional" is false for this detector.

**2. `medium` — `StagnationDetector` is the fifth scaling detector and is left with the pre-fix shape.**
`src/snagline/detectors/stagnation.py:200-218`
It scales windows (`effective_window_size` at 121/214) but its `load_state` neither infers nor seeds the scaler position: `self._counts` reads positions only from the payload, so a window whose episode is missing from `counts` gets `self._counts.get(str(ep), 0)` = 0, making `target = base` and truncating flags to `[-base:]` — the same "window present, counts absent" shape #403 fixes in the other three. Impact is restore-fidelity and cadence loss (the window regrows from base while `stale_windows` is carried over, so patience can complete against a much shorter sample) rather than a wrong alert — but `dump_state` carries enough flags to infer a position, and the PR's stated rule ("the inferred position must seed the counts, not merely size the window") is not applied here.

**3. `low` — sink `load_state` calls in `restore_dict` remain unguarded, after detectors are already restored.**
`src/snagline/monitor.py:959-963`
The #384 containment wraps detector `load(entry)` calls but not the sink loop. A custom sink whose `load_state` raises aborts the restore *after* all detector state landed, leaving detectors on snapshot state with no clocks restored — the same component-disagreement outcome the PR's containment exists to prevent. Benign today (DedupSink is documented fail-open), but it is a latent asymmetry for third-party sinks.

**4. `low` — the inference+seed rule now exists in three copies.**
`src/snagline/detectors/loop.py:419-430`, `src/snagline/detectors/error_cascade.py:164-172`, `src/snagline/detectors/meltdown.py:198-216`
The helper was extracted to keep the fix "in one place" for the loop families, but error_cascade and meltdown duplicate the same `counts.get`/`setdefault` shape inline. A future change to the inference rule applied to the helper silently misses the other two, re-introducing the #403 class of drift between detectors that share one windowing module. A shared helper in `windowing.py` would hold the rule once.

PR #407, opened 2026-09-21
fix(server): gate mutating POSTs on content type and request origin (Fixes #388)

### LINKED ISSUE #388: security: the sidecar's mutating POST endpoints check the token but never the request origin, so a visited page can forge cross-site writes against a token-less localhost serve
## Summary

The sidecar's mutating POST endpoints authenticate the *token* but never check *where the request came from*. `do_POST` (server/http_server.py:620) gates only on `_authorized()`, and inspects no `Origin`, `Sec-Fetch-Site`, `Referer` or `Host` header and requires no `Content-Type`. `snagline serve` defaults to `--auth-token None` — "Unset means all endpoints are open" (cli.py:277-283) — with only a stderr warning. A web page the operator visits can therefore issue a cross-site POST to a token-less localhost sidecar as a CORS "simple request" (no preflight), and the sidecar honours it. Companion to #375, which is the same class of gap on the token check itself.

## Reproduction

Stock configuration, fresh clone of master (`b667169`):

```
snagline serve   # 127.0.0.1:8787, no --auth-token
```

Any web page then does:

```js
fetch("http://127.0.0.1:8787/episodes/end", {
  method: "POST",
  headers: {"Content-Type": "text/plain"},   // simple request -> no preflight
  body: '{"episode_id": "the-episode-you-are-monitoring"}',
});
```

`Content-Type: text/plain` is CORS-safelisted, so the browser sends it without an OPTIONS preflight. The sidecar routes it (http_server.py:647-653) into `_post_episodes_end` → `Monitor.end_episode(episode_id)` (http_server.py:764), which calls `reset(episode_id)` on **every** detector for that episode. The loop detector's window, the cascade counters, and the latency/token CUSUM baselines are all discarded — and because CUSUM baselines are frozen, that episode restarts its warm-up from zero. The attacker cannot read the response (no CORS headers are emitted), but the destructive write lands regardless.

`POST /events` is the same shape in the other direction: arbitrary fabricated telemetry drives the halt/pause policy and the `/metrics` gauges.

The auth gate does not close this even when a token *is* set, because a simple cross-site request carries no `Authorization` header at all — it 401s, which is the correct outcome, but the failure is silent from the operator's side and there is no log line distinguishing "cross-site forgery attempt" from any other 401.

## Why localhost is not sufficient

Binding to 127.0.0.1 limits who can *reach* the port, but the browser is inside that boundary: any page loaded over the public internet can address `http://127.0.0.1:8787`. The private-network request is still sent by current Firefox and Safari, and Chrome sends it (with console warnings) in most configurations. The feature's own design — a local sidecar an operator inspects while browsing — is exactly the deployment where this is reachable.

## Suggested fix

Pick whichever subset fits, in rough order of value:

1. **Require a non-simple content type.** Reject POSTs whose `Content-Type` is not `application/json`. A forged cross-site request cannot use `application/json` without triggering a preflight, which the sidecar can then answer with a CORS denial. This is the smallest change that closes the hole and does not affect any documented client (the shipped sinks and the Claude Code bridge all send JSON).
2. **Check `Origin` / `Sec-Fetch-Site`** when present, and reject `cross-site`.
3. **Log a distinct warning** for 401s that carry a cross-site `Origin` or `Sec-Fetch-Site` so an attempt is visible rather than indistinguishable from a mistyped token.
4. Consider emitting a minimal `Access-Control-Allow-Origin` of `null` (or no ACAO at all, which is already the case) plus `Vary: Origin`, so the absence is explicit rather than incidental.

## Severity

High when the token is unset (the documented default for local use) — a visited page can silently destroy an in-flight episode's detection state or inject telemetry that drives an enforcement decision. With a token set it is closed by the 401, which is why I'd still file #1/#2 as hardening rather than relying on operators to always set one.

Adhi1-2:
## Summary

`snagline serve` authenticates the **token** but never the **sender**. With the default no-token configuration (`--auth-token None`, \"unset means all endpoints are open\"), any page the operator is visiting can issue a cross-site POST to the loopback sidecar as a CORS *simple request* — no `OPTIONS` preflight, so the browser sends it and the write lands (#388). The attacker cannot read the response (no CORS headers are emitted), but the destructive write lands regardless.

This is the companion to #375 (the timing leak in the token check itself): same class of gap, on the other side of the gate.

## Impact

Verified against `b667169` with the issue's reproduction — stock config, no token:

```js
fetch("http://127.0.0.1:8787/episodes/end", {
  method: "POST",
  headers: {"Content-Type": "text/plain"},   // simple request -> no preflight
  body: '{"episode_id": "the-episode-you-are-monitoring"}',
});
```

`end_episode` calls `reset(episode_id)` on **every** detector, so the loop detector's window, the cascade counters, and the frozen latency/token CUSUM baselines are all discarded — the episode restarts its warm-up from zero. `POST /events` is the same shape in the other direction: fabricated telemetry drives the halt/pause policy and the `/metrics` gauges.

## Fix

`POST` is now gated on two things a forged cross-site request cannot supply together:

1. **A JSON content type.** The CORS-safelisted types (`text/plain`, `application/x-www-form-urlencoded`, `multipart/form-data`) are the only ones a cross-site `fetch` can send without a preflight, so requiring `application/json` forces one — and this server answers any preflight with no `Access-Control-Allow-Origin`, so the browser never sends the POST at all. A bad type gets **415**. Every shipped client already sends JSON (the three network sinks, `snagline hook`, and the documented `curl` recipes).
2. **A same-site origin, when the client declares one.** `Sec-Fetch-Site` is authoritative when present; otherwise `Origin` is compared against the `Host` the request arrived on, and `Origin: null` (sandboxed iframe / `file://`) counts as cross-site. Non-browser clients — curl, urllib, a sink — send neither header, so a **missing declaration is not an error**; a request that *declares* itself cross-site gets **403**.

An `Origin` with no `Host` to compare against is now treated as cross-site rather than same-site (self-review, see below).

Both refusals log the origin, so an attempt is distinguishable from a mistyped token instead of vanishing into the general 401 noise — the issue's point 3.

`GET` is untouched: a cross-site GET can be issued but cannot read the response without CORS headers, so an origin check there would add nothing and would break a status-page embed.

## A fail-open inside the gate, found in self-review

The `Origin`/`Host` comparison cannot run when `Host` is absent, and that arm read the request as **same-site** — so a request that had just declared itself cross-site was waved through:

```python
host = self.headers.get("Host")
if not host:
    return None          # treated as same-site, so the 403 never fires
```

A `Host`-less POST with `Origin: https://evil.example` against a token-less sidecar returned 202 and the write landed. A browser always sends `Host` (mandatory in HTTP/1.1), so this is **not reachable from a page** — only an HTTP/1.0 sender or a proxy that strips it. But the gate's premise is that an origin it cannot verify is a cross-site one, and a security gate does not get to fail open because the sender is unlikely to cooperate. The arm now returns the declared origin.

## Test changes

14 new tests in `tests/test_http_server.py`. Verified with a pre-fix overlay (new tests copied over unmodified upstream source): **9 fail pre-fix, all 14 pass after**. The 5 that pass on both sides are the guards — the origin/content-type gates must not lock out `curl`, a same-origin browser POST, or a charset-suffixed content type.

The `Host`-less case needs a raw socket: `http.client` always sets `Host`, so nothing built on it can reach the arm.

Also updated the existing suites that POST without declaring a content type (`tests/test_http_server.py`, `tests/server/test_episodes_end.py`, `tests/server/test_content_length_129.py`) to send the documented header — `urllib` defaults a POST body to `application/x-www-form-urlencoded`, which is now correctly refused.

`test_cli_mentions_read_timeout_next_to_max_body` and `test_instrument_openai_without_sdk_is_safe_noop` fail in this dev venv identically on `upstream/master` (missing `snagline` console script; `openai` installed locally where CI installs only `langchain-core`). Full suite otherwise: **885 passed, 1 skipped**, coverage **88.06%**. `ruff check src tests` and `ruff format --check src tests` clean; `mypy --no-incremental --python-version 3.12 src tests` shows only the 7 pre-existing `auto/openai.py` errors (CI's mypy legs install no SDKs, so neither artifact reaches CI).

Fixes #388


Cyrax321 (commented):
**Code Review — PR #407**

Well-reasoned CSRF-style hardening: all mutating POST routes pass through a single gate (Sec-Fetch-Site authoritative, Origin-vs-Host fallback, JSON content type), the gate ordering is correct (origin → auth → length → content-type → size → route), and every in-repo client (three sinks, `snagline hook`, the monitor halt webhook) sends `application/json`. Test coverage is strong. Three issues: one arm fails open despite the body claiming it was fixed, one documented curl recipe is broken by the 415 gate, and the Origin/Host echo is transparently bypassable under DNS rebinding.

Findings, most severe first:

**1. `medium` — the Host-less-Origin arm still returns `None` (treated as same-site), though the PR body claims it was changed to fail closed.**
`src/snagline/server/http_server.py:609-611`
The PR body's self-review says "The arm now returns the declared origin", but head code reads `host = self.headers.get("Host")` / `if not host: return None`, and `do_POST:705` treats `cross_site is None` as allowed (skips the 403). So `POST /episodes/end` with `Origin: https://evil.example` and no Host header (HTTP/1.0 sender or stripping proxy) lands the destructive write — the exact scenario the body claims was eliminated. Marginal exploitability (a non-browser can omit Origin anyway), but the gate violates its own stated premise and the description misdescribes the code; either fix the arm or correct the body.

**2. `medium` — the shipped curl recipe for `/hooks/claude-code` sends `application/x-www-form-urlencoded` and now gets 415.**
`docs/FRAMEWORK_BRIDGES.md:95-96`
`curl -s -X POST http://127.0.0.1:8787/hooks/claude-code -d '{...}'` sets curl's default content type, which the new `if not self._json_content_type(): ... 415` (`http_server.py:752`) rejects — so a user following the quickstart verbatim gets `{"error": "unsupported media type"}` and the loop-risk demo silently stops working. The PR body claims "the documented curl recipes" all send JSON; this one doesn't (the other three recipes do). Add `-H 'Content-Type: application/json'` to the doc, or accept the form default for this bridge.

**3. `low` — the Origin/Host fallback authenticates a client-supplied header against another client-supplied header, so DNS rebinding defeats it.**
`src/snagline/server/http_server.py:612-619`
Under rebinding, `Origin == f"http://{Host}"` by attacker construction and `Sec-Fetch-Site` reads `same-origin`, so both gates pass and a page at the rebound origin reaches the token-less loopback sidecar. `host` is never anchored to the server's *bound* address, and `_SAME_SITE_VALUES` admits `same-origin`/`same-site` with no anchor. Modern browser private-network-access restrictions narrow the window, so this is a design note rather than a confirmed exploit — but the origin gate cannot distinguish rebinding from same-origin as designed. Pinning to the configured bind address (or a shared token on mutating routes) would close it.

PR #408, opened 2026-09-21
fix(batching): bound the shutdown flush so close() cannot hang on the delivery lock (#393)

### LINKED ISSUE #393: bug: BatchingSink.close() can block forever on the delivery lock when the background flusher is stuck
## Summary

`BatchingSink.close()` (batching.py:112-120) joins the flusher for `flush_interval + 1.0` seconds, then unconditionally calls `self._flush()` on the caller's thread. `_flush` → `_deliver` takes `self._delivery_lock` with **no bound** (batching.py:93). If the background thread is stuck inside that same lock — which a hung wrapped sink makes reachable — the join times out and `close()` then parks on the lock forever. Shutdown hangs instead of degrading.

The comment on that line argues "`_flush` is a no-op on an empty queue, so this is safe either way" — but the hazard is not the queue, it is the lock: the queue can be non-empty because the background thread is still working through it, and the caller's `_flush` blocks on `_delivery_lock` regardless of queue state.

## How the background thread comes to be stuck

`_deliver` wraps each `emit` in `contextlib.suppress(Exception)` (batching.py:104), so an *exception* cannot strand it — only a *hang* can. Two reachable hangs:

- A wrapped network sink whose `urlopen` never returns. `emit` is dispatched synchronously from the flusher, and the sink's `timeout` does not bound DNS resolution, so an unresolvable host against a black-holed resolver blocks there indefinitely.
- `_deliver` sleeps while holding `_delivery_lock` when `max_per_second` pacing is on (batching.py:95-103). A large batch at a small gap holds the lock for a long time by design, which is fine on its own but makes the join-timeout arithmetic meaningless: `flush_interval + 1.0` has no relationship to how long a batch can take to deliver.

## Reproduction

```python
import threading, time

class HungSink:
    def emit(self, risk):
        time.sleep(3600)          # a sink that never returns (e.g. hung DNS)

outer = BatchingSink(HungSink(), flush_interval=0.05)
# ...enqueue a risk so the background thread is inside _deliver...
done = threading.Event()
t = threading.Thread(target=outer.close, daemon=True)
t.start()
t.join(timeout=5)
print("close() still blocked after 5s:", not done.is_set() and t.is_alive())
```

The join times out in `flush_interval + 1.0` s, then the caller's `_flush` blocks on the lock the background thread still holds.

## Suggested fix

Bound the fallback flush the same way the join is bounded, and make a blocked shutdown observable instead of infinite:

```python
if not self._thread.is_alive():
    self._flush()
elif not self._delivery_lock.acquire(timeout=self._flush_interval + 1.0):
    logger.warning("snagline: BatchingSink shutdown timed out; %d alert(s) undelivered",
                   len(self._queue))
```

Releasing the lock in the timeout branch is unnecessary — it was never acquired. The undelivered count in the warning matters: today a hung shutdown produces no diagnostic at all, so the operator cannot tell "clean exit" from "stuck forever".

## Severity

Moderate. It needs a hung downstream sink to trigger, and those are the exact failures BatchingSink exists to shield ingest from — so the failure arrives precisely when the operator most wants the process to shut down and restart cleanly.

Adhi1-2:
## What

`BatchingSink.close()` could hang forever. It joined the flusher for `flush_interval + 1.0` s and then called `self._flush()` unconditionally — and `_flush` → `_deliver` takes `_delivery_lock` **with no bound** (`batching.py:93`). When the background thread was parked inside `_deliver` on a hung endpoint, shutdown parked on the very lock that thread held.

## Reproduction

The issue's repro as written does **not** hang, and the reason is the bug's precondition: the flusher clears the queue *before* taking `_delivery_lock`, so by the time it holds the lock there is nothing for the fallback flush to do and `_flush` returns early. The deadlock needs an alert enqueued **while** the flusher is parked — which is exactly what a live alert stream does:

```python
outer = BatchingSink(HungSink(), flush_interval=0.05)
outer.emit(risk); time.sleep(0.3)   # flusher now inside _deliver, holding the lock
outer.emit(risk2)                   # queue was empty -> fallback flush has real work
t = Thread(target=outer.close); t.start(); t.join(5)
print("still blocked:", t.is_alive())   # True, pre-fix
```

Verified: pre-fix, `close()` is still blocked after 5 s. The wrapped sink hangs because a network sink's `timeout` bounds socket ops, not DNS — an unresolvable host against a black-holed resolver blocks indefinitely — and `_deliver` wraps each emit in `suppress(Exception)`, so only a hang can strand it, not an exception.

## Change

`close()` now branches on whether the flusher actually exited:

- **flusher dead** — unchanged behaviour, pick up anything enqueued after its final pass.
- **flusher still alive** — acquire `_delivery_lock` with the same budget as the join. A slow-but-progressing delivery still delivers; a genuinely stuck one terminates instead of hanging, and the undelivered count is logged at WARNING so a stuck shutdown is observable rather than silent.

## Tests

Three new cases in `tests/test_batching_sink.py`:

| Test | Asserts |
|---|---|
| `test_close_does_not_hang_when_the_flusher_is_stuck` | `close()` returns in < 5 s; the late alert is dropped, not delivered (delivery never finished) |
| `test_close_reports_undelivered_alerts_when_shutdown_times_out` | a stuck shutdown logs `"undelivered"` — today it is silent |
| `test_close_delivers_after_a_slow_delivery_recovers` | the acquire-with-timeout is not "always give up": a sink that finishes mid-wait still delivers both alerts |

The hang test fails on the pre-fix tree (it took the full 41 s — it genuinely hung) and the undelivered-warning test fails pre-fix too; the recovery test passes both ways, which is the right role for a positive control.

Full gate on this branch:

- `pytest`: **871 passed, 1 skipped, 4 failed** — the 4 are the pre-existing environmental baseline on this dev machine (a sibling Continuum checkout on disk; `openai`/`anthropic` installed in the venv so the without-SDK no-op tests cannot be exercised; a `snagline` binary not installed for the help-text subprocess test), none related to batching.
- `mypy --no-incremental src tests`: **11 errors in 2 files** (`auto/openai.py`, `auto/anthropic.py`) — the known local baseline, in optional-SDK modules invisible to CI.

Fixes #393.


Adhi1-2:
Amended with a self-review fix that CI did not catch.

`close()` acquired `_delivery_lock` and then called `self._flush()` inside it. `_flush` -> `_deliver` re-acquires that same non-reentrant `Lock`, so whenever the probe succeeded the re-acquire blocked forever with no timeout covering it. Reachable when a delivery finished around the deadline (freeing the lock, so the probe succeeds) and an alert arrived mid-shutdown — an end-to-end repro hung `close()` in 6 of 10 trials.

The acquire is now a readiness probe and is released *before* the flush. `_flush`'s snapshot-and-clear under the queue lock already keeps the two flush paths from double-delivering, so nothing else needed to change.

New test `test_close_does_not_hold_the_delivery_lock_across_its_own_flush` pins it: it fails on the previous commit (`close() must terminate... holding _delivery_lock across _flush() deadlocks shutdown`) and passes now. The two existing tests cover the stuck and the slow-but-recovering paths; neither reaches this interleaving, which is why it survived.

Cyrax321 (commented):
**Code Review — PR #408**

The core mechanism is right: `close()` branches on flusher liveness, probes `_delivery_lock` with a bounded acquire, releases before its own `_flush` (avoiding the non-reentrant self-deadlock), and logs undelivered counts; no lock-order inversion is introduced and no other `close()` paths (context manager, atexit) exist in the repo. The tests genuinely fail pre-fix, and the lock-release fix is regression-guarded. The main residual gap is that the bound covers lock acquisition but not the `emit()` the fallback flush runs on the caller's thread — so `close()` can still hang forever on a freshly-stuck sink.

Findings, most severe first:

**1. `high` — the timeout covers `_delivery_lock` acquisition only, not the `sink.emit()` that `_flush` → `_deliver` runs on the caller's thread.**
`src/snagline/sinks/batching.py:95-109,121-123,140-142`
Both branches run `_flush()` *after* confirming the lock is free: the dead-flusher branch (`if not self._thread.is_alive(): ... self._flush(); return`) and the post-probe flush (`acquire` succeeds → `release()` → `self._flush()`). `_deliver` then wraps the sink in `contextlib.suppress(Exception)` — which bounds exceptions, not hangs. So a sink that stalls exactly at that final emit (the PR's own #393 scenario, where the emit timeout does not bound DNS) blocks `close()` forever, unobserved and unbounded. The join/acquire budgets never cover this call; the fallback flush needs its own bound (or a note that the residual hang is out of scope).

**2. `medium` — `logger.exception` in the webhook/slack sinks emits the traceback, which still carries the URL; redaction covers only the `%s` argument.**
`src/snagline/sinks/webhook.py:88-91`, `src/snagline/sinks/slack.py:84`
`redacted_destination(self._url)` redacts the format arg, but `http.client.InvalidURL("URL can't contain control characters. 'https://hooks.slack.com/services/T.../B.../<secret>'")` and `ValueError("unknown url type: ...")` embed the raw URL in the exception message, which lands in the logged traceback. Verified mechanism (urllib raises both for malformed URLs); narrow trigger (malformed/credential-bearing URL), but it defeats the #390 fix in exactly the operator-reads-the-log moment. Same treatment as the message: class-name-only, or `getattr(exc, "reason", None)` redacted too.

**3. `low` — `flush_now()` is an unbounded shutdown path one method away from the fix.**
`src/snagline/sinks/batching.py:111-113`
Its docstring says "used at shutdown", and it calls `_flush()` → `_deliver()` → `sink.emit()` while blocking on `_delivery_lock` with no timeout — the exact #393 hang, still reachable if a caller follows the docstring at shutdown instead of `close()`. Either bound it or fix the docstring to point at `close()`.

**4. `low` — on the acquire-timeout path the queued alerts are left in `_queue` and the flusher is still parked, so the "N alert(s) undelivered" warning can be false.**
`src/snagline/sinks/batching.py:142-148`
If the endpoint recovers after `close()` returned, the still-running daemon's `finally` re-runs `_flush()` and delivers them — operators who paged off the warning chased alerts that were actually delivered, and code treating `close()`'s return as "queue drained" is wrong. Worth logging them as "delayed, may still deliver" or dropping them definitively.

**5. `low` — the module docstring still promises `close()` drains the queue before returning.**
`src/snagline/sinks/batching.py:16-17`
Now false on the new timeout path (returns with alerts dropped, warning only). The CHANGELOG documents the new behaviour; the module contract should match.

**6. `low` — `self._flush_interval + 1.0` is spelled out three times in `close()` (join, acquire, warning text).**
`src/snagline/sinks/batching.py:119,140,147`
The "acquire uses the same budget as the join" invariant is enforced only by convention; a local `budget = ...` would make it structural. Cosmetic.

**7. `low` — three unrelated fixes ride in one PR under a batching-only title.**
`CHANGELOG.md:14`
This PR carries #390 URL redaction (`sinks/base.py`, `slack.py`, `webhook.py`), #394 max-body-bytes validation (`cli.py`, `server/http_server.py`), and #393's close() bound (`sinks/batching.py`) in 13 files / +477, while the conventional-commit title names only `fix(batching)`. A regression in any one must be reverted by reverting all three (or a hand-crafted partial revert), and the title gives no hint that server/CLI code changed. Noting for the changelog/revert history rather than as a code defect.

PR #409, opened 2026-09-21
fix(state): fall back on an unusable redis URL instead of crashing startup (#392)

### LINKED ISSUE #392: bug: default_state_backend() crashes startup on a Redis URL with a non-redis scheme; only ImportError is caught but redis-py raises ValueError
## Summary

`default_state_backend()` resolves `SNAGLINE_STATE_BACKEND=redis` with `RedisStateBackend(url)`, but the surrounding `try` catches only `ImportError` (state.py:217-223). redis-py's `ConnectionPool.from_url` raises `ValueError` for any URL whose scheme is not `redis://`, `rediss://` or `unix://`, and that `ValueError` escapes — out of `default_state_backend()`, through `Monitor.default()`, and into the process. The module's contract, stated by the two neighbouring arms (the missing-URL warning at state.py:206-215 and the missing-dependency warning at state.py:219-223), is to warn and fall back to in-memory state.

## Reproduction

Fresh clone of master (`b667169`), redis-py 8.1.0 installed:

```
$ SNAGLINE_STATE_BACKEND=redis SNAGLINE_STATE_REDIS_URL=postgres://host:5432/db python -c \
    "from snagline.state import default_state_backend; print(type(default_state_backend()).__name__)"
  ...
ValueError: Redis URL must specify one of the following schemes (redis://, rediss://, unix://)
```

The same `ValueError` fires from `redis.Redis.from_url(...)` directly, confirming it is redis-py's URL parser rather than a connection attempt — so no reachable Redis is needed to trigger it, and it fails at construction, not later on first lock use.

Realistic triggers, none of them exotic:

- A typo or a copy-paste from a sibling service's config: `SNAGLINE_STATE_REDIS_URL=postgres://...`.
- An operator who writes the bare host (`SNAGLINE_STATE_REDIS_URL=host:6379`) instead of a URL, which `urlparse` also classifies as a bad scheme.
- A secrets-manager template that substitutes a `rediss://` URL with a placeholder that is not a URL at all.

In every case the operator gets a traceback at startup instead of the documented warn-and-fallback, on the configuration knob whose entire purpose is to be optional.

## Suggested fix

Broaden the caught exception to match what the constructor can actually raise, and warn consistently with the two existing arms:

```python
else:
    try:
        return RedisStateBackend(url)
    except (ImportError, ValueError) as exc:
        logger.warning(
            "snagline: redis backend requested but unusable (%s); "
            "falling back to in-memory state", exc,
        )
```

`ImportError` and `ValueError` cover the two failure shapes (`import redis` inside `RedisStateBackend.__init__`, and redis-py's URL parser). Catching `Exception` would also work but is broader than needed and would hide a genuine programming error in this optional path.

Worth noting while here: `SNAGLINE_STATE_BACKEND` is `.lower()`-stripped at state.py:203 but the value `" redis"` (leading space) survives `.lower()` untouched and silently selects the memory backend — the same class of silent-wrong-backend as #319's missing-URL case. A `.strip()` there would close it for free.

## Severity

Moderate. It takes one malformed URL to take a process down at startup, and the failure contradicts the module's own documented fallback contract.

Adhi1-2:
## What

`default_state_backend()` resolves `SNAGLINE_STATE_BACKEND=redis` with `RedisStateBackend(url)`, but the surrounding `try` caught only `ImportError`. redis-py's `from_url` raises `ValueError` for a URL it cannot use, and that `ValueError` escaped — out of `default_state_backend()`, through `Monitor.default()`, and into the process.

The module's own contract is to warn and fall back. Both neighbouring arms do: the missing-URL warning (`state.py:206-215`) and the missing-dependency warning (`state.py:219-223`). An unparseable URL was the one arm that escaped.

Reproduced against `b667169` with a stub standing in for redis-py (no server or extra needed):

```
SNAGLINE_STATE_BACKEND=redis SNAGLINE_STATE_REDIS_URL=postgres://host:5432/db
-> RAISED ValueError: Redis URL must specify one of the following schemes
   (redis://, rediss://, unix://)
```

The failure is at construction, not on first lock use — no reachable Redis is needed to trigger it.

## Fix

Catch `ValueError` alongside `ImportError` and warn consistently, naming the offending URL. The fallback is silent at the protocol level (in-memory state looks healthy), so the log line is the operator's only signal that their cross-process coordination knob did not take — which is the same reason the two existing arms log.

Nothing else about the knob changes: a URL the parser accepts still builds `RedisStateBackend`.

### The same crash had a second shape, found in self-review

redis-py's `parse_url` also reads `urlparse(url).port`, unguarded, and **urllib defers port validation to that attribute access** rather than doing it at parse time. So a typo'd port raises the same `ValueError` from the same constructor:

```
SNAGLINE_STATE_REDIS_URL=redis://localhost:99999/0  ->  ValueError: Port out of range 0-65535
SNAGLINE_STATE_REDIS_URL=redis://localhost:abc/0    ->  ValueError: Port could not be cast to integer value as 'abc'
```

The new arm caught that `ValueError` correctly — and then called the redaction helper for the log line, which read `.port` **a second time** and re-raised out of the handler. So the port was the one URL shape that still crashed startup, through the very fallback meant to contain it, and a mistyped port is at least as likely as a mistyped scheme.

`_redact_url` now tolerates an unusable port: it is dropped from the redacted form (it is the part that is broken) and the warning's exception text names the actual problem. The parse failure for a scheme-less or host-less URL is unchanged.

## Scope

`src/snagline/state.py` — the `except ValueError` arm, plus the port guard in `_redact_url`. `tests/test_state_backend.py` — 8 tests (6 parametrized + 2 guards): the `fake_redis` stub now hits `.port` before scheme validation the way redis-py does. CHANGELOG. No behaviour change beyond the one arm and its redaction.

I checked the region against every open PR before starting: `state.py:195-235` is touched by no other open PR, mine or another contributor's, so this lands without a rebase collision.

## Verification

Reproduced both `ValueError` shapes before the fix (output above); both gone after.

- `pytest`: **889 passed, 2 failed, 1 skipped**, coverage **88.02%** (80% gate). The 2 failures are this machine's known environmental baseline — a missing `snagline` console script, and an `auto/` no-SDK test that needs the OpenAI SDK absent — unrelated to this change.
- Both new port tests fail on the pre-redaction-fix tree and pass after, with the existing 6 unaffected.
- `ruff check src tests`, `ruff format --check src tests` — clean (139 files).
- `mypy --no-incremental --python-version 3.12 src tests` — the 7 pre-existing errors in `auto/openai.py` only. The repo pins `python_version = "3.10"`, which numpy 2.5.3's bundled stubs reject at their own `numpy/__init__.pyi:737` before any project file is checked — an environment artifact, not a project error; CI's mypy legs install no SDKs and are unaffected.
- Mutation check: removed the new arm and re-ran — the four original scheme cases plus the URL-is-named assertion fail. `test_good_redis_url_still_builds_the_backend` is a regression guard that passes both before and after: a widening that fell back on every URL would be worse than the crash it fixes, so it must not pass silently.

Fixes #392.


Adhi1-2:
Acted on the Major finding on the redis URL warning: the URL is the credential — redis-py reads the password out of it, and `SNAGLINE_STATE_REDIS_URL` is normally *set* with a secret in it — so this PR's own warning was the one line guaranteed to log the secret, at exactly the moment an operator goes looking in the logs.

The value is now redacted to scheme and host before it reaches the line, keeping the parts that identify the misconfigured endpoint. redis-py's parse failures can also quote the offending value back in the exception text, so once the URL carries a secret the exception's message is withheld and only its type is logged; a credential-free URL keeps the full message, which is the diagnosis.

Tests added (`tests/test_state_backend.py`): a parametrized check that `redis://:pw@host`, `redis://user:pw@host`, and a query-string password all stay out of the line while `host` stays in; separate coverage for the exception-text leak path; a positive case proving a credential-free URL keeps its full message; and direct `_redact_url` cases including a unix socket and an unparseable value.

Cyrax321 (commented):
**Code Review — PR #409**

The added `except ValueError` arm in `default_state_backend()` does fall back to in-memory state with a redacted, class-first warning for the advertised URL shapes, and `_redact_url` correctly refuses to echo back unparseable input. But the handler re-parses the URL in two places of its own, and both can raise the very `ValueError` being contained — so several malformed-URL classes still crash startup, exactly the shapes the PR body claims are fixed.

Findings, most severe first:

**1. `high` — `_redact_url`'s unguarded `parsed.port` re-raises for port-typo URLs, so the fallback crashes startup on the two shapes the PR body claims are fixed.**
`src/snagline/state.py:53`, `src/snagline/state.py:286`
redis-py's `parse_url` reads `urlparse(url).port` unguarded and raises for a bad port; the new handler catches it, computes `detail`, then evaluates `_redact_url(url)` as a `logger.warning` argument — and line 53's `if parsed.port is not None:` has no try around it. Verified empirically against the PR head with a redis-py-faithful stub: `redis://localhost:99999/0` and `redis://localhost:abc/0` both crash `default_state_backend()` with `ValueError: Port out of range 0-65535` / port-cast error, while `postgres://host:5432/db` and `host:6379` fall back correctly. The PR body also asserts `_redact_url` "now tolerates an unusable port" and cites two new port tests — `rg '99999|Port out of range' tests/test_state_backend.py` at head returns nothing, and the existing `fake_redis` stub never touches `.port`. Wrap the `parsed.port` access (mirroring the guard `_redact_url` already needs anyway).

**2. `medium` — the handler itself calls `urlparse(url)` unguarded at state.py:271, so URLs that `urlparse` rejects defeat the fallback for that URL class.**
`src/snagline/state.py:271`
A truncated copy-pasted IPv6 endpoint (`redis://[::1`, or `redis://[::1:6379/0`) makes redis-py raise `ValueError("Invalid IPv6 URL")` at construction; the handler catches it, then line 271's bare `urlparse(url)` raises the same error again, uncaught, out through `Monitor.default()` — the original #392 crash in a new costume. Verified: "CRASH 'redis://[::1' -> ValueError: Invalid IPv6 URL" while every non-IPv6 case fell back. Reuse the guarded parse from `_redact_url` (which correctly returns `"<invalid redis url>"` for this same input).

**3. `medium` — the fallback covers only construction-time failures; a URL that parses but points at an unreachable Redis builds the backend silently and first fails mid-ingest.**
`src/snagline/state.py:217`
redis-py connects lazily, so `redis://redis.internal:6379/0` with a down or mis-DNSed host constructs cleanly, logs nothing (both neighbouring arms are silent), and then each `episode_lock()` acquisition can raise `redis.exceptions.ConnectionError` unhandled inside the ingest path — the operator's only signal is a per-step traceback after startup reported healthy. This is incompleteness of the stated "knob is optional, warn and fall back" contract rather than a regression (and `ConnectionError` is not `ValueError`, so widening is a policy decision) — but the unreachable-Redis case deserves at least a first-use warning.

**4. `low` — the URL is parsed twice per warning with divergent guarding.**
`src/snagline/state.py:271` vs `src/snagline/state.py:41-44`
The handler parses the URL once unguarded, `_redact_url` parses it again inside a `try`. The guarded copy already implements exactly the tolerance the handler needs; threading one parsed result through both would remove the duplication and eliminate finding #2 by construction.

Issue #19, opened 2026-08-15
[ConsoleSink] emit() raises on broken stream, violating AlertSink fire-and-forget protocol


Cyrax321:
### System Information
- OS Platform and Distribution: macOS (Apple Silicon)
- Language / runtime version: Python 3.13.5
- Package / library version(s): snagline-agent 0.1.0
- Installation method: source

### Describe the current behavior
`ConsoleSink.emit()` in `src/snagline/sinks/console.py` calls `self._stream.write()` and `self._stream.flush()` without a try/except. If the stream is broken (e.g., stderr is redirected to a closed pipe, or the file descriptor is invalid), the method raises an `IOError` / `OSError`.

The `AlertSink` protocol at `src/snagline/sinks/base.py:17` states:
> `emit`: Emit one risk. Must be fire-and-forget and never block ingest().

`ConsoleSink` violates this contract. While the Monitor's fail-open wrapper catches the exception at the orchestrator level (`src/snagline/monitor.py:72`), the sink itself is not fire-and-forget.

Compare with `WebhookSink` at `src/snagline/sinks/webhook.py:45-50`, which wraps its HTTP call in try/except and never raises.

### Describe the expected behavior
`ConsoleSink.emit()` should catch exceptions from `stream.write()` and `stream.flush()`, log them, and return without raising, matching the behavior of `WebhookSink`.

### Standalone code to reproduce the issue
```python
from snagline.sinks.console import ConsoleSink
from snagline.risk import FailureRisk

class BrokenStream:
    def write(self, s):
        raise IOError('stream broken')
    def flush(self):
        pass

sink = ConsoleSink(stream=BrokenStream())
risk = FailureRisk('ep-1', 'step-1', 0.8, 'loop', 'test', 1.0)

try:
    sink.emit(risk)
    print('No exception')
except Exception as e:
    print(f'RAISED: {type(e).__name__}: {e}')
    print('ConsoleSink.emit() is not fire-and-forget')
```

### Other info / logs
Output:
```
RAISED: IOError: stream broken
ConsoleSink.emit() is not fire-and-forget
```

The fix is to wrap the write/flush calls in a try/except that logs and returns, matching `WebhookSink`'s pattern.

Cyrax321:
Resolved in code: ConsoleSink.emit wraps stream write/flush in try/except (logs once, returns) so a broken stream is fire-and-forget. Added tests/test_console_sink.py. PR #23

PR #410, opened 2026-09-21
fix: reject binary streams in ConsoleSink at construction (#391)

### LINKED ISSUE #391: bug: ConsoleSink raises TypeError out of emit() on a binary stream, which except OSError does not catch, so every alert is silently dropped
## Summary

`ConsoleSink`'s raw-stream path catches only `OSError` (console.py:69), but writing a `str` to a binary stream raises `TypeError`, which is not an `OSError` subclass. `stream=` is documented as accepting "stdout, a file, etc." (console.py:4-6, 33-35), so `stream=open("alerts.jsonl", "wb")` or `stream=sys.stdout.buffer` are natural misconfigurations with no validation in `__init__`. Every `emit()` then raises into the caller — under a `Monitor` with `fail_open=True` (the default) the exception is swallowed by the dispatch layer, so every alert is silently dropped and the sink's own fire-and-forget contract is violated; with `fail_open=False` it reaches the host's ingest path.

Sibling issue of #327 (closed stream raising `ValueError` past the same `except OSError`) — same mistake, different exception type, and this one has no external precondition: the stream is open and healthy, it is just the wrong type.

## Reproduction

Fresh clone of master (`b667169`):

```python
import io
from snagline.sinks.console import ConsoleSink

ConsoleSink(stream=io.BytesIO()).emit(FailureRisk("ep", "s1", 0.9, "loop", "detail", 0.0))
# -> TypeError: a bytes-like object is required, not 'str'
#    (issubclass(TypeError, OSError) is False)
```

Same outcome for `open(path, "wb")`, which is arguably the more likely spelling for anyone routing alerts to a file.

## Suggested fix

Two options, in increasing thoroughness:

1. Widen the guard to cover the type mismatch as well, since both are "the stream is unusable, drop the alert":

```python
except (OSError, TypeError, ValueError):
    logger.warning("snagline ConsoleSink: write to stream failed; ...")
```

2. Reject the bad stream at construction, which is better for the operator because it fails loudly at startup instead of dropping alerts one at a time for the whole run:

```python
if stream is not None and not isinstance(stream, io.TextIOBase):
    raise TypeError("ConsoleSink stream must be a text stream")
```

I would take #2 with #1 kept as the safety net — a stream can be closed or replaced after construction, so the guard should stay regardless.

Note `logging_sink.py` writes through a `logging.Handler`, which encodes internally and so is unaffected; `heartbeat.py`'s path handling is a separate (#323) shape.

## Severity

Low to moderate. Needs a misconfiguration to trigger, but the failure is the worst kind once it does: the monitor runs cleanly, the banner prints, and every alert is discarded with the error hidden behind the fail-open contract.

Adhi1-2:
## What

`ConsoleSink(stream=open(p, "wb"))` — or `stream=sys.stdout.buffer` — constructs fine and then drops **every** alert for the whole run. A `str` write to a binary stream raises `TypeError`, which is not an `OSError` subclass, so the fire-and-forget guard in `emit()` doesn't catch it. The exception escapes `emit()` into the host's ingest path, or if the host swallows it, the alert is gone; either way the monitor is otherwise completely healthy and reports nothing.

These are the natural ways to reach for routing alerts to a file, so this is an easy misconfiguration to hit, and once hit it is invisible.

Reproduced against `b667169`:

```
>>> ConsoleSink(stream=open("/tmp/a.jsonl", "wb")).emit(risk)
TypeError: a bytes-like object is required, not 'str'
```

## Fix

Probe the supplied stream with an empty write at construction and raise a `TypeError` naming the stream type, the underlying error, and the two correct spellings (`open(path, 'w')`, or the logging module via `logger=`). Failing at construction rather than alert-time turns a silent whole-run outage into a startup error, which is what the operator wants from a misconfiguration of the default sink.

The empty write writes nothing, so the only thing under test is whether the stream accepts a `str` at all. A stream that is already closed trips the same check (`ValueError`), so it is caught here too rather than on the first alert.

Two deliberate scope decisions:

- **`OSError` is not a rejection.** A stream whose write fails with `OSError` is a runtime breakage (a pipe that closes, a bad descriptor mid-run), not a misconfiguration. Issue #19's contract is that such a sink still constructs and stays fire-and-forget in `emit()`; rejecting it here would change that contract for no benefit, since `emit()` already drops the alert and logs rather than raising.
- **`emit()` now catches `ValueError`/`TypeError` as well.** Construction-time validation cannot cover a stream that is closed or swapped out afterwards, so the guard stays as the second layer.

### The probe missed the mistake its own message was written for (self-review)

The probe caught only `(TypeError, ValueError)`. A caller passing the file *path* rather than an opened file — precisely the mistake the new message advises on with `pass open(path, 'w')` — hit neither, because a `str` has no `write` attribute at all:

```
>>> ConsoleSink(stream="alerts.log")
AttributeError: 'str' object has no attribute 'write'
```

So the one operator who needed the guidance got a bare `AttributeError` instead. `AttributeError` is now in the caught tuple and the constructed message reaches them.

## Scope

`src/snagline/sinks/console.py` — the constructor probe, widening its caught tuple to `AttributeError`, plus widening the `emit()` `except` tuple. `tests/test_console_sink.py` — 8 tests. CHANGELOG. No other behaviour change.

**Relationship to #382 (correction of my earlier claim):** I checked the region before starting and wrote above that this lands without a collision. That was wrong, and I'm correcting it rather than leaving it standing. #382 inserts its `_fault_logged` latch inside the same `__init__` block my probe occupies, and both PRs rewrite the same `except OSError:` line in `emit()`. Verified with `git merge-tree` against #382's head `3d16a91`: conflicts in both `console.py` and `test_console_sink.py`. The two changes are complementary — #382 is the log-once latch for a dead stream at alert time, this is construction-time rejection of a misconfigured one — but they touch the same lines, so they need ordering. I'm happy to rebase this onto #382 if that one lands first, or drop the `emit()` widening from this PR if the maintainer would rather keep them disjoint.

## Verification

Reproduced the `TypeError` before the fix (output above); gone after.

- `pytest`: **879 passed, 2 failed, 1 skipped**, coverage **88.00%** (80% gate). The 2 failures are this machine's known environmental baseline — a missing `snagline` console script, and an `auto/` no-SDK test that needs the OpenAI SDK absent — unrelated to this change. The new path-string test fails on the pre-widening tree (`AttributeError` at the probe) and passes after.
- `ruff check src tests`, `ruff format --check src tests` — clean (139 files).
- `mypy --no-incremental --python-version 3.12 src tests` — the 7 pre-existing errors in `auto/openai.py` only. The repo pins `python_version = "3.10"`, which numpy 2.5.3's bundled stubs reject at their own `numpy/__init__.pyi:737` before any project file is checked — an environment artifact, not a project error; CI's own legs are unaffected.
- Mutation check: removed both hunks and re-ran — **5 of the 7 new tests fail**, the two construction-rejection tests, the file-spelling variant, the closed-stream variant, and the message-content assertions. The remaining two pass both before and after, and they are guards that *must* pass both ways: `test_console_sink_default_stream_is_stderr` (probing the default `sys.stderr` on every construction would write to the terminal, so the probe must only run on an explicitly supplied stream) and `test_console_sink_probe_writes_nothing_to_a_good_stream` (the probe must leave a good stream clean for the real payload). A fix that rejected every stream would fail the first; one that emitted junk would fail the second.

Fixes #391.


Cyrax321 (commented):
**Code Review — PR #410**

The headline fix is correct and well-tested: the empty-write probe rejects `BytesIO`, `open(p, "wb")` and `sys.stdout.buffer` at construction with an actionable message, and the `OSError` carve-out is argued well. However, the branch is based on a pre-#329 master (`b667169`): as written it silently reverts two merged behaviors, breaking existing master tests, and the PR body claims a change the diff does not contain.

Findings, most severe first:

**1. `high` — merging this as-is reverts merged #327/#329 behavior: the log-once latch is deleted and closed streams are now rejected at construction, breaking two green master tests.**
`src/snagline/sinks/console.py:107`
The PR head's `emit()` rewrites the except clause without the `_fault_logged` latch (master's `console.py:46,72-75` has it). After merge, `tests/test_issues_321_327.py::test_console_sink_closed_stream_does_not_raise` raises `TypeError` at `ConsoleSink(stream=closed_StringIO)` where master asserts fire-and-forget, and `test_console_sink_warns_only_once_for_a_dead_stream` fails — one `logger.warning` per alert instead of once, the exact warning-spam regression #329 fixed. The PR was validated against `b667169`, which predates the 2026-09-21 merge of #329. Rebase onto current master before this merges; the two contract changes conflict.

**2. `medium` — the body claims `AttributeError` was added to the caught tuple and a path-string test was added; neither is in the diff.**
`src/snagline/sinks/console.py:71`
The head code catches only `(TypeError, ValueError)`, and the 8 added tests include no `stream="alerts.log"` case. So `ConsoleSink(stream="alerts.log")` — the exact mistake the new message coaches on with "pass `open(path, 'w')`" — still escapes the constructor with a bare `AttributeError: 'str' object has no attribute 'write'` and never reaches the actionable message. Adding the claim to the diff (one tuple element + one test) or dropping it from the body is a one-line fix.

**3. `medium` — a read-mode file handle slips through the probe and constructs, then drops every alert for the run.**
`src/snagline/sinks/console.py:67`
`open(p, "r")`'s `write("")` raises `io.UnsupportedOperation`, an `OSError` subclass, so the `except OSError: pass` arm accepts it and the sink constructs just to silently drop every alert behind the ambient logger. The comment frames `OSError` as runtime breakage (closed pipe, bad descriptor mid-run), but a wrong-mode handle is a construction-time misconfiguration symmetric with the closed stream the probe does reject — the split is incidental to Python's exception taxonomy (`UnsupportedOperation` inherits both `OSError` and `ValueError`), not to misconfiguration semantics. Checking `stream.writable()` (or catching `UnsupportedOperation` explicitly) would close it.

**4. `low` — a closed-but-correctly-typed stream is reported as a type misconfiguration.**
`src/snagline/sinks/console.py:71`
A `StringIO` closed moments earlier (e.g. a context manager that exited before sink construction) raises `ValueError("I/O operation on closed file")`, which this converts into a `TypeError` whose text says "must be a writable text stream" and recommends `open(path, 'w')` — sending the operator to fix the type when the real bug is construction ordering after stream teardown.

**5. `low` — the probe assumes stdlib `write()` semantics.**
`src/snagline/sinks/console.py:66`
A third-party wrapper whose `write` requires extra args is false-rejected at construction, and a wrapper that emits a record boundary per `write()` call gets an empty artifact written before `self._stream` is assigned. No in-repo call site (`monitor.py:1221`, `cli.py:136`, examples) triggers it, so impact is limited to custom stream wrappers.

Adhi1-2:
Rebased onto the latest `master` and resolved one real conflict with the merged #327 work.

#327 (now upstream) established that a *closed* stream must still construct — closing is runtime breakage, and its contract is that `ConsoleSink` builds and drops the alert in `emit()`, warning once. This PR originally rejected a closed stream at construction, which would have broken that. The probe is now scoped to the type mismatch alone: only a binary stream (`open(p, "wb")`, `sys.stdout.buffer`), whose `str` write raises `TypeError`, is a construction-time misconfiguration.

Concretely: `OSError` and the `ValueError` a closed stream raises both fall through to the fire-and-forget `emit()` path (issues #19 and #327); only `TypeError` raises at construction. `test_console_sink_rejects_closed_stream_at_construction` was flipped to `..._accepts_...` as the control that pins that scope, and the emit guard keeps its deduped warn-once latch from #327 while still catching `TypeError`.

PR #411, opened 2026-09-21
fix(ml): validate EsnCusumDetector knobs and dedupe fail-open logging

### LINKED ISSUE #386: bug: EsnCusumDetector does not validate cusum_h / reservoir_size; cusum_h=0 deadens the detector and floods ERROR logs per step
## Summary

`EsnCusumDetector` (the optional `ml` extra) does not validate its own constructor knobs, and its fail-open wrapper logs at ERROR level on every step. Together: `cusum_h=0` raises `ZeroDivisionError` inside `_observe` on *every* step, the wrapper swallows it and logs a full traceback on *every* step, and the detector emits zero risks for the rest of the run. A negative `cusum_h` inverts into a false-positive storm instead. `reservoir_size=0` fails outright at construction.

This is the same defect shape #331/#333/#372 covered for the deterministic detectors, in the one detector those issues do not reach. It is also the one place the codebase's own convention was not followed: `SideEffectGuardDetector`, `StagnationDetector`, `TokenRunawayDetector` and `CompactionTripwireDetector` all re-validate their knobs in `__init__` with the explicit comment that "direct construction with explicit kwargs skips the Config check, hence the duplicate guard here". `EsnCusumDetector` clamps `warmup_steps` (`max(1, warmup_steps)`) and then validates nothing else.

## Reproduction

Against a fresh clone of master (`b667169`) with `numpy` installed:

```python
from snagline.ml.esn_ensemble import EsnCusumDetector

d = EsnCusumDetector(cusum_h=0.0)     # _h is a divisor in the score formula
for e in events:                     # 50 ordinary tool_call steps
    d.observe(e)
# -> ZeroDivisionError on all 50 steps, 50 ERROR-level log records, 0 risks
```

The offending line is `esn_ensemble.py:224`:

```python
score = min(1.0, 0.5 + 0.5 * (st.cusum - self._h) / self._h)
```

The alarm test `st.cusum >= self._h` is satisfied with `h = 0` (cusum starts at 0), so the division by `self._h` runs on the very first step. Measured:

| constructor value | effect over 50 ordinary steps |
| --- | --- |
| `cusum_h=0.0` | 50 `ZeroDivisionError`, 50 ERROR log records, 0 risks — detector permanently dark |
| `cusum_h=-1.0` | 40 risks over 40 steps of identical healthy traffic — `cusum >= h` holds every step and the `st.cusum = 0.0` re-arm makes it fire on each one |
| `reservoir_size=0` | `ValueError: zero-size array to reduction operation maximum` at construction |

The log-flood half is separate and reaches the default configuration too. `observe()` is documented as

> Never raises: an internal error is logged once per occurrence by the Monitor and here as well

but the Monitor-level dedupe it refers to (`MLOrchestrator._log_fault_once`, which this detector is documented to run inside) is unreachable: `observe` swallows the exception and returns `None`, so the orchestrator sees a quiet detector rather than a raising one and never records the fault. Any permanent internal failure therefore emits one full ERROR traceback per step for the life of the process.

## Suggested fix

Validate in `__init__`, following the existing precedent:

```python
if cusum_h <= 0.0:
    raise ValueError(f"cusum_h must be positive; got {cusum_h!r}")
if cusum_k < 0.0:
    raise ValueError(f"cusum_k must be non-negative; got {cusum_k!r}")
if reservoir_size < 1:
    raise ValueError(f"reservoir_size must be >= 1; got {reservoir_size!r}")
```

and make the fail-open wrapper match the orchestrator's dedupe rather than logging unconditionally — a module-level `logging.getLogger("snagline").warning(...)` gated on a small `set` of seen faults, so a persistent failure is reported once instead of per step. The `cusum_h` divisor also deserves a defensive non-zero denominator in the score formula, so a future caller path that bypasses construction cannot reproduce the dead-detector outcome.

## Severity

Low to moderate. The knobs are not reachable through `Config` — `Monitor.default` builds this detector with `EsnCusumDetector(baseline=cfg.goal_drift_baseline)` and every other argument defaulted — so the trigger is direct library construction, not operator misconfiguration. Filing it because the silent-deadening and log-flood outcomes are both bad enough to be worth the three-line guard, and because the class is the one outlier against the codebase's own validation convention.

Adhi1-2:
## What

`EsnCusumDetector` (the optional `ml` extra) did not range-check its constructor knobs. Unlike every other detector, these knobs have **no `Config` field** — `Monitor.default` builds the detector with defaults only — so `__init__` was the *only* place a bad value could be caught, and it did not check.

## Reproduced (50 identical healthy steps)

| knob | before |
|---|---|
| `cusum_h=0.0` | `ZeroDivisionError` on **every** step: 50 swallowed exceptions, 50 ERROR log records, **0 risks** — the detector is permanently dark, and the log makes it look merely noisy |
| `cusum_h=-1.0` | a non-negative accumulator can never cross a negative threshold → detection silently disabled (**50 false positives**, since the re-arm path still trips on healthy drift) |
| `cusum_k=-1.0` | negative slack inflates the accumulator each step → **16 false alarms** on clean traffic |
| `reservoir_size=0` | opaque numpy error at construction (`zero-size array to reduction operation maximum`) |

## Fix

- `reservoir_size >= 1`, `cusum_k >= 0.0`, `cusum_h > 0.0` now raise `ValueError` at construction, naming the knob and its effect. Follows the existing duplicate-guard convention (`stagnation`, `side_effect_guard`, `token_runaway`); the comment notes the difference — there is no upstream `Config` check to duplicate here, so the constructor is the sole gate.
- **The log flood, separately:** `observe()` swallows the exception, so `MLOrchestrator._log_fault_once` layered above it is *unreachable* — the per-step traceback came from this detectors own `logger.exception`. Faults are now reported once per `(phase, exception class)`, so a *different* fault surfacing later still gets reported rather than being silenced by the first.
- The score denominator is floored independently of the constructor check, so a future path that mutates `_h` after construction degrades to a saturated score instead of dividing by zero every step. The floor applies to the **denominator alone** — flooring the numerator too made `cusum == h == 0` yield score 0.0, which is out of range and gets silently dropped by the orchestrator.

Severity is low-to-moderate as the issue notes: the knobs are not reachable through `Config`, so this only bites direct construction. The reason to fix it is that the class is the one outlier against the codebases own validation convention.

## Tests

15 new cases in `tests/test_ml_esn_ensemble.py`. Verified against pre-fix source (overlay recipe): **12 fail pre-fix, all 15 pass after**. The 3 that pass on both sides are deliberate — two positive controls (the boundary value `cusum_k=0.0` still detects; the shipped defaults are unaffected) and one design guard (the dedupe key must carry the exception class, else the first fault silences every later one).

Full suite: **886 passed**, 1 skipped, 2 pre-existing environment-only failures identical on `upstream/master` (`test_cli_mentions_read_timeout_next_to_max_body` — no console script in the editable install; `test_instrument_openai_without_SDK_is_safe_noop` — openai is installed locally but CI installs only `langchain-core`). `ruff check` and `ruff format --check` clean. `mypy src tests` reports nothing in the touched files.

Fixes #386


Cyrax321 (commented):
**Code Review — PR #411**

The three issue-#386 knobs are range-checked at the constructor (the only gate — `Config` has no ESN fields, verified in `config.py`), with messages matching the codebase's knob-naming convention, and the per-step fail-open log flood is deduped in the same spirit as `MLOrchestrator`. Well-tested. The main residual gap: the new checks are not NaN/inf-safe, so exotic float values still reproduce the exact "permanently dark, silently disabled" mode the PR says it eliminates.

Findings, most severe first:

**1. `medium` — the checks pass NaN and +inf, and a NaN knob silently disables detection rather than being rejected.**
`src/snagline/ml/esn_ensemble.py:162-171`
The guards are `reservoir_size < 1`, `cusum_k < 0.0`, `cusum_h <= 0.0` — all `False` for `nan` (and `inf < 0.0`). With `cusum_h=nan`, `st.cusum >= self._h` (line 276) is always `False`; with `cusum_k=nan`, `max(0.0, st.cusum + signal - self._k)` (line 275) clamps to `0.0` every step. Either way the detector never fires and never raises — a worse outcome than the `-1.0` case the PR rejects, since there is no log at all. `math.isfinite` on the two float knobs would close it, and would match the sibling treatment other float knobs are getting in #412.

**2. `low` — a fractional `reservoir_size` still dies with the opaque numpy error the PR said it was replacing, and of the wrong type.**
`src/snagline/ml/esn_ensemble.py:162,180`
`EsnCusumDetector(reservoir_size=32.5)` passes `< 1`, then `rng.uniform(-1.0, 1.0, (32.5, 32.5))` raises `TypeError: 'float' object cannot be interpreted as an integer`. The PR body frames this knob's fix as replacing the opaque numpy error with a named `ValueError`; that holds only for ints `< 1`. An `isinstance(reservoir_size, int)` (or `int(reservoir_size) == reservoir_size`) check would cover it, and callers that catch `ValueError` for config errors get the right type.

**3. `low` — the dedupe key is exception-class only, so a second, different fault of the same class is never logged for the instance's lifetime.**
`src/snagline/ml/esn_ensemble.py:217`
`key = f"{self.name}:{phase}:{type(exc).__name__}"` carries no message or location, and `_fault_logged` has no expiry. A `ValueError` from one malformed event is logged once; a genuinely different `ValueError` bug surfacing 10,000 steps later is swallowed without even a debug line. `Monitor._log_fault_once` keys on the full message, so it distinguishes such faults; this collapses them. If the coarseness is deliberate, saying so in the docstring would prevent a future "why is nothing logged" hunt.

**4. `low` — third divergent copy of the fail-open fault-once pattern.**
`src/snagline/ml/esn_ensemble.py:210`, `src/snagline/monitor.py:674-683`, `src/snagline/ml_ensemble.py:59-66`
Monitor keys on message text and guards with `self._fault_lock`; MLOrchestrator keys on name+phase with no lock; this one keys on name+phase+class with no lock and logs a traceback. Three implementations of "log once" with three policies. A small shared helper would make the keying policy uniform — the same "reuse" note applies here as in the other two.

PR #412, opened 2026-09-21
fix(config): reject non-finite float knobs instead of deadening detectors (Fixes #383)

### LINKED ISSUE #383: bug: Config accepts inf/NaN for every float knob (env or JSON file), silently deadening the CUSUM detectors
## Summary

`Config._coerce` (config.py) accepts `inf`, `-inf` and `nan` for every float knob, from both `SNAGLINE_*` environment variables and JSON config files, and no validator downstream rejects them. A non-finite CUSUM knob does not crash — it silently disables the detector for the whole run, with no error and no log line. This is the same silent-deadening shape as #350 (NaN input data) and #331/#351 (zero divisors), but the NaN arrives from *configuration* rather than from adapter data.

## Reproduction

Against a fresh clone of master (`b667169`):

```python

cfg = Config(cusum_h=float("inf"), cusum_min_samples=3)   # or resolve() below
d = LatencyAnomalyDetector(config=cfg)
for i in range(3):
    d.observe(StepEvent("ep", f"s{i}", 0.0, "tool_call", "sig", latency_ms=10.0))
for i in range(20):
    r = d.observe(StepEvent("ep", f"b{i}", 0.0, "tool_call", "sig", latency_ms=10000.0))
    assert r is None      # 20 consecutive 10-second latencies vs a 10ms baseline: zero risks
```

The alarm test is `self.cusum > self.h`; `cusum > inf` is never true, so the detector can never fire again for that episode. With `h = nan` the comparison is likewise always false. With `sigma0 = nan` (a NaN sigma floor) `update()` computes `max(0.0, <NaN>)`, which is `0.0` — the accumulator is pinned at zero.

Every float knob accepts every non-finite value through `resolve()`:

```python
# value='inf' / 'nan' / '1e400' (which parses to inf) — all accepted, none rejected:
cusum_k, cusum_h, cusum_sigma_floor_abs, cusum_sigma_floor_rel,
token_cusum_k, token_cusum_h, goal_drift_error_tolerance, goal_drift_latency_k,
goal_drift_score_threshold, calibration_alpha, halt_timeout_s,
min_severity_for_halt, server_read_timeout, meltdown_low_entropy,
meltdown_high_entropy, semantic_drift_tolerance, semantic_drift_cusum_k,
semantic_drift_cusum_h
```

`warn_fraction` is the one float knob that is safe — `_validated_horizon` range-checks it, and `0.0 < inf <= 1.0` is False, so it is rejected. That is exactly the missing treatment for the rest.

A JSON config file is a second entry point, because the stdlib `json` module accepts the bare `Infinity` / `NaN` tokens by default:

```python
# file {"cusum_h": Infinity, "min_severity_for_halt": NaN}
cfg = Config.load_file(path)     # accepted: cusum_h == inf, min_severity_for_halt == nan
```

(TOML has no non-finite literal, so `load_file` on a `.toml` file is not affected.)

Observed impact:

| knob | value | effect |
| --- | --- | --- |
| `cusum_h` / `token_cusum_h` | `inf` or `nan` | latency / token CUSUM never alarms again (verified: 0 risks over 20 steps at 10000 ms and 30 steps at 200k tokens) |
| `cusum_sigma_floor_abs` | `nan` | `sigma0` becomes NaN; `max(0.0, NaN)` pins the CUSUM at 0 (verified: 0 risks over 20 steps at 5000 ms) |
| `min_severity_for_halt` | `inf` | no risk ever pays the halt-webhook cost, so the enforcement layer is unreachable |
| `halt_timeout_s` | `inf` | a halt consultation can block its own thread indefinitely instead of defaulting to continue |
| `server_read_timeout` | `0` | also accepted (finite but meaningless); 0 means non-blocking, not "no timeout" |

## Suggested fix

Reject non-finite values where they are coerced, rather than per knob. `_coerce` already raises `ValueError` for a misspelled bool or an unparseable int, and that path is already logged-and-ignored by `from_env_overrides`, so extending it is consistent:

```python
if hint is float:
    v = float(value)
    if not math.isfinite(v):
        raise ValueError(f"non-finite value: {value!r}")
    return v
```

plus the same `math.isfinite` guard on values coming from `load_file`, since `json.loads` accepts `Infinity`/`NaN` unless `parse_constant` is overridden. A `0` (or negative) `server_read_timeout` is a separate but adjacent gap worth range-checking at the same time.

## Severity

Moderate. It needs a misconfiguration to trigger, but the failure mode is the worst kind: the process starts cleanly, the banner prints, and the safety net is off for the entire run with nothing in the logs to say so — the operator's monitoring is silently inert.

Adhi1-2:
## Summary

Fixes #383. `Config` accepted `inf` / `-inf` / `nan` (and overflow literals like `1e400`) for every float knob, from both `SNAGLINE_*` env vars and JSON config files. Nothing downstream rejected them, and a non-finite CUSUM knob never crashed — it silently disabled the detector for the whole run:

- `cusum_h = inf`: the alarm test is `cusum > h`, and `cusum > inf` is never true, so the detector can never fire again for that episode.
- `cusum_sigma_floor_abs = nan`: `_floored_sigma` returns nan, and `update()` computes `max(0.0, nan)`, which is `0.0` in CPython — the accumulator is pinned at zero.

The failure mode is the worst kind: the process starts cleanly, the banner prints, and the safety net is off for the entire run with nothing in the logs to say so. `warn_fraction` was the one float knob that was safe, because `_validated_horizon` range-checks it and `0.0 < inf <= 1.0` is False. This gives the rest of the knobs the same treatment, at the point of entry rather than per knob.

## Changes

- **`_coerce`** rejects a non-finite float, the same place a misspelled bool or an unparseable int already fails. The env path keeps its existing un-coercible-value contract (issue #66): `from_env_overrides` logs the bad value at warning and drops it, so the built-in default survives and startup does not abort.
- **`load_file`** checks every parsed float value, raising `ValueError` naming the offending key. `json.loads` accepts the bare `Infinity` / `NaN` tokens by default, and `1e400` overflows to inf without a complaint; the check catches both. TOML has no non-finite literal, so the same check is inert there.

`server_read_timeout` was also listed in the issue, but it is already guarded at its use site (`_resolve_read_timeout` rejects non-finite and non-positive, failing open to the default), so it is not touched here.

## Tests

`tests/config/test_non_finite_knobs_383.py` (16 tests). Verified against pre-fix source in a scratch tree at `b667169` (the overlay recipe, not stash): **13 fail pre-fix, all 16 pass after.** The 3 that pass both ways are the intended positive controls — a finite env float still lands, a finite JSON float still loads, and the default threshold still alarms on the same burst that `cusum_h=inf` swallowed. They prove the failure-only assertions are narrow rather than a detector that never fires regardless.

Full suite: **864 passed, 4 failed**, unchanged from the pristine `b667169` baseline. The 4 are environmental, not regressions — they fail identically on unmodified upstream/master (a missing `snagline` console script, a sibling CONTINUUM checkout, and the openai/anthropic SDKs being installed in this venv). `ruff check`, `ruff format --check`, and `mypy src tests` are all clean at the known 11-error `auto/` baseline (optional-SDK modules invisible to CI).

## Notes

- An out-of-range config file now raises from the CLI with a traceback, the same as an out-of-range `warn_fraction` does today. Making the CLI exit 2 with a clean message is a wider change across every validation site and is out of scope for this fix.
- Unknown file keys with a non-finite value are also rejected. Keys are still ignored for *selection*, but a file carrying a bare `Infinity` is malformed regardless of where it sits, and accepting it under an unknown key would leave the guard trivially bypassable by nesting.


Cyrax321 (commented):
**Code Review — PR #412**

Well-targeted fix: the `_coerce` finite-check and the key-named `ValueError` in `load_file` close the env and JSON/TOML vectors (verified: `json.loads` accepts `Infinity`/`NaN`/`1e400`, `tomllib` parses `inf`/`nan`, and the shared check loop covers both branches). Tests are unusually rigorous with positive controls verified pre-fix. Two entry points the PR didn't touch still admit non-finite knobs and reproduce the exact silently-deadened-detector failure the issue describes.

Findings, most severe first:

**1. `medium` — `Config.__post_init__` does not run the finite-check, so direct construction and `dataclasses.replace` still admit inf/nan.**
`src/snagline/config.py:515-529`
Every other knob family is validated in `__post_init__` and re-run in `resolve()` (horizon, stagnation, token-runaway, max-live-episodes); the non-finite check is the only one placed solely at the two I/O adapters. Verified against the PR head: `Config(cusum_h=float('inf'))` constructs without error and `LatencyAnomalyDetector(config=cfg)` returns no risk across 20 consecutive 10000 ms observations after warm-up (`cusum > inf` never fires) — permanently dark, healthy process, zero log lines. Host code, the documented primary API ("a caller can retune everything from one object"), is exactly this path. Same for any knob set via `dataclasses.replace(cfg, ...)`.

**2. `medium` — CLI flags bypass the check entirely: `--halt-timeout inf` silently deadens halt enforcement.**
`src/snagline/monitor.py:300-301`, `src/snagline/cli.py:1053`
argparse `type=float` parses `inf` (and `1e400`) happily; `dataclasses.replace(cfg, halt_timeout_s=...)` bypasses `__post_init__`; `Monitor`'s only guard is `<= 0`, which `inf` passes. Then every `urlopen(timeout=inf)` raises `OverflowError` ("timestamp out of range"), caught by the fail-open handler at `monitor.py:617` — so above `min_severity_for_halt` every POST logs one "halt webhook failed; continuing" and the pause/halt control loop never engages. Startup is clean and nothing names the knob. A finite-check (or `isfinite` in the `halt_timeout_s` guard) at the CLI boundary would close it.

**3. `low` — the TOML rationale is wrong, and the guard's placement there is untested.**
`tests/config/test_non_finite_knobs_383.py:106`
TOML 1.0 defines `inf`/`nan` float literals — `tomllib.loads("cusum_h = nan")` returns NaN — so the PR body's "TOML has no non-finite literal, so the same check is inert there" is inaccurate. The code is safe only because the finite-check loop happens to sit after the JSON/TOML branch; a future refactor moving it into the JSON-only branch (a "simplification" the rationale invites) would re-open the vector with the suite still green, since only JSON files are tested. Add one TOML case.

**4. `low` — typo in a permanent test ID: `test_inf_cusum_h_no_longer_deadows_the_latency_detector` → "deadens".**
`tests/config/test_non_finite_knobs_383.py:50`
Cosmetic, but the malformed word ships in every future CI annotation and defeats `pytest -k deaden` searches against the term used in the issue, CHANGELOG, and module docstring.

Issue #395, opened 2026-09-20
bug: the network sinks' timeout= is per-socket-operation and skips DNS entirely, so emit() can block ingest for far longer than configured (verified: 2s budget, 36s actual)


abyyxhek:
## Summary

The three network sinks promise that "`emit` never raises and never blocks `ingest()` for long" (slack.py:4-6; the same wording in webhook.py and pagerduty.py) and pass a `timeout` to `urllib.request.urlopen`. That timeout does not bound the request. It is applied to individual socket operations only — and only *after* name resolution, which `socket.create_connection` performs before it ever calls `settimeout`. Two separate gaps follow, both of which leave the ingest thread parked on a network call while the monitor believes the sink is fire-and-forget.

## Reproduction

**1. The timeout is per-operation, not a deadline.** A server that answers 200 and then trickles its body one byte every 1.9 s never trips a 2.0 s read timeout — every individual read completes in time — so `emit` runs until the server finishes:

```
timeout=2.0s configured; emit() returned after 36.2s
a bounded request would have returned by ~2s: False
```

18× the configured budget, on a request the sink ultimately reported as successful.

**2. DNS resolution is not covered by the timeout at all.** `urllib` → `socket.create_connection` runs `getaddrinfo(host, port, ...)` and only then constructs the socket and applies `settimeout`. Against a black-holed or slow resolver — `SNAGLINE_WEBHOOK_URL=https://unresolvable.example.internal/...` is the realistic shape — `emit` blocks for as long as the resolver cares to hold it, which is unbounded and unrelated to any configured `timeout`. This part is established by reading `socket.create_connection` in the stdlib rather than by measurement, because triggering it needs a controlled slow resolver.

In both cases the sinks are dispatched synchronously from the ingest path unless the operator has separately wrapped them in `BatchingSink` — so a stuck sink stalls the episode's ingestion, and by extension the host agent's step, for as long as the network holds out.

## Suggested fix

Apply the timeout as a deadline rather than a per-operation hint. The minimal, dependency-free version is to resolve the host explicitly and bound the connect, then use a socket-level timeout for the exchange:

```python
import socket

deadline = time.monotonic() + self._timeout
host, port = urllib.parse.urlsplit(self._url).hostname, ...
family, _, _, _, addr = socket.getaddrinfo(host, port, proto=socket.IPPROTO_TCP)[0]
with socket.create_connection(addr, timeout=max(0.0, deadline - time.monotonic())) as sock:
[...]

Issue #389, opened 2026-09-20
bug: webhook/Slack/PagerDuty sinks follow 301/302 redirects as a bodyless GET, so a redirecting endpoint silently drops the alert and reports success


abyyxhek:
## Summary

All three network sinks pass a URL to `urllib.request.urlopen` and let the stdlib follow redirects. `urllib` honours a `301`/`302`/`303` by re-issuing the request to the `Location` URL **as a GET with no body**, and returns the final 2xx to the caller. The sink sees a successful delivery, logs nothing, and the alert payload was silently discarded and sent to a destination the *server* chose. A `307`/`308` instead raises `HTTPError` and is caught and logged — so the failure mode is inconsistent, and silent for exactly the common redirect codes.

## Reproduction

Against a fresh clone of master (`b667169`), with a local server that answers 302 to a POST:

```
requests received by the server:
   POST /webhook b'{"text": "[CRITICAL] SNAGLINE failure detected\nTrigger: loo'
   GET /elsewhere b'<empty>'
alert payloads that reached a POST body: 1   <- but at /webhook, not /elsewhere
```

The alert body travelled with the POST that the server *rejected*; the request that the sink counted as success was the bodyless follow-up. With `Monitor.fail_open=True` (the default) the risk is simply gone.

Relevant code, identical in all three sinks — `webhook.py:73-75`, `slack.py:71-72`, `pagerduty.py:88-89`:

```python
with urllib.request.urlopen(req, timeout=self._timeout) as resp:
    resp.read()
```

No sink installs a redirect policy (a `NoRedirect`-style handler), and none inspects the final status or the final URL afterwards.

## Why this is reachable without a misconfigured destination

- A trailing-slash redirect from the provider (`https://hooks.slack.com/services/...` → the canonical path) is a 301/302 in the wild.
- An HTTP→HTTPS or host-canonicalisation hop in a corporate proxy.
- A PagerDuty or webhook endpoint in front of a login redirect when the credential is wrong or expired.

That last one is the worst pairing: the alert is not just mis-routed, the operator is told nothing about an auth failure that would have been visible as a 401.

## Suggested fix

Install a handler that refuses to follow redirects, then treat the resulting `HTTPError` exactly like any other failure (it already logs fail-open):

```python
class _NoRedirect(urllib.request.HTTPRedirectHandler):
    def redirect_request(self, req, fp, code, msg, headers, newurl):
        return None
opener = urllib.request.build_opener(_NoRedirect)
```

[...]

PR #418, opened 2026-09-21
fix(monitor): bound the halt webhook round trip and refuse its redirects (#415, #416)

### LINKED ISSUE #415: bug: the halt-webhook POST (and `snagline hook --url`) use urlopen's per-socket timeout, so a stalled endpoint parks the agent's step past its budget
## What's wrong

`Monitor._run_halt_webhook` POSTs the halt decision request with:

```python
with urllib.request.urlopen(req, timeout=self._halt_timeout_s) as resp:
    body = resp.read(_MAX_HALT_RESPONSE_BYTES)
```

`urlopen`'s `timeout=` is applied **per socket operation**, and only after
`socket.create_connection` has finished name resolution. Two gaps follow:

1. **Name resolution is not covered at all.** A slow or black-holed resolver
   holds the call for as long as it likes, with no timeout in force.
2. **A trickling body never trips the read timeout.** A server that sends one
   byte every interval just under `timeout` completes every individual read
   inside the budget while the *total* grows without bound.

This is the identical defect to #395, which measured a configured 2.0 s budget
taking 36.2 s on exactly that trickle, on a request reported as successful.

## Why it matters more than the sinks

The sinks are fire-and-forget, and PR #8 already moved their dispatch outside
the ingest lock. The halt webhook is different: it is the **enforcement
policy**, called synchronously from `ingest()`:

```
ingest() -> _dispatch(risk) -> _enforce(risk) -> _run_halt_webhook(risk)
```

The docstring on `_enforce` says a slow halt webhook "delays only its own
dispatch thread, never other episodes' ingest". That is accurate about *other*
episodes, but this episode's `ingest()` still has not returned, so the **host
agent's step is parked** for however long the endpoint stalls. An operator who
set `halt_timeout_s` to the 0.25 s default to keep the enforcement round-trip
out of the critical path gets an unbounded stall instead, and `policy_errors`
never increments because no exception was ever raised.

`snagline hook --url` (`cli.py`) has the same shape and blocks the hook
process the same way.

## Suggested fix

Route both call sites through the `bounded_post` helper that #413 added to
`sinks/base.py`, which bounds the whole exchange by a wall-clock deadline on a
daemon thread. It currently drains and discards the reply, so it needs to
return the body (and take the `max_bytes` cap the halt path already enforces)
for the halt webhook to parse its directive.

`bounded_post` raises `TimeoutError` on overrun, which the halt path's existing
`except Exception` already counts as a `policy_error` and logs fault-once, so
the fail-open contract is unchanged.

### LINKED ISSUE #416: bug: the halt-webhook follows a redirect, so the enforcement directive can come from a server the operator never configured
## What's wrong

`Monitor._run_halt_webhook` POSTs the halt decision request through bare
`urllib.request.urlopen`. `urllib` honours a `301`/`302`/`303` by re-issuing
the request to the `Location` URL **as a GET with no body** (see
`HTTPRedirectHandler.redirect_request`, which returns
`Request(newurl, method="GET", ...)` with no `data`) and hands the final `2xx`
back to the caller. The halt path then parses that body:

```python
with urllib.request.urlopen(req, timeout=self._halt_timeout_s) as resp:
    body = resp.read(_MAX_HALT_RESPONSE_BYTES)
parsed = json.loads(body.decode("utf-8"))
action = str(parsed.get("action", HALT_ACTION_CONTINUE)).strip().lower()
...
directive = HaltDirective(action=action, reason=reason, timestamp=time.time())
```

So when the configured halt endpoint redirects, the enforcement **directive is
read from a server the operator never configured** — the very body the POST
carried was rejected, and the decision came back from the redirect target.

This is the identical defect to #389, but the consequence is worse there than
at a sink: a sink silently misroutes an *alert*, while `halt_webhook` silently
misroutes the *enforcement decision*. A `{"action": "halt", ...}` from the
redirect target sets `last_directive` and halts the episode on a response from
a host nobody pointed at.

The reachable cases are the mundane ones, as in #389: a trailing-slash hop, an
HTTP->HTTPS or proxy canonicalisation — or, worst, an **auth redirect from an
expired credential**, which converts a visible `401` into a silent misrouting
that still looks like a successful policy round-trip.

`snagline hook --url` (`cli.py`) follows redirects the same way; there the
payload is the step event, which leaks to the redirect target.

## Suggested fix

`sinks/base.py` already carries a `_NoRedirect` handler on `bounded_post`'s
opener (#414), which returns `None` from `redirect_request` so the `3xx`
surfaces as `HTTPError`. Routing the halt webhook and the hook forward through
`bounded_post` extends that policy to both call sites for free. The halt
path's existing `except Exception` already counts an `HTTPError` as a
`policy_error` and continues fail-open, so the operator sees the misrouting
instead of a directive from the wrong host.

Adhi1-2:
## What's wrong

`Monitor._run_halt_webhook` and the `snagline hook --url` forward both posted
through bare `urllib.request.urlopen`. Two inherited defects followed, and both
are **worse here than they were at a sink** (#395 / #389):

1. **`timeout=` is a per-socket hint, applied only after name resolution.** A
   stalled resolver or a body trickled one byte per interval just under the
   timeout parks the exchange far past the configured `halt_timeout_s`, and no
   `policy_error` is counted because nothing ever raised. The halt webhook is
   the enforcement policy, called synchronously from `ingest()`:

   ```
   ingest() -> _dispatch(risk) -> _enforce(risk) -> _run_halt_webhook(risk)
   ```

   So the stall is the **host agent's own step**, while the operator's budget
   reads 250 ms. (`_enforce`'s docstring is accurate that other episodes are
   unaffected — this episode's `ingest()` still hasn't returned.) #415
2. **A 301/302/303 is rewritten into a bodyless GET whose final 2xx is handed
   back.** The halt path parses that reply into a `HaltDirective`, so a
   redirect let the **enforcement decision arrive from a server the operator
   never configured**, on a round trip that reported success. An auth redirect
   from an expired credential — which would otherwise have shown as a `401` —
   became a silent misrouting. #416

While in the same two blocks, the failure path logged the halt URL and the
hook URL verbatim; both can carry basic auth, and that line is what an operator
reads when enforcement stops working. The `snagline serve` startup banner had
the same leak for the same reason — it echoes `--halt-forward` to stderr, which
a process supervisor captures and keeps after the process is gone (#390).

## The fix

Both call sites now go through the shared `bounded_post` in `sinks/base.py`,
added for #395 and given the `_NoRedirect` handler in #414 — one place, and the
policy that already covers the three network sinks now covers the enforcement
path too. `monitor.py` already imported from `sinks.base`, so this adds no new
dependency direction.

`bounded_post` needed two small generalisations, both backwards compatible:

- it **returns the reply body** (the three sinks ignore it; the halt webhook
  parses it), and
- it takes an optional **`max_bytes`** cap on the read, so an endpoint
  streaming an endless body cannot unbound the exchange by another route. The
  halt path passes the `_MAX_HALT_RESPONSE_BYTES` cap it already enforced.

`bounded_post` raises `TimeoutError` on overrun and `HTTPError` on a 3xx; the
halt path's existing `except Exception` counts either as a `policy_error` and
continues fail-open, so the failure posture is unchanged — only the *kind* of
failure that now surfaces.

## Tests

New `tests/test_halt_webhook_bounded_post_415_416.py` (8 tests) plus one in
`tests/test_cli.py`, behavioural through real loopback servers because the
claims are about urllib's machinery end to end:

| test | pre-fix |
|---|---|
| a trickling endpoint does not hold ingest past its budget | **fails** |
| a redirect cannot set the directive (or its reason) | **fails** |
| the payload does not leak to the redirect target | **fails** |
| the halt URL is not logged with its credentials | **fails** |
| the serve banner does not echo the halt URL's credentials | **fails** |
| the halt webhook goes through `bounded_post` | **fails** (guard) |
| `bounded_post` returns the body and honours the cap | **fails** (signature) |
| a stalled endpoint is abandoned at the deadline | passes |
| a direct 2xx still sets the directive and logs nothing | passes |

The last two are the positive controls — they pass both before and after, so
the deadline and redirect changes add no failure to the working path.

Two tests took care to get right:
- the misrouted directive had to use `"pause"`, not `"halt"` — `VALID_HALT_ACTIONS`
  is `("continue", "pause")`, so `"halt"` was rejected by an unrelated safety net
  and the bug never surfaced;
- the trickle interval must sit *under* the per-socket timeout (0.05 s against
  a 0.5 s budget), otherwise the read timeout fires and the exchange is bounded
  by accident.

## Verification

- `pytest`: **909 passed, 2 failed**. The 2 are the known environmental
  baseline on this machine (a missing `snagline` console script, and an
  `auto/` no-SDK test that needs the OpenAI SDK absent), unrelated to this
  change.
- `ruff check src tests` and `ruff format --check src tests`: clean (143 files).
- `mypy`: no new errors. Locally the pinned `--python-version 3.10` chokes on
  numpy 2.5.3's stubs, so this was verified under 3.12, where the only errors
  are the 7 pre-existing ones in `auto/openai.py` (the OpenAI SDK is installed
  here; CI's mypy jobs install no SDKs, so neither artifact reaches CI).
- The `snagline hook --url` path is covered end to end by the existing
  `tests/test_hook_bridge.py`, including its failure case.

## Stacking

Based on #414's head (`af6f042`), which is based on #413's head (`d282d00`).
All three live in `sinks/base.py` + the POST call sites, so the stack keeps
them as one clean change each. If #413 lands first, the others rebase to a
no-op; the three can equally be squashed together.

Fixes #415. Fixes #416.


Cyrax321 (commented):
**Code Review — PR #418**

The design is sound: the halt webhook and the `snagline hook --url` forward now share `bounded_post`, which enforces a wall-clock deadline on a daemon thread and refuses redirects, with credential-bearing URLs redacted on failure paths. #415/#416 are correctly fixed and tested. The remaining findings are about the deadline's real-world strength and two inconsistencies in the same failure lines this PR cleans up.

Findings, most severe first:

**1. `medium` — the wall-clock deadline is only a join timeout; the abandoned worker keeps the socket and response alive.**
`src/snagline/sinks/base.py:152-163`
`worker.join(timeout)` merely abandons the still-running thread. The per-request `timeout=`, `getaddrinfo` DNS resolution, TLS handshake, and `resp.read()` loop each run on their own; a halt endpoint that accepts the connection and trickles the body one byte per interval keeps the worker looping in `resp.read()` forever (socket reads keep succeeding within their own timeout). Every halt POST to a sick endpoint then leaks one daemon thread plus an open socket, reaped only at process exit — a monitor ingesting for days accumulates them. `ingest()` unblocks on time, but the resource is not bounded, and wall-clock can still overshoot when a single per-operation timeout exceeds the configured budget.

**2. `medium` — `resp.read(max_bytes)` silently truncates a longer valid reply into a JSON parse error.**
`src/snagline/sinks/base.py:156-158`, `src/snagline/monitor.py:617`
A halt endpoint returning a legitimate directive larger than 65,536 bytes (e.g. an embedded remediation policy) has its body cut at the cap; `json.loads` then raises `JSONDecodeError` on the truncated stream, the directive is discarded, `policy_errors` increments, and the episode runs fail-open even though the operator's policy answered in time. A `len(body) == max_bytes` check (or honouring `Content-Length`) would distinguish "server sent too much" from "server faulted" and could at least log the truncation.

**3. `low` — a `BaseException` in the worker bypasses `except Exception`, so `bounded_post` returns an empty body instead of raising.**
`src/snagline/sinks/base.py:154,169`
The helper's contract is "raises whatever the exchange raised", but `MemoryError`/`KeyboardInterrupt` delivered inside the thread escapes `except Exception`, leaving `outcome` with neither `body` nor `error`; `outcome.get("body", b"")` then returns empty bytes and the halt path mis-reads it as a JSON decode failure. Rare, and downstream handling degrades safely, but the contract is not met.

**4. `low` — the hook forward's failure line still interpolates `{exc}`, unlike the sinks that were converted to class-name-only for exactly this reason.**
`src/snagline/cli.py:649-651`
The line redacts the destination (`redacted_destination(args.url)`) but then appends `str(exc)`, which for `URLError` embeds the destination (and can carry userinfo/secret path fragments). The sinks in this same PR switched to `describe_failure` "because a URLError embeds the destination URL in its reason" — the third call site (#390) was left inconsistent, so a credential can still ride into stderr on the same failure line the redaction targets.

**5. `low` — `tests/test_slack_sink.py:98` still patches `urllib.request.urlopen`, a seam SlackSink no longer uses.**
`tests/test_slack_sink.py:98`
Since SlackSink now goes through `bounded_post` → `_opener.open`, the patched `urlopen` is never invoked, so `boom()` cannot fire and the test passes vacuously — the URLError-embeds-URL regression it exists to guard could return silently. The sibling webhook test correctly patches `_opener.open`; retarget this one the same way.

PR #419, opened 2026-09-21
fix(detectors): make token_runaway and stagnation load_state atomic (#417)

### LINKED ISSUE #417: restore: three load_state implementations are not atomic, so a rejected entry leaves the detector half-restored
PR #402 made `Monitor.restore_dict` contain a malformed per-detector entry: it catches the exception, logs `detector %r keeps its live state`, and lets the rest of the restore continue. That is the right call overall — pre-#402 the exception aborted the restore halfway, so every detector after the failing one kept live state and the sink/time-axis restoration never ran.

But the log line's guarantee does not hold for every detector. Three `load_state` implementations clear live state first and then populate per-episode **in a loop**, so a malformed entry mid-payload leaves the detector neither restored nor live, but half-cleared — and the containment layer reports it as the third thing.

```python
# detectors/meltdown.py
def load_state(self, state):
    self._eps = {}                       # live state discarded here
    for ep, keys in state.get("windows", {}).items():
        ...
        self._eps[ep] = w                # published per-episode
    self._counts = {...}                 # never reached if the loop raised
    self._fired = {...}
    self._clear_streak = {...}
```

Repro (meltdown):

```python
>>> d = MeltdownDetector()
>>> d.load_state({"windows": {"ep-a": ["a","b"], "ep-b": ["c"]},
...               "counts": {"ep-a": 2, "ep-b": "not-an-int"},
...               "fired": {}, "clear_streak": {}})
ValueError: invalid literal for int() with base 10: not-an-int
>>> d._eps        # partially restored
[ep-a]
>>> d._counts     # never set
{}
```

`token_runaway` is the same shape (`self._states = {}` then a per-episode loop; a `KeyError' on one episode leaves the earlier ones restored and `_totals` / `_warned` / `_breached` unset). `stagnation` assigns `_counts` first (atomically) then builds `_windows` in a loop, so a rejected entry can leave **counts without windows** — the scaler believes an episode has N steps of history while its window is empty.

Consequences after a rejected entry: lost dedupe (`_fired` never restored, so a known anomaly can re-alarm), a scaler positioned by counts that no longer match the windows it sizes, and — in the meltdown/token_runaway case — live episodes silently discarded. None of this fails loudly; the warning says the detector kept its live state.

`error_cascade`, `latency_anomaly`, and `side_effect_guard` already do this right: build everything into locals and publish in one assignment (`error_cascade` carries a comment crediting the review of #402). The fix is to bring the three above to that pattern, and to soften `restore_dict`'s message so it does not assert a guarantee the callee can break ("detector %r was not fully restored" rather than "keeps its live state").

This only matters now because #402 made per-detector rejection reachable; before it, the exception aborted everything and the question never arose.

Adhi1-2:
## What

Follow-up to #402, which taught `Monitor.restore_dict` to contain a malformed per-detector entry: catch the exception, log, let the rest of the restore continue, and report the detector as keeping its live state.

That last guarantee only holds when `load_state` publishes its whole result at once. Two detectors did not:

- **`TokenRunawayDetector`** cleared `_states` and repopulated it per episode.
- **`StagnationDetector`** assigned `_counts` from the snapshot, then built `_windows` in a loop.

A malformed entry raising partway through left the detector neither restored nor live:

```python
>>> d.load_state({"states": {"ep-bad": {"n": "not-an-int"}}, ...})   # after live traffic on "ep"
KeyError: mean
>>> d._states       # live episode discarded, nothing rebuilt in its place
{}
>>> d._breached     # still describing the episode that no longer exists
{ep: True}
```

For stagnation the equivalent is `_counts` from the snapshot paired with `_windows` still holding the live ones — the scaler then believes an episode has N steps of history its window no longer carries.

## The fix

Both build into locals and publish in one assignment, matching the three implementations that already do (`error_cascade`, `latency_anomaly`, `side_effect_guard`) and `MeltdownDetector`, which #406 fixed the same way. A rejected snapshot now leaves the detector exactly as it was.

## Tests

Two regression tests (`test_load_state_is_atomic_when_an_entry_is_malformed`) plus a positive control each. The bug-pinning shape that works: a detector holding **live** state, plus a snapshot that mentions a *different, malformed* episode — the clear-then-populate discards the live episode before the bad entry raises. Pre-fix, both fail on exactly the preservation assertions (`a rejected snapshot must not discard live episodes` / `... live windows`); post-fix all 36 in the two files pass.

A first attempt at the token_runaway test failed pre-fix for the wrong reason — an `_warned` precondition that was never populated — so it pinned nothing. Fixed to assert `_states` / `_totals` / `_breached`, which the four 500-token events actually populate.

## Scope

`meltdown` is not touched: #406 already made its `load_state` publish-once (with a comment crediting the review of #402). Neither file conflicts with an open PR — `stagnation.py` is untouched by any open branch, and the one PR touching `token_runaway.py` (#363) changes the module header, not `load_state`.

The `restore_dict` wording (`"detector %r keeps its live state"`) still overstates the guarantee for these two; it is one line in #402 itself, where the message lives, so it is left there rather than duplicated here.

Fixes #417

## Summary by CodeRabbit

* **Bug Fixes**
  * State restoration now applies atomically for token runaway and stagnation detection.
  * Invalid snapshots no longer partially overwrite existing detector state.

* **Tests**
  * Added coverage confirming invalid snapshots preserve live state.
  * Added coverage confirming valid snapshots restore detector state correctly.


Cyrax321 (commented):
**Code Review — PR #419**

The core fix is correct and well-tested: both `StagnationDetector.load_state` and `TokenRunawayDetector.load_state` now parse the whole snapshot into locals and publish only on success, so a malformed entry raising mid-parse leaves the live state intact (verified against the PR head: reverting the two source files fails exactly the two new tests; `tests/detectors` is 131-passed green). The findings below are about the changelog/body overclaiming the fix's coverage and about untested integration assumptions.

Findings, most severe first:

**1. `medium` — the changelog claims the meltdown half of #417 "was already fixed in #406", but #406 is still open.**
`CHANGELOG.md:24`
PR #406 is `OPEN`/unmerged, and base `master`'s `meltdown.py` still has the clear-then-populate `load_state` (`self._eps = {}` before the per-episode loop). If this PR merges before #406, a malformed meltdown entry still clears live `_eps` while `_counts`/`_fired` keep live values — yet the release notes tell operators restart-safe restore is fixed. #417's repro is literally meltdown's. Reword to reference #406 as a dependency, not a done deal.

**2. `medium` — `error_cascade.load_state` (cited as the already-atomic model) is not atomic.**
`src/snagline/detectors/error_cascade.py:152-169`
The PR body says `error_cascade` (with latency_anomaly/side_effect_guard) "already publishes atomically", but it assigns `self._windows` (152) then `self._counts` (165, `int(n)` coercion) then `_consecutive`/`_fired` sequentially. A snapshot with a non-int count raises inside the `_counts` comprehension *after* `_windows` was replaced, leaving snapshot windows paired with live counts — exactly the half-restore #417 describes, in the very detector held up as the correct pattern (`loop.load_state` has the same shape).

**3. `medium` — the containment premise ("a rejected snapshot keeps its live state" being non-fatal) lives in unmerged #402, not here.**
`src/snagline/monitor.py:918`
On base master, `restore_dict` calls `load(entry)` unguarded, so a malformed entry aborts the whole restore — detectors after the failing one and the sink/time-axis restore never run. The new tests call `load_state` directly and never exercise the Monitor path the fix is justified by. This is fine as long as #402 lands first, but the PR should state that dependency.

**4. `low` — other `load_state` implementations remain non-atomic.**
`src/snagline/detectors/compaction_tripwire.py:183-199`, `src/snagline/detectors/silent_abort.py:92-103`
`compaction_tripwire` clears `_episodes` then repopulates in a loop (a malformed entry leaves it with zero episodes); `silent_abort` adopts the snapshot's `output_action_types` before building `_last`. The containment log's guarantee ("keeps its live state") is false for these detectors; same pattern, same fix.

**5. `low` — "published in one assignment" is exception-atomic, not reader-atomic.**
`src/snagline/detectors/token_runaway.py:190-193`, `src/snagline/detectors/stagnation.py:241-242`
The code performs four (resp. two) sequential assignments with no lock held; a concurrent reader can interleave and see snapshot `_states` paired with live `_totals`/`_breached`. Harmless today because `restore_dict` is setup-time, but the stated invariant overstates what the code guarantees.

PR #420, opened 2026-09-21
fix(sinks): keep the POST body across a same-host redirect (#389)

### LINKED ISSUE #389: bug: webhook/Slack/PagerDuty sinks follow 301/302 redirects as a bodyless GET, so a redirecting endpoint silently drops the alert and reports success

Adhi1-2:
## What

`urllib`'s redirect handler rewrites **any** POST that meets a 301/302/303 into a bodyless `GET` — it builds the replacement `Request` with `method="GET"` and drops `data` and the content headers. It does the same for 307/308, which the RFCs say must preserve the method. The redirect target then answers that `GET` with 200, `emit` never raises, and the alert was never delivered.

An endpoint that moved, or one that wants the trailing slash the operator left off, silently stops paging. There is no error anywhere — the sink believes it succeeded.

## Reproduction

A local `http.server` that 302s `/start` to `/final`, with the old code:

```
POST /start  -> 302
GET  /final  -> 200      # the body was dropped here; emit() returned cleanly
```

The reproduction is a real server rather than a mock because the whole failure is in the interaction between `urllib`'s handler chain and the 302 — mocking `urlopen` mocks the bug away along with the call.

## The fix

`bounded_post` now uses a private opener carrying a `_SameHostRedirectHandler` that re-issues the redirect as a `POST` with the original body, `Content-Type` preserved and `Content-Length` recomputed (which is what makes a JSON body still arrive as JSON).

### The credential restriction

The redirect is **only** followed when it points at the same scheme, host and port. These sinks POST a credential:

- `PagerDutySink` — the `routing_key` *is* the body
- `WebhookSink` — the URL can carry it in the userinfo (`user:pass@host`)
- `SlackSink` — the URL embeds it as the final path segment

A redirect to a different host hands that credential to whoever answers there. That is an exfiltration path, not a delivery path, so an off-origin redirect **raises** instead, naming the redacted destination through the existing `redacted_destination` helper. The sink still logs it fail-open — the operator gets a nameable thing to fix, which is strictly better than the alert quietly vanishing.

A private opener rather than `install_opener`: that mutates the process-global default and would change the host application's own HTTP semantics — the off-host POST raise would start firing on the host's own requests. A monitoring library embedded in a host must not do that.

## Tests

`tests/sinks/test_redirect_body_389.py` — 8 tests against a real local server, covering all three network sinks, plus scope guards:

- the body and `Content-Type` survive the redirect, and no `GET` is ever issued
- an off-origin redirect delivers nothing to the other host, does not raise out of `emit`, and logs the destination without the routing key
- no redirect → a single POST, unchanged (regression guard)
- a redirect loop is still bounded by `urllib`'s own counter

**Mutation check:** 6 of the 8 fail with the fix reverted; the 2 that pass are the scope guards, which is the correct shape.

Three existing test files retarget their transport seam from `urllib.request.urlopen` to `snagline.sinks.base._opener.open`; the assertions are unchanged.

## Verification

- `pytest`: **899 passed, 2 skipped**, 88.06% coverage
- `ruff check src tests` / `ruff format --check src tests`: clean
- `mypy src tests`: clean

## Stacking

This branch is based on the tip of #413, which is itself based on #404 — both touch `sinks/base.py` and the sink test seams. If they land in a different order I will rebase.

Fixes #389


Adhi1-2:
Closing as a duplicate of #414, which I opened earlier today for the same issue — I lost track of it mid-session and re-derived the same bug as unclaimed.

The two PRs differ on policy, and #414's is the one I want to keep:

- **#414** — refuse *any* redirect; the `3xx` surfaces as `HTTPError` and is logged fail-open.
- **#420** (this) — follow a redirect on the same scheme/host/port with the body preserved; raise off-host.

For a POST whose body is a credential, "never send it anywhere the operator did not point at" is the guarantee worth having, and it costs only the convenience of an automatic trailing-slash or HTTP→HTTPS hop — which surfaces as a loud, named failure the operator fixes once, rather than the credential travelling a path nobody configured. Refusing also makes the failure mode uniform across `301/302/303/307/308` instead of silent for exactly the common codes and loud for the rare ones.

#415 and #416 are also already specified against `_NoRedirect`, so keeping #414 keeps that follow-up plan coherent.

No maintainer attention needed on this one — sorry for the noise.

PR #422, opened 2026-09-21
fix(config): reject a negative CUSUM slack, which inverts the alarm (Fixes #421)

### LINKED ISSUE #421: bug: a negative cusum_k / token_cusum_k inverts the CUSUM and storms false positives on perfectly stable data
## Summary

`cusum_k` (latency anomaly) and `token_cusum_k` (token runaway) are the CUSUM
*slack*: each scored step subtracts `k` from the accumulator, tuning how many
sigma of sustained drift the detector needs before it alarms. Both accept any
finite value — a negative one is silently accepted and **inverts the detector**:
the per-step subtraction becomes an addition, so the accumulator grows by `|k|`
on every step regardless of the observed data, and the detector storms
false-positive risks on perfectly healthy runs.

`cusum_h` / `token_cusum_h` (the alarm bars, and the denominators of the score
formulas) are range-checked in #337. The `k` knobs are not checked anywhere —
neither in `Config` nor at detector construction. #412 rejects *non-finite*
floats, but a finite `SNAGLINE_CUSUM_K=-1.0` sails through both layers.

This is the same defect class as #386 (EsnCusumDetector) and #370 (semantic
drift), both accepted — but in the two *core* detectors that ship enabled by
default in every monitor.

## Repro

```python

d = TokenRunawayDetector(k=-1.0, min_samples=5, config=Config())
risks = [r for i in range(30)
         if (r := d.observe(StepEvent(step_id=str(i), episode_id="ep",
                    timestamp=float(i), action_type="tool_call",
                    action_signature=f"s{i}", tokens_in=100, error=False))) is not None]
print(len(risks), sorted({round(r.score, 2) for r in risks}))
```

Output — 30 steps of an *exactly constant* 100 tokens/step, no anomaly anywhere:

```
20 [0.62, 0.64, 0.66, ..., 0.98, 1.0]
```

Twenty risks in thirty steps, score climbing to the maximum 1.0. The latency
detector does the same on constant latency (`SNAGLINE_CUSUM_K=-1.0`, constant
`100.0` ms tool latency, `action_type="tool_call"`): the accumulator climbs
`1.0, 2.0, ... 25.0` and alarms from step 10 to the end. The `k` term is added
every step, so the alarm is *unavoidable* once enough steps elapse — no data
pattern can prevent it, and no amount of baseline accuracy helps.

The user-visible symptom is an alert storm that looks like the run is burning
out of control when it is perfectly flat, plus score-1.0 `latency_anomaly` /
`token_runaway` risks reaching halt policy (`min_severity_for_halt=0.8`), so a
typo'd env var can page or halt a healthy agent.

## Expected behaviour

Following the precedent set by #317 / #132 / #386 (a knob whose range can
silently deaden or invert a detector is a configuration error, checked both in
`Config` after env/file layering and again at detector construction, because
direct kwargs skip the `Config` check): a negative `k` should be rejected with a
message naming the knob. `k = 0` is legitimate (no slack — maximally sensitive,
but not inverted) and must keep working, matching the `>= 0` bound #371
established for `semantic_drift_cusum_k`.

## Suggested fix

In `Config._validated_cusum` (introduced by #337), alongside the existing
`cusum_h` / `token_cusum_h` loop, add `cusum_k` / `token_cusum_k` with a
`< 0` check — a negative slack adds to the accumulator instead of subtracting.
Mirror the guard in `LatencyAnomalyDetector.__init__` and
`TokenRunawayDetector.__init__` for direct construction.

Adhi1-2:
## What

`cusum_k` (latency anomaly) and `token_cusum_k` (token runaway) are the CUSUM
*slack*: each scored step subtracts `k` from the accumulator. Both accepted any
finite value, and a negative one **inverts the detector** — the subtraction
becomes an addition, so the accumulator grows by `|k|` per step regardless of
the data and the alarm becomes unavoidable once enough steps elapse.

## The failure

30 steps of an *exactly constant* 100 tokens/step — no anomaly anywhere — with
`token_cusum_k=-1.0`:

```
20 risks/30 steps, scores [0.62 .. 1.0]
```

Twenty risks, climbing to the maximum score. The latency detector is identical
on constant latency (`cusum_k=-1.0`, constant 100 ms): the accumulator climbs
`1.0, 2.0, ... 25.0` and alarms from step 10 onward. Because the `k` term is
added every step, **no data pattern can prevent it and no baseline accuracy
helps** — the alarm is a function of step count alone. Score-1.0
`latency_anomaly` / `token_runaway` risks also clear `min_severity_for_halt`,
so a typo'd env var can page or halt a healthy agent.

## Why this is a gap in the existing surface

- `cusum_h` / `token_cusum_h` (the alarm bars) are range-checked in **#337**.
- `semantic_drift_cusum_k` is range-checked in **#371** — with a `>= 0` bound.
- Non-finite floats are rejected in **#412**.

But a *finite negative* `SNAGLINE_CUSUM_K=-1.0` sailed through every layer:
no `Config` check, no detector guard. This closes that gap in the two **core**
detectors that ship enabled in every monitor, following the #317 / #132 / #386
precedent — a knob whose range can silently deaden or invert a detector is a
configuration error.

## Changes

- `Config._validated_cusum` (#337's validator): added `cusum_k` / `token_cusum_k`
  with a `< 0` check, raising a `ValueError` naming the knob. Checked at
  construction, `from_env`, and `resolve`, so every config layer is covered.
- `LatencyAnomalyDetector.__init__` and `TokenRunawayDetector.__init__`: the
  same guard, because direct construction with explicit kwargs skips the
  `Config` check — and both detectors are constructed directly by library
  embedders and throughout the test suite.
- `k = 0` is deliberately **still valid**: it means "no slack" (the accumulator
  is never decayed — maximally sensitive, but not inverted), matching the `>= 0`
  bound #371 established for the semantic-drift detector. A new positive
  control asserts `k=0` pages zero times on constant traffic, on both detectors.

## Verification

New tests in `tests/test_issues_331_333.py` (appended to #337's file, since the
validator lives there):

- `Config` / `from_env` / `resolve` reject `cusum_k` / `token_cusum_k` in
  `{-0.001, -1.0, -100.0}` and accept `{0.0, 0.001, 0.5, 2.0, 100.0}`.
- Both detectors reject a negative slack at direct construction.
- `Monitor.default` refuses the config before any risk is dispatched (an
  end-to-end layer the env var actually hits).
- Positive control: `k=0` on constant latency and constant token volume pages
  zero times in 30 steps.

**Pre-fix gate** (base `f7283fd` with the tests overlaid): **9 fail / 11 pass**.
The 9 failures are exactly the bug-pinning rejections; the 11 passes are the
positive controls, which pass both sides — so the failing assertions are narrow
rather than over-broad.

Full suite on the fixed tree: **975 passed**, plus the 2 known pre-existing
env-only failures (editable-install console script; `openai` installed locally
where CI installs only `langchain-core`). `ruff check` / `ruff format` clean.
`mypy` identical to the pre-fix base — 7 errors, all in `src/snagline/auto/openai.py`,
the known local-only artifact.

## Stacking

Branched from **#337's head `f7283fd`**, because the `k` check belongs in the
same `_validated_cusum` function #337 introduces. #337 is approved and 13/13
green; extending it directly would have wasted that approval and re-triggered
review, so this stacks on top and lands with a dependency. If #337 is reworked,
this rebases cleanly — the change is one loop plus two constructor guards.

Fixes #421.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation for detector thresholds, slack values, sigma floors, sample counts, and window sizes.
  * Prevented invalid settings from causing false alerts, uncontrolled accumulator growth, or unstable scoring.
  * Improved baseline version retention validation to prevent accidental loss of stored history.
  * CLI now reports invalid retention settings as usage errors.
  * Ensured stable fallback behavior when observed latency variation is zero.
  * Documented validation rules and CUSUM slack behavior in the changelog.


Cyrax321 (commented):
**Code Review — PR #422**

The change correctly closes the #421 gap: a negative `cusum_k` / `token_cusum_k` is now rejected in `Config._validated_cusum` (constructor, `from_env`, and `resolve`) and defensively in both detector constructors, with `k=0` kept valid and a positive control. The core fix is right and the rejection tests are thorough. Two gaps remain in the *added* guards, plus a stack-level merge conflict worth flagging before this lands.

Findings, most severe first:

**1. `medium` — the new constructor "defense in depth" mirrors only `k`, not the sibling denominator `h`.**
`src/snagline/detectors/latency_anomaly.py:212` and `src/snagline/detectors/token_runaway.py:79`
The added guards check `self.k < 0` but say the check "Mirrors `Config._validated_cusum`". That validator also rejects `h <= 0`, yet the detector constructors don't. Direct construction with `h=0` still reaches `alarm_cusum / self.h` (`latency_anomaly.py:291`) and `state.cusum / self.h` (`token_runaway.py:137`) → `ZeroDivisionError`, which `Monitor.ingest`'s fail-open path swallows, leaving the detector installed but dead for the whole run — the exact #331 symptom the PR set out to close. Failure scenario: `LatencyAnomalyDetector(h=0.0)` (library-embedder path, bypasses `Config`) on a warm baseline, first anomalous step raises and the latency safety net is off for the run; `h<0` makes `cusum > h` trivially true and pages on healthy traffic. Add the same `h <= 0` mirror guard to both constructors.

**2. `low` — the `k` check is not non-finite-safe, unlike the `h` check right beside it.**
`src/snagline/config.py:269` (`if value < 0:`) vs `src/snagline/config.py:253` (`if not value > 0:`)
`float('nan') < 0` and `float('inf') < 0` are both `False`, so a JSON `"cusum_k": Infinity` or a `nan` slack passes this new bound (and both detector guards, which repeat `if self.k < 0`). A NaN slack pins the accumulator via `max(0.0, nan) == 0.0`, so the alarm never fires. The PR notes #412 handles non-finite floats, but until that lands the "`k >= 0`" invariant is not actually enforced; using `if not value >= 0:` would make the two sibling checks consistent and self-contained.

**3. `low` — the end-to-end test never actually reaches the layer it claims to test.**
`tests/test_issues_331_333.py:608`
`with pytest.raises(...): Monitor.default(config=Config(token_runaway_enabled=True, token_cusum_k=-1.0), ...)` — Python evaluates the `Config(...)` argument first and its `__post_init__` raises, so `Monitor.default` never runs. The docstring says the test "reaches through `Monitor.default`", but that layer is unexercised; a future change that validates only in `Config` and not when the monitor builds detectors would not be caught here.

**Stack / merge note (not #422's own lines).** The diff is 8 files / +676 because it stacks on #337; the `_validated_counts_and_windows`, `BaselineStore`, and CLI `_resolved_max_versions` hunks are #337's. Two integration hazards ride along: (a) `_cmd_baseline` now calls `_resolved_max_versions()` before the `args.list_versions` branch, so the read-only `snagline baseline --store-dir X --list-versions --max-versions 0` returns exit 2 on a knob listing never uses; (b) the file is already `CONFLICTING` with `master`, and `_validated_counts_and_windows`'s `loop_window_size < 1` / `cascade_window_size < 1` rejection contradicts master's `tests/test_window_scaling.py::test_zero_window_size_does_not_raise_under_scaling`, which constructs `Config(window_scale_steps=5, max_window=8, loop_window_size=0)` and expects no raise (and master's `windowing.py:104` comment relies on `Config` accepting 0). Resolving toward the PR side fails a base-green test; these should be reconciled in #337 before either lands.

